PR #646, opened 2024-09-24
dev: enable `noImplicitAny`

tefkah:
- **dev: add noImplicitAny to tsconfig**
- **fix: fix all type errors uncovered by enabling noImplicitAny**

I noticed that we weren't using `noImplicitAny` in our `tsconfig`, which allows some `any`s to slip through. One common place for this to happen is in the arguments of functions, very easily making you specify `(arg: any) => ...` when you probably didn't want to do that.

This PR enables `noImplicitAny` everywhere except for `integrations`, as I don't really care for fixing issues there.

## Test Plan
1. Run tests
2. Convince yourself that no application logic is changed


core/playwright/formBuilder.spec.ts
```diff
@@ -96,8 +96,8 @@ test.describe("Submission buttons", () => {
 		page.on("request", (request) => {
 			if (request.method() === "POST" && request.url().includes(`forms/${FORM_SLUG}/edit`)) {
 				const data = request.postDataJSON();
-				const buttons = data[0].elements.filter((e) => e.type === "button");
-				const declineButton = buttons.find((b) => b.label === newData.label);
+				const buttons = data[0].elements.filter((e: any) => e.type === "button");
```
tefkah, line 99: any's are fine in tests

core/lib/server/pub.ts
```diff
@@ -328,11 +346,13 @@ export const createPubRecursiveNew = async ({
...
-		return {
-			error: validated.error,
-			cause: validated.error,
-		};
+		throw new Error(validated.error);
+		// return {
+		// 	error: validated.error,
+		// 	cause: validated.error,
+		// };
 	}
```
tefkah, line 356: i'll leave this here to remember to handle this next time we work on the API

core/app/c/[communitySlug]/settings/tokens/PermissionField.tsx
```diff
@@ -11,13 +11,63 @@ import { Popover, PopoverContent, PopoverTrigger } from "ui/popover";
...
+ *        }) => JSX.Element
+ *   } | undefined
+ * }
+ * ```
+ * The idea is that you can customize what form fields look like depending on the
+ * `permissionsSchema` configuration.
+ * E.g. the `write` field of the `pub` scope can be configured to only write to specific stages.
+ * You should be able to configure a custom element to do this.
+ *
+ */
```
Comment (tefkah) [inline core/app/c/[communitySlug]/settings/tokens/PermissionField.tsx:65]: even i had some problems remembering what the hell this did, so i left a comment for future reference

core/app/c/[communitySlug]/pubs/[pubId]/components/JsonSchemaHelpers.tsx
```diff
@@ -32,27 +32,32 @@ export function recursivelyGetScalarFields(schema: JSONSchemaType<AnySchema>, va
...
+				<CardContent>
+					{recursivelyGetScalarFields(
+						fieldSchema as JSONSchemaType<AnySchema>,
+						fieldValue ?? null
+					)}
+				</CardContent>
+			</>
+		);
+	});
+	return fields;
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/[pubId]/components/JsonSchemaHelpers.tsx:60]: this is one of the few places application logic is changed. I think this is equivalent, but do check.

allisonking (approved):
🦸

core/prisma/seed.ts
```diff
@@ -30,12 +30,10 @@ async function createUserMembers({
 	role: "editor" | "admin" | "contributor";
 	prismaCommunityIds: string[];
 }) {
-	let user;
-
 	await prisma.user.create({
 		data: {
 			slug,
-			email: user ? user.email : email,
```
kalilsn, line 38: i wonder what i was thinking 🤔

kalilsn (approved):
💯

core/prisma/seed.ts
tefkah: it did something at first, there was logic there to fetch the user from supabase, and if so, use the supabase users email. I just forgot to remove this in the cleanup!

PR #648, opened 2024-09-24
dev: upgrade vitest and properly set environment per type of test

tefkah:
- **dev: upgrade vitest related dependencies and setup proper environments for db tests-n**

## Issue(s) Resolved
Replaces weird and unmaintained `vitest-tsconfig-paths` with `vite-tsconfig-paths`.

While trying to fix a type error from `vitest-tsconfig-paths` in #646, I realized this packages is literally just an old version of `vite-tsconfig-paths`.

I installed `vite-tsconfig-paths` instead, which has proper types. This also encouraged an upgrade to vite 2.

While doing this upgrade, I realized that tests now would fail bc our environment variable validation treats the database tests as client tests, as we are running in the `jsdom` environment.

Rather than just turn off validation, I thought it better to just set the environment properly.

Database tests now run in the `node` environment if their file is like `*.db.test.ts`.

## Test Plan
1. Run `pnpm --filter core vitest`
2. See no errors.


core/vitest.config.mts
```diff
@@ -1,13 +1,18 @@
...
 import { defineConfig } from "vitest/config";
 
 export default defineConfig({
 	plugins: [react(), tsconfigPaths()],
 	test: {
 		globalSetup: ["./globalSetup.ts"],
 		environment: "jsdom",
+		environmentMatchGlobs: [
+			["**/(!db).test.ts", "jsdom"],
+			// all tests in tests/ with .edge.test.ts will run in edge-runtime
```
allisonking, line 12: is this comment right? what is edge?

allisonking (approved):

PR #649, opened 2024-09-24
Update pub values directly instead of appending

kalilsn:

### DIFF
--- a/core/app/components/PubCRUD/actions.ts
+++ b/core/app/components/PubCRUD/actions.ts
@@ -101,16 +101,16 @@ const stageMove = (pubId: PubsId, stageId?: StagesId) =>
 
 export const _upsertPubValues = async ({
 	pubId,
-	fields,
+	values,
 	stageId,
 }: {
 	pubId: PubsId;
-	fields: Record<string, JsonValue>;
+	values: Record<string, JsonValue>;
 	stageId?: StagesId;
 }) => {
 	// First query for existing values so we know whether to insert or update.
 	// Also get the schemaName for validation. We want the fields that may not be in the pub, too.
-	const toBeUpdatedPubFieldSlugs = Object.keys(fields);
+	const toBeUpdatedPubFieldSlugs = Object.keys(values);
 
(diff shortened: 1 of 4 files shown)


core/app/components/PubCRUD/actions.ts
```diff
@@ -154,11 +147,8 @@ export const _upsertPubValues = async ({
 		db
 			.insertInto("pub_values")
 			.values((eb) => {
-				return Object.entries(fields).map(([slug, value]) => {
+				return Object.entries(values).map(([slug, value]) => {
 					return {
-						id:
-							pubFields.find((pf) => pf.slug === slug && pf.pubId === pubId)
-								?.pubValueId ?? undefined,
```
kalilsn, line 161: We no longer need to pass the id here because we don't rely on the id to detect if there's an existing value. Now we use the unique constraint in the `on conflict` clause below

infrastructure/terraform/modules/deployment/main.tf
```diff
@@ -217,6 +217,7 @@ module "service_flock" {
        { name = "PGHOST", value = module.core_dependency_services.rds_connection_components.host },
        { name = "PGPORT", value = module.core_dependency_services.rds_connection_components.port },
        { name = "SUPABASE_URL", value = var.NEXT_PUBLIC_SUPABASE_URL },
+       { name = "PAGER", value = "less -S" },
```
tefkah, line 220: what is this?

tefkah (approved):
Looks good, seems to work!

If you could explain what setting the `PAGER` env is meant to do I'm happy to have it merged!

infrastructure/terraform/modules/deployment/main.tf
kalilsn: this is a convenience setting on the bastion host that improves (IMO) how psql handles long lines. the default pager is already `less` so this is just passing `-S` to it, which turns off wrapping. rn the wrapping is just unreadable for me:
image

but with `PAGER="less -S":
image

PR #651, opened 2024-09-24
Form Builder Name Edit

### LINKED ISSUE #449: Form builder name editing
## Motivation

So admins can rename a form.

## Requirements

- Add name editing button to form builder title
- Should open a dialog on click
- Should save independently of form builder saving

## Acceptance Criteria

https://www.figma.com/design/SqOTMElJn7fuPljfQL1gIY/PubPub-Platform?node-id=408-14023&m=dev

qweliant:
## Issue(s) Resolved
#449 

This allows a user to update the title to their form

## Test Plan
Navigate to the form editing page. Click the edit button. Edit the title. Be sure it updates in real time

image

image


qweliant:
note all of the Metadata stuff is the result of pnpm fix

core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts
```diff
@@ -0,0 +1,37 @@
...
+				.set({ name: name })
+				.where("communityId", "=", communityId)
+				.returning("id")
+		);
+		//
+		await formUpdateQuery.execute();
+	} catch (error) {
+		if (isUniqueConstraintError(error)) {
+			const column = error.constraint === "forms_slug_key" ? "slug" : "name";
+			return { error: `A form with this ${column} already exists. Choose a new ${column}` };
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts:32]: This logic for rendering the error doesn't make sense here, but checking for the unique constraint still makes sense! Can you just return the `"...this name already exists...` error without the templating?

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+		},
+	});
+
+	const runUpdateFormTitle = useServerAction(updateForm);
+	const onSubmit = (data: z.infer<typeof editFormTitleSchema>) => {
+		const result = runUpdateFormTitle({ communityId, name: data.name });
+		if (didSucceed(result)) {
+			toast({
+				title: "Success",
+				description: "Form name was successfully updated",
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:45]: image
slightly different wording in the designs

Comment (kalilsn) [reply core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:45]: This should also be centered, green, and have the check, like the save toast https://github.com/pubpub/v7/blob/b94ac8bbbdd4533305cc08e8affdf13ff35f9ac9/core/app/components/FormBuilder/FormBuilder.tsx#L178-L185

could you extract that config to its own toast component (or make it an option on the existing one) and then use it here so they have the same style? as a bonus you could also make it actually centered, since I seem to have missed that before!

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+								}}
+							>
+								Cancel
+							</Button>
+							<Button
+								type="submit"
+								variant="default"
+								className="bg-[#0090FF] hover:bg-[#0079E3]"
+							>
+								Save
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:96]: this says `Update` in the design

core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts
```diff
@@ -0,0 +1,37 @@
...
+		};
+	}
+	try {
+		const formUpdateQuery = await autoRevalidate(
+			db
+				.updateTable("forms")
+				.set({ name: name })
+				.where("communityId", "=", communityId)
+				.returning("id")
+		);
```
Comment (3mcd) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts:26]: This query will set every form in the community with id `communityId` to the same name. Instead, I think we should pass the form id to the server action and update the where clause to be `.where("id", "=", formid)`.

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+			<DialogTrigger asChild>
+				<Button
+					variant="link"
+					className="ml-2 text-sm text-gray-500 hover:text-blue-600 hover:underline"
+				>
+					Edit
+				</Button>
+			</DialogTrigger>
+			<DialogContent>
+				<DialogTitle>Edit Name</DialogTitle>
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:61]: There's an `<hr />` between the title and the content in the design

core/app/c/[communitySlug]/ContentLayout.tsx
```diff
@@ -1,10 +1,21 @@
 import type { ReactNode } from "react";
 
-const Heading = ({ title, action }: { title: ReactNode; action: ReactNode }) => {
+const Heading = ({
+	title,
+	action,
+	editFormTitleButton,
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/ContentLayout.tsx:6]: I don't think this prop is necessary on either component. Can you just pass the button in as part of the `title` prop?

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+						<DialogFooter className="flex justify-end">
+							<Button
+								variant="secondary"
+								className="mr-2"
+								onClick={() => {
+									form.reset();
+									setIsOpen(false);
+								}}
+							>
+								Cancel
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:89]: Clicking this button submits the form right now, because it doesn't have a type. Adding `type="button"` should fix that

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+			});
+			setIsOpen(false);
+		}
+	};
+	return (
+		<Dialog onOpenChange={setIsOpen} defaultOpen={false} open={isOpen} modal={true}>
+			<DialogTrigger asChild>
+				<Button
+					variant="link"
+					className="ml-2 text-sm text-gray-500 hover:text-blue-600 hover:underline"
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:55]: I think this is always blue in the designs, unless you know something I don't!

core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts
Comment (qweliant) [reply core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts:26]: that makes sense. I beleive i thought the prescence of a FKR would automatically update the the row at name_communityId

core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts
```diff
@@ -0,0 +1,37 @@
...
+		//
+		await formUpdateQuery.execute();
+	} catch (error) {
+		if (isUniqueConstraintError(error)) {
+			const column = error.constraint === "forms_slug_key" ? "slug" : "name";
+			return { error: `A form with this ${column} already exists. Choose a new ${column}` };
+		}
+		logger.error({ msg: "error creating form", error });
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts:34]: This log message isn't correct. I also think this is a situation where we should be explicitly reporting the error to sentry, since it means something unexpected happened. It's ok if you don't do this now, but it would be really nice if we had a utility function that both called logger.error and Sentry.captureException and then used it here (and in many other places!)

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
Comment (qweliant) [reply core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:55]: the icon used was only on hover so i figured its blue on hover but i can ask katy

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+		resolver: zodResolver(editFormTitleSchema),
+		defaultValues: {
+			name,
+		},
+	});
+
+	const runUpdateFormTitle = useServerAction(updateForm);
+	const onSubmit = (data: z.infer<typeof editFormTitleSchema>) => {
+		const result = runUpdateFormTitle({ communityId, name: data.name });
+		if (didSucceed(result)) {
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:42]: I don't completely know why but it looks like the form is both submitting successfully and throwing the unique constraint error here.

https://github.com/user-attachments/assets/9ce55d15-8fe0-4a89-9a99-d6f9a2b87a0a

Comment (kalilsn) [reply core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:42]: Oh you need to await the action!

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
Comment (kalilsn) [reply core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:55]: https://www.figma.com/design/SqOTMElJn7fuPljfQL1gIY/PubPub-Platform?node-id=408-14041&m=dev
this is what i was going off of

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
Comment (qweliant) [reply core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:42]: lmao

core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts
Comment (qweliant) [reply core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts]:
```typescript
export function logError(message: string, error: Error) {
	logger.error({ message, error });
	Sentry.captureException(error);
}

``` 

i have added this. I dont really know how fancy it should be but it does the two things together now.

core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts
```diff
@@ -0,0 +1,30 @@
...
+	try {
+		const formUpdateQuery = await autoRevalidate(
+			db.updateTable("forms").set({ name: name }).where("id", "=", formId)
+		);
+		await formUpdateQuery.execute();
+	} catch (error) {
+		if (isUniqueConstraintError(error)) {
+			return { error: `A form with this name already exists. Choose a new name` };
+		}
+		logError("error updating form", error);
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts:27]:
```suggestion
		logError("error updating form name", error);
```

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+				success
+					? "border-green-200 bg-green-100 text-green-800"
+					: "border-red-200 bg-red-100 text-red-800"
+			}`}
+		>
+			<div className="mr-3 text-2xl">{success ? <Check /> : <X />}</div>
+			<div className="text-lg font-semibold">{message}</div>
+		</div>
+	);
+};
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:40]: https://github.com/pubpub/v7/blob/b94ac8bbbdd4533305cc08e8affdf13ff35f9ac9/core/app/components/FormBuilder/FormBuilder.tsx#L178-L185

Could you use this implementation instead? The point of the component was to make sure these two toasts are the same, and matching the design:
image

Current save toast:
image

Toast on this branch:
image

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+								<FormItem>
+									<FormLabel>Form Name</FormLabel>
+									<FormControl>
+										<Input placeholder="Name" {...field} />
+									</FormControl>
+									<FormMessage />
+								</FormItem>
+							)}
+						/>
+						<DialogFooter className="flex justify-end">
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:91]: There needs to be some margin between the footer and the form above 
image.

The cancel button should also be a different variant, and it looks like the gap between the buttons should be smaller

image

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
Comment (kalilsn) [reply core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:100]: this wasn't fixed so i unresolved it

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
Comment (kalilsn) [reply core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:107]: still says save

core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts
```diff
@@ -0,0 +1,30 @@
+"use server";
+
+import { db } from "~/kysely/database";
+import { isUniqueConstraintError } from "~/kysely/errors";
+import { getLoginData } from "~/lib/auth/loginData";
+import { logError } from "~/lib/logging";
+import { autoRevalidate } from "~/lib/server/cache/autoRevalidate";
+import { defineServerAction } from "~/lib/server/defineServerAction";
+
+export const updateForm = defineServerAction(async function updateForm({ formId, name }) {
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/actions.ts:10]: this updateForm parameter should be explicitly typed to take a `FormsId` and string, otherwise it's `{formId: any, name: any}`

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+			});
+			setIsOpen(false);
+		}
+	};
+	return (
+		<Dialog onOpenChange={setIsOpen} defaultOpen={false} open={isOpen} modal={true}>
+			<DialogTrigger asChild>
+				<Button
+					variant="link"
+					className="ml-2 text-sm text-blue-500 hover:text-blue-600 hover:underline"
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:68]: the underline should always be present, not just on hover

Comment (kalilsn) [reply core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:68]: the `ml-2` is also not part of the design

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+	name: z.string().min(1, "Form name is required"),
+});
+
+const EditFormTitleButton = ({ formId, name }: { formId: FormsId; name: string }) => {
+	const [isOpen, setIsOpen] = useState(false);
+
+	const form = useForm<z.infer<typeof editFormTitleSchema>>({
+		resolver: zodResolver(editFormTitleSchema),
+		defaultValues: {
+			name,
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:29]: This default doesn't change if you edit the name and then click edit again. But tbh I don't entirely know why, since the name on the form builder page does!

kalilsn (approved):
Looks good now! One bug that is quite minor IMO that should be looked at but would also be fine in a followup PR/issue

core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx
Comment (qweliant) [reply core/app/c/[communitySlug]/forms/[formSlug]/edit/EditFormTitleButton.tsx:29]: https://github.com/user-attachments/assets/e82c19db-55f0-4d30-aa6f-3f8b6be003cb

like this but its not updating to the new name?

PR #658, opened 2024-09-26
dev: run integration tests in ci

### LINKED ISSUE #351: Setup production testing in CI
Resolved by #658

## Motivation

 - Developers can perform e2e testing on our prod CI pipeline
 - Developers are confident that tests work against production — e.g. [Reliability ](https://github.com/orgs/pubpub/projects/46/views/1?pane=issue&itemId=57862775)
 
## Requirements

- End-to-end testing runs after PRs are created against main

## Acceptance Criteria

- Pull down the core image from ECR that was built earlier in CI.
   - Ideally this should be pulled into the same `docker-compose.test.yml` file, so we can reuse the `postgres` and `jobs` instance running there
   - You can do this with environment variables in the `docker-compose.test.yml`
- Run playwright tests against it, ideally reusing an existing docker container in e.g. the CI container.
  - You will either need to figure out how to do run Playwright tests in Alpine containers, or find an efficient way to setup yet another container.
- The possibility for the playwright  to manipulate the database should be left open.
  - Meaning that the container that runs the tests should also be able to connect to the db, use our packages etc.
 
 
 ## Extra work

tefkah:
## Issue(s) Resolved
Resolves #351

This PR adds the ability to run integration tests in CI.

### Terms

- Base CI: the CI that runs from `ci.yml`, i.e. typechecking, linting, formatting, and non-playwright tests.
- E2E CI: the CI that runs from `e2e.yml`, i.e. the integration tests.
- Build/Build CI: the CI that runs from `build-all.yml`, i.e. the build of the core and jobs images (and integrations but who cares)

### Changes in general approach to CI

This PR reverts the general approach introduced in #543, which made it so tests we run inside of their own Docker container that was built during CI. Instead, we just run the tests inside of the runner directly, by doing e.g.

```yaml
      - name: Run type check
        run: pnpm type-check
```
rather than
```yaml
      - name: Run type check
        run: docker compose -f docker-compose.test.yml core run type-check
```

This was changed because:
1. Setting up caching while building a Docker image is hard, I haven't figured out a good way to do it that isn't slower than not caching at the moment.
2. We cannot run playwright tests in the same kind of Docker image as Build CI, thus we would have to make our Dockerfile even more complicated.
3. I find it personally easier to reason about the state of the CI this way. When building an image in CI that you end up using, you are defining steps that create inputs that build an image which takes other inputs etc etc. For Build CI this is acceptable, because reusing the built `core` image saves a lot of time, but setting up the tests is really fast, so i feel like it's less necessary.

### ✏️ Rough CI sketch

First, two CIs are run:
- 🧪 Base CI: typechecking, linting, formatting, and non-playwright tests.
    - steps
      - Install depedencies (can be cached)
      - Spin up DB (quite fast)
      - Run migrations (quite fast)
      - Run `preconstruct build` (somewhat slow, hard to cache atm)
      - Run type checking (~50seconds), linting (~20seconds), formatting (~15 seconds), and non-playwright tests  (~30 seconds). (can be cached with `turbo`)
    - This is fast as 
      - we are able to cache `pnpm install` if `pnpm-lock.yaml` has not changed.
      - we are able to cache `lint`, `test`, `type-check` and `format` if the packages they affect have not changed.
      - These things are not as easy to cache when running them in a docker image, as that requires being able to mess with volumes which i'm not capable of doing at the moment.
- 🔨 Build CI:
  - Steps:
    - Login to ECR
    - Build images for `core`, `jobs` and `integrations` (slow, ~5 minutes for core, hard to cache at the moment)
    - Push images to ECR

After those are completed, the `E2E CI` is run.
- 🏗️ E2E CI: run `pnpm --filter core playwright:test`
  - Steps:
    - Install depedencies (will be cached from the `Base CI` step)
    - Spin up DB and inbucket (quite fast)
    - Run migrations (quite fast)
    - Run `preconstruct build` (slow (20+s), hard to cache atm)
    - Pull down and spin up ECR images for `core` and `jobs` (quite fast, like 10seconds. Could maybe be cached, but probably not worth it as ECR download speed is faster than GHA cache download speed)
    - Install `playwright` (~20s, not worth caching)
    - Run `playwright:test` (~1m atm, will obviously grow)

### 🔁 Changes made to built Docker image

We are now building the Docker images with `output: standalone` (https://nextjs.org/docs/app/api-reference/next-config-js/output#automatically-copying-traced-files)

This has two main benefits:
1. It drastically reduces the size of the final docker image. Roughly 800MB compressed/2GB uncompressed ->  250MB compressed/750MB uncompressed. This makes deployments, CI, and our AWS bill much lighter.
2. It improves the time to build an image a lot, as we do not have to do the final `pnpm deploy` step anymore, which took a long time. 
  - The reason for this is that `output: standalone` already does what `pnpm deploy` intends to do but better: only keep the files that are actually used by the app.
  - Whereas `pnpm deploy` gets rid of all devDependencies, `output: standalone` also gets rid of files inside of those dependencies (or that we wrote) that we do not use.

This did create a bit of an awkward setup in the Dockerfile here

https://github.com/pubpub/platform/blob/9d3c905afddb0f2e2b036e0b51fd270804e10875/Dockerfile#L136-L160

If you have a better idea of handling this, please let me know!

What sucks about this setup is that, because we still use `pnpm deploy` for `jobs`, we now need to specify both a `target=jobs` and a `--build-arg PACKAGE=jobs`, whereas previously we could get away with using `--build-arg PACKAGE=jobs`

## Test Plan

1. See those beautiful green checks in CI!

Optionally you can try and run the CI locally!

You can do this using `act` (as is also described in the `DEVELOPMENT.md` file)
https://github.com/pubpub/platform/blob/695b4f10d3c42bded1d36fdf81ae1ea7636cec0f/DEVELOPMENT.md#L105-L140

1. `brew install act`
2. Make sure you have the AWS CLI installed and configured locally.
3. Copy the AWS secrets from `~/.aws/credentials` to `~/.aws/pubpub.secrets`
  - The two files should look like this:
    - ~/.aws/credentials
```
[pubpub]
aws_access_key_id = xxxxx
aws_secret_access_key = xxxxxx
```
    - ~/.aws/pubpub.secrets
```
AWS_ACCESS_KEY_ID=xxxxx
AWS_SECRET_ACCESS_KEY=xxxxxx
```
4. Then you can run CI like this:   
    - Base CI: `act -W .github/workflows/ci.yml --container-architecture linux/amd64 --secret-file ~/.aws/pubpub.secrets -j ci`
    - Complete CI (will take a while!): `act -W .github/workflows/on_pr.yml --container-architecture linux/amd64 --secret-file ~/.aws/pubpub.secrets -j e2e`

#### 👷 Docker layer caching

With the way we are running tests right now (on Github Action Runners), it is almost always less efficient to cache Docker layers than it is to just create them.

I have successfully been able to both cache the layers in Github Actions cache and in ECR cache, but both of these are too slow to be worth it.

The way you achieve this is by setting the `cache-from` and `cache-to` properties here (for ECR, for GHA cache it's even easier): 

https://github.com/pubpub/platform/blob/a9ab79f86dd455630775ae63313c04fbea41695e/.github/workflows/ecrbuild-template.yml#L80-L81

I can illustrate this with the example of caching the `pnpm install` step.

The `pnpm install` step in our Docker build can take like 30 seconds or more, which is worth caching (which is why we do so in Base CI and E2E CI).

However:
1. Fetching that layer already takes like 20 seconds.
2. _Extracting_ that layer takes _60_ seconds!!!
3. Pushing the cached layers to ECR takes like _100_ seconds!!!

So we turned 30 seconds into almost 200 seconds!! Not worth it at all!

Other, smaller layers could be worth caching (like installing `npm`, installing other dependencies), but still, pushing up to ECR takes a long time, nullifying the benefit.

##### 🤔 When worth it?

I think it might be worth revisiting Docker layer caching when either:

- Then non build steps of our Docker build take over 5 minutes (currently it's maybe 2 to 3, and unlikely to keep growing).
- We choose to run our actions on something else than Github Actions built in runners
  - Maybe if we host our runners on AWS as well, pushing and pulling from ECR will be faster.
  - Maybe if we use much more powerful runners, extracting times will be faster.
  - Maybe if we use different drop in runners from e.g. [Depot](https://depot.dev/blog/docker-layer-caching-in-github-actions), caching and extracting times will be faster.

For now, not caching is faster than caching.

#### 🔮 Future caching improvements we can still make

We could leverage `turbo` more by caching e.g. `preconstruct build` somehow. I have not been able to figure out how to do this yet, but it's adding `20+s` to every step of our CI.

Being able to share `turbo` caches would also massively speed things up, as we might be able to skip a build step altogther. This would require us to host a `turborepo` cache server somewhere, using e.g https://github.com/ducktors/turborepo-remote-cache.

This is not that hard I think.


tefkah:
@allisonking The formbuilder test still seems kind of flaky, you have any idea why?

image

Dockerfile
```diff
@@ -33,86 +37,114 @@ WORKDIR /usr/src/app
...
 
-################################################################################
-# Create a stage for building the application.
+
 FROM base as fetch-deps
 
-# install postgres utilities for scripts
-RUN apk add postgresql
-# if booting without a command, just sit and wait forever for a term signal
-CMD exec /bin/sh -c "trap : TERM INT; sleep infinity & wait"
```
kalilsn, line 43: do we ever run this image without an explicit command?

allisonking:
> @allisonking The formbuilder test still seems kind of flaky, you have any idea why?
> 
> image

Yeah I'm experiencing that too :( I'm guessing the form builder just has so much it has to prepopulate in the `beforeEach` (creating fields, creating a form, getting a pubId...) not quite sure what to do about it, though this might be a case where being able to directly populate the db could help

core/app/c/[communitySlug]/forms/actions.ts
```diff
@@ -113,4 +116,8 @@ export const createForm = defineServerAction(async function createForm(
 		logger.error({ msg: "error creating form", error });
 		return { error: "Form creation failed" };
 	}
+
+	const community = await findCommunityBySlug();
+	assert(community);
+	redirect(`/c/${community.slug}/forms/${slug}/edit`);
```
Comment (allisonking) [inline core/app/c/[communitySlug]/forms/actions.ts:122]: what was the motivation for moving this behavior here?

tefkah:
We are still slightly flaky. I don't really want to fix that now, it could maybe be a follow up PR to make these things slightly more stable. 
We could also disable the flaky tests for now.

3mcd (approved):
Looks great. I agree with the decision to move the test execution out of docker into the runner.

Thanks for the awesome PR description as usual. I have no questions that we haven't already discussed over slack. And awesome catch on the `output: standalone` thing! That's huge!

PR #661, opened 2024-09-26
Fix `NULLS` migration for pg14

allisonking:
## Issue(s) Resolved

https://knowledgefutures.slack.com/archives/CMAQM0BNV/p1727361690973769

I used `NULLS NOT DISTINCT` which is only available in pg15, not realizing we are only on pg14 in prod!


kalilsn (approved):
Looks good!

Issue #604, opened 2024-09-11
Default input component for array fields


gabestein:
## Motivation

So users can input an unconstrained list of things in a form.
So vector and array fields have a default input component that doesn't rely on admin configuration
So users can add multiple of unrestricted things to their pubs -- e.g., lists of DOIs or external resources

## Requirements

- Ability to input a string or a number
- Ability to add another string or number
- Ability to remove entries from the list
- Options:
   - Maximum and minimum number of items (default 0)
   - Required, which requires at least minimum number of items
   - Label
   - Maybe placeholder

## Acceptance Criteria

- Should be relatively reusable for the option interfaces for radio button and checkbox form field configuration

## Mockups 
[Multi Value Component](https://www.figma.com/design/jJIpEL3bBAHPFCHYoEPRzY/Platform-Components?node-id=608-5588&t=t2wMJw37do66Tut0-1) (open to ideas for other naming)

![Image](https://github.com/user-attachments/assets/114e32bd-3677-4b24-b22b-de14901df351)

Here is a gif demo of another component that I based my design off of. Hopefully this gives you a good idea of how I envision this working

![Image](https://github.com/user-attachments/assets/d12a2b0a-cee6-478e-8fd9-e786c17c90e0)

allisonking:
Closed by https://github.com/pubpub/platform/pull/667

PR #664, opened 2024-09-26
Add storybook to repo

allisonking:
## Issue(s) Resolved

To get started on https://github.com/pubpub/platform/issues/604, adding storybook!

Adds storybook for easier development of components. Includes one story for the Button component.

## Test Plan
* There's a new turbo package! You'll likely have to run `pnpm i && pnpm p:dev` at the root of the repo
* In the new `storybook` folder, run `pnpm run dev`
* You should be able to see storybook at localhost:6006. All tailwind stuff should be working, whether it's css classes added via storybook or in the original `ui` package.

image

## Notes
This is a repo-wide storybook, so we should be able to import components from other packages in here to test as well (as long as the package exports it—so not sure if we can test components in `core` easily, though I suppose us wanting to storybook it is perhaps a sign that it should go in `packages/ui`!)


packages/ui/src/stories/Button.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+export const Link: Story = {
+	args: {
+		variant: "link",
+	},
+};
+export const Loading: Story = {
+	render: (args) => (
+		<Button {...args}>
+			{/* animate-spin not working? */}
+			<Loader2 className="mr-2 h-4 w-4 animate-spin" />
```
allisonking, line 49: it's a mystery to me why the `animate-spin` class isn't working here... the other tailwind classes seem to work ok

allisonking: mystery solved: `animate-spin` was only being used by the storybook file, which isn't included in the tailwind config. having tailwind also monitor storybook files makes things work

tefkah: Should we have a separate config for storybook maybe? Not sure if that's something that's done, but don't the story book classes then end up in the prod css, even if they aren't used maybe?

allisonking: yeah I think you're right—I haven't found a way to add that tailwind stuff for only storybook yet, but I'll keep looking!

tefkah:
Nice!
Do you think we could set up storybook more generically in some way, like we have with e.g. `eslint`?
I recall travis also setting up StoryBook in another package over here: https://github.com/pubpub/platform/blob/tr/context-editor/packages/context-editor/.storybook/main.ts

would be nice not to have to dedupe that info! `turbo` has some docs for this i see: https://turbo.build/repo/docs/guides/tools/storybook

this repo might be able to serve as an example? https://github.com/gmickel/turborepo-shadcn-nextjs/blob/main/apps/storybook/.storybook/main.ts

packages/ui/src/stories/Button.stories.tsx
tefkah: Maybe setting up a storybook app instead of colocating the storybook instance in `ui` like how `turbo` suggests below could work?
Then we would indeed have a separate `tailwind.config` for storybook, which would not affect the `ui`.
See here?: https://github.com/gmickel/turborepo-shadcn-nextjs/blob/main/apps/storybook/tailwind.config.ts

allisonking:
woo thanks so much for the helpful links @tefkah ! I think I got it working. will update the PR description to explain how to use it now!

3mcd (approved):

PR #667, opened 2024-09-26
Multivalue component

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/604

Adds a multilvalue component and story for it

## Test Plan
* In packages/ui, start storybook via `pnpm run storybook`
* Play around with the `MultiValueInput` story
* As a side quest, I also added a story for `MultiSelect` since I wanted to quickly see what that one looked like in comparison

https://github.com/user-attachments/assets/781b6193-8405-4205-8675-a22643ed8042

As a form

https://github.com/user-attachments/assets/5b278cec-6804-41f3-926d-8009f962368c

## Notes

"required", "min/max", and a label were also requirements for this ticket, but I think it makes more sense to define those outside the component with the help of a zod schema and react-hook-form. See the "Form Usage" story for an example of how that would work!


packages/ui/src/stories/MultiValueInput.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+			"seal",
+			"mouse",
+			"turtle",
+			"lion",
+			"tiger",
+			"bobcat",
+		]);
+		return <MultiValueInput values={values} onChange={setValues} />;
+	},
+	play: async ({ canvasElement }) => {
```
allisonking, line 50: I was curious about storybook component testing so I added this. I think it's okay but seems a little buggy, or I'm doing something wrong, since this test doesn't pass on first render but always passes later. Anyway, this isn't automated at all, but may be worth looking more into if we do more storybook based dev

3mcd (approved):
Nice work!

Issue #614, opened 2024-09-16
Checkbox Field Input


gabestein:
## Motivation
So admins can add multi-select inputs to forms for constrained sets of options, e.g. demographics fields.

## Requirements
- For string and number input types, adds an input option for checkboxes.
- Field options:
  - Group description
  - Ability to add individual options
     - Option value (required) - the value stored in the pub field
     - Option name (optional) - if filled out, shows on the form instead of the value
  - Ability to include an "other" field, which allows a user to write in their own value (if input is string only)
  - Ability to set the minimum and maximum number of selections possible (defaults to 0)
  - Required, which requires at least the minimum number of selections
  - Defaults to unselected
- When added to the form, users can select from the checkbox options. Design does not need to be highly polished, can be "good enough"™ for now.

## Design
[Figma Link](https://www.figma.com/design/jJIpEL3bBAHPFCHYoEPRzY/Platform-Components?node-id=118-1310&t=t2wMJw37do66Tut0-1)

Note that the very last select 'User should select' is expected to have 3 options:
1. Select At Least
2. Select At Most
3. Select Exactly

ALSO TO NOTE: Previously on this side panel, we have been fixing the 'Cancel' and 'Add' buttons so they are always fixed to the bottom of the panel - but for these I would advise we do not do that, as I worry some admins will forget or not realize they can scroll, which would lead to potentially un-filled/unselected options

![Image](https://github.com/user-attachments/assets/1a4a5523-3d66-489c-b4c5-b5163e8a8f53)

PR #674, opened 2024-10-01
Add Number, NumericArray, StringArray schemas

### LINKED ISSUE #590: Add array(number), array(string), number field types
## Motivation

So users can add checkbox and multi-select fields to their forms.

## Requirements

- Users can create a array[number] field type that accepts an unlimited array of numbers.
- Users can create an array[string] field type that accepts an unlimited array of strings.
- Users can create a number field that accepts a single number

## Acceptance Criteria

- Will be a Javascript 64 float
- Does not cause create/edit form or Form editor to crash
- Consult with @ktyjscott/design about how to describe these fields to user

## Mockups
Adding design for agreed-upon names and descriptions
Image

3mcd:
## Issue(s) Resolved
Resolves #590 

This PR simply adds three new schemas: `Number`, `NumericArray`, and `StringArray`.

The new `Number` schema can be used in forms. It re-uses the existing `<TextElement />` input with `type="number"`. `NumericArray` and `StringArray` currently do not have any mapped inputs.

## Test Plan

- Add a new field to a community with the new `Number` schema
- Add two new fields: one using the `NumericArray` schema and one for `StringArray`
- Add these fields to a new form. Note that you can select an number input component for the field with the `Number` schema, but no component for the array components.
- Save the form and open the form for an existing pub. Change the value of the field with the backing `Number` schema. Confirm that the form persists the numeric value properly.


core/app/components/forms/elements/TextElement.tsx
```diff
@@ -27,6 +27,13 @@ export const TextElement = ({ label, name, ...rest }: ElementProps & InputProps)
 								{...fieldRest}
 								{...rest}
 								disabled={!isEnabled}
+								onChange={(e) => {
+									field.onChange(
+										rest.type === "number"
+											? Number(e.target.value)
+											: e.target.value
+									);
+								}}
```
3mcd, line 36: This casts the value of the `type="number"` input to a string before emitting it to the field change handler

core/prisma/schema.prisma
```diff
@@ -192,7 +195,6 @@ model PubValue {
 
   // Add userId and integrationId to track who created the value
 
-  @@unique([pubId, relatedPubId, fieldId])
```
3mcd, line 195: Correct me if I'm wrong, but I think this is necessary because @allisonking did not remove this after altering the migrations manually?

packages/schemas/src/index.ts
```diff
@@ -7,31 +7,40 @@ import {
...
+		case CoreSchemaType.NumericArray:
+			return NumericArray;
+		case CoreSchemaType.String:
+			return String;
+		case CoreSchemaType.StringArray:
+			return StringArray;
+		case CoreSchemaType.URL:
+			return URL;
+		case CoreSchemaType.Vector3:
+			return Vector3;
```
3mcd, line 43: I may have impulsively alphabetized the switch cases...

core/prisma/schema.prisma
allisonking: yes, I think that's right! although it's a little odd that the unique constraint can't be captured in the schema, it's just hanging out in the migrations... I think that should be okay though!

packages/schemas/src/schemas.ts
```diff
@@ -7,26 +7,31 @@ registerFormats();
...
+
 export const Vector3 = Type.Array(
 	Type.Number({
 		error: "Invalid number",
 	}),
 	{
 		minItems: 3,
 		maxItems: 3,
-		description:
+		descriptin:
```
allisonking, line 34:
```suggestion
		description:
```

allisonking (approved):
looks great!!

core/app/components/forms/elements/TextElement.tsx
kalilsn: What do you think about making this numeric cast functionality part of our `Input` component? I am working on something that needs it but outside of a FormElement

kalilsn: Oh that is a nice find! Eric mentioned that this was more complicated than I thought in a slack message on Thursday so we were not going to bother with it. Sorry we didn't update you and ty for confirming!

allisonking: oh okay! no worries, I'm only poking around here because I thought I'd try to get this PR across the line so I can use it for https://github.com/pubpub/platform/issues/614 :)

tefkah:
awesome, thanks for fixing the tests @allisonking !

PR #675, opened 2024-10-01
Fix pub creation and editing

### LINKED ISSUE #673: Creating and editing pubs fails with error
## Motivation

Submitting a form fails with an `undefined` error reported in the failure toast. Logging the error reveals the following message:

```
error: duplicate key value violates unique constraint "pub_values_pubId_fieldId_key"
```

## Requirements

Restore submission behavior to forms.

## Acceptance Criteria

Because forms update existing pub values, this error is actually expected. But it isn't being handled by [Kysely's `onConflict` handler for the query](https://github.com/pubpub/platform/blob/main/core/app/components/pubs/PubEditor/actions.ts#L195).

kalilsn:
## Issue(s) Resolved
#673 
The query we use to update pub values uses the presence of an `id` on the values we insert to determine whether they should be new rows or updates to existing rows. Because of an error in the prior query, we were not setting the id on every value, which this PR fixes.

This also fixes a bug where we were creating invalid dates while making the default values for the pub create form.
## Test Plan
1. Update a pub, it should succeed!
2. Create a dateTime field and add it to a pub type
3. On the pubs page, open the create pub form and select the type with the dateTime field
4. Make sure you can create a pub of that type

### DIFF
--- a/core/app/components/pubs/PubEditor/actions.ts
+++ b/core/app/components/pubs/PubEditor/actions.ts
@@ -145,18 +145,19 @@ export const _updatePub = async ({
 		const existingPubFieldValues = await autoCache(
 			trx
 				.selectFrom("pub_fields")
-				.leftJoin("pub_values", "pub_values.fieldId", "pub_fields.id")
+				.leftJoin("pub_values", (join) =>
+					join
+						.onRef("pub_values.fieldId", "=", "pub_fields.id")
+						.on("pub_values.pubId", "=", pubId)
+				)
 				.where("pub_fields.slug", "in", toBeUpdatedPubFieldSlugs)
-				.distinctOn("pub_fields.id")
-				.orderBy(["pub_fields.id", "pub_values.createdAt desc"])
+				.where("pub_fields.isRelation", "=", false)
 				.select([
 					"pub_values.id as pubValueId",
-					"pub_values.pubId",
 					"pub_fields.slug",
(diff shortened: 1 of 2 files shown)


core/app/components/pubs/PubEditor/actions.ts
```diff
@@ -145,18 +145,19 @@ export const _updatePub = async ({
 		const existingPubFieldValues = await autoCache(
 			trx
 				.selectFrom("pub_fields")
-				.leftJoin("pub_values", "pub_values.fieldId", "pub_fields.id")
+				.leftJoin("pub_values", (join) =>
+					join
+						.onRef("pub_values.fieldId", "=", "pub_fields.id")
+						.on("pub_values.pubId", "=", pubId)
+				)
```
kalilsn, line 152: This query is intended to select field ids, field names, and field schemaNames for all of the values submitted so they can be validated. It has a secondary purpose of selecting pub_value ids for each of these values if they already exist. 
Before this change it was joining the values table without narrowing the results by the current pubId, then using distinctOn to filter out the extra rows.

kalilsn: Also since we're already doing this query before the updates, arguably we should be selecting the values here too and using that information to avoid updating rows when the value doesn't change!

3mcd (approved):

tefkah:
nice, great job!

Issue #557, opened 2024-08-31
Allow creation of stageless pubs


gabestein:
## Motivation

So pub, stage, and community contributors can create pubs (including child pubs) without also adding them to a stage. E.g. creating an evaluation that is yet to be sent; an internal reviewer e.g. copyeditor creating a review

## Requirements

- The stage dropdown on the pub creation/edit form should be made optional.
- A user should be able to select a null option with the label "No stage". This option should be selected by default.
- Submitting the form with the null option selected creates the pub with no stage, or in the case of updating a pub, removes it from its current stage.

## Acceptance Criteria

- Both the `createPub` and `updatePub` server actions should account for an empty/null stage.

PR #676, opened 2024-10-01
Allow creating a pub without a stage

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/557

Allows the stage to be undefined on pub create. There's a new dropdown in the stage dropdown that is "No stage" which leads to an undefined stage.

The bulk of this change was actually using `createPubRecursiveNew` instead of `createPub`. `createPubRecursiveNew` is used by the API and already takes an optional stage ID. I had to rejigger `createPubRecursiveNew` a little bit to allow for empty pub values

## Test Plan

* Visit http://localhost:3000/c/croccroc/pubs
* Click "Create"
* The create form should pop up with "No stage" as the default under "Stage"
* You should be able to save the pub without selecting a stage

image


core/app/components/pubs/PubEditor/actions.ts
```diff
@@ -45,24 +45,26 @@ export const createPub = defineServerAction(async function createPub({
...
+		);
+
+		const query = stageId
+			? pubQuery.with("stage_create", (db) =>
+					db.insertInto("PubsInStages").values((eb) => ({
+						pubId: eb.selectFrom("new_pub").select("new_pub.id"),
+						stageId,
+					}))
+				)
+			: pubQuery;
```
tefkah, line 67: you can solve the type issue this causes below with

```ts
				? (query as typeof pubQuery).insertInto("pub_values").values((eb) =>
						pubValueEntries.map(([pubFieldSlug, pubValue]) => ({
							pubId: eb.selectFrom("new_pub").select("new_pub.id"),
							value: JSON.stringify(pubValue),
							fieldId: eb
								.selectFrom("pub_fields")
								.where("pub_fields.slug", "=", pubFieldSlug)
								.select("pub_fields.id"),
						}))
					)
				: (query as typeof pubQuery).selectFrom("new_pub").select([])
```

tefkah, line 67: kind of ugly, but this fixes the type issue below.

For readability's sake though, it might be worth splitting these paths up into separate functions anyway though. Even if it causes some code duplication.

```suggestion
		const query = (
			stageId
				? pubQuery.with("stage_create", (db) =>
						db.insertInto("PubsInStages").values((eb) => ({
							pubId: eb.selectFrom("new_pub").select("new_pub.id"),
							stageId,
						}))
					)
				: pubQuery
		) as typeof pubQuery;
```

core/app/components/pubs/PubEditor/PubEditorClient.tsx
```diff
@@ -171,12 +177,14 @@ export function PubEditorClient(props: Props) {
 			}
 
 			const result = await runCreatePub({
-				pubId,
+				body: {
+					id: pubId,
+					pubTypeId: pubTypeId as PubTypesId,
+					stageId: stageId as StagesId,
+					values: enabledPubValues as Record<string, any>,
```
allisonking, line 184: I don't like the cast to any here, but I wasn't sure how to get these types to match up 😕 

```
Type 'PubValues' is not assignable to type 'Record<string, Json>'.
  'string' index signatures are incompatible.
    Type 'JsonValue' is not assignable to type 'Json'.
      Type 'JsonObject' is not assignable to type 'Json'.
        Type 'JsonObject' is missing the following properties from type 'Json[]': length, pop, push, concat, and 35 more.ts(2322)
integrations.ts(97, 2): The expected type comes from property 'values' which is declared here on type 'CreatePubRequestBodyWithNullsNew | { id?: string | undefined; pubTypeId: string; parentId?: string | null | undefined; values: Record<string, Json>; assigneeId?: string | undefined; children?: (Omit<CreatePubRequestBodyWithNulls, "stageId"> & { stageId?: StagesId | undefined; })[] | undefined; stageId?: StagesId | undefined; }'
```

3mcd: I think this is fine for now! I've had trouble with the `Json` and `JsonValue` types before, too. At some point we can do a pass to improve the types for pub values objects across the project.

3mcd (approved):
Works great!

PR #677, opened 2024-10-02
refactor: detach some dialogs from their buttons

tefkah:

Performance demo

https://github.com/user-attachments/assets/9ca3f929-67c7-4797-a80a-c9a118acf076


core/app/c/[communitySlug]/pubs/PubList.tsx
```diff
@@ -32,14 +31,7 @@ const PubListInner: React.FC<Props> = async (props) => {
 	return (
 		<div className={cn("flex flex-col gap-8")}>
 			{allPubs.map((pub) => {
-				return (
-					<PubRow
-						key={pub.id}
-						pub={pub}
-						token={token}
-						searchParams={props.searchParams}
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/PubList.tsx:40]: search params no longer needed bc the `updatePub` is detached.

core/app/c/[communitySlug]/pubs/[pubId]/components/PubChildrenTableWrapper.tsx
```diff
@@ -33,12 +33,7 @@ const NoActions = () => {
 const getChildPubRunActionDropdowns = (row: ChildPubRow, pageContext: PageContext) => {
 	const stage = row.stages[0];
 	return stage && row.actionInstances.length > 0 ? (
-		<PubsRunActionDropDownMenu
-			actionInstances={row.actionInstances}
-			pubId={row.id}
-			stage={stage}
-			pageContext={pageContext}
-		/>
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/[pubId]/components/PubChildrenTableWrapper.tsx:41]: `pageContext` no longer necessary because the `runactionform` is detached

core/lib/regexp.ts
```diff
@@ -0,0 +1 @@
+export const uuidRegex = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
```
tefkah, line 1: maybe there's a better place to store reusable regexs

core/app/components/ActionUI/ActionRunDialog.tsx
tefkah, line 1: the action run dialog, similar to the edit pub dialog, is detached from the action run button and can now be placed separately on a page

core/app/components/ActionUI/ActionRunDialog.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+	if (!actionInstance || !community) {
+		// TODO: Show error state
+		return null;
+	}
+
+	return (
+		<>
+			<DialogHeader>
+				<DialogTitle>{actionInstance.name || actionInstance.action}</DialogTitle>
```
tefkah, line 63: could maybe be `Run ${actionInstance.name} Action` to be more clear

tefkah:
```suggestion
				<DialogTitle>Run {actionInstance.name || actionInstance.action} Action</DialogTitle>
```

core/app/components/PathAwareDialogButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+	className?: string;
+	children?: React.ReactNode;
+	id: string;
+};
+
+export const PathAwareDialogButton = (props: PathAwareDialogueProps) => {
+	const { isPending, toggleDialog: toggleDialogue } = usePathAwareDialogSearchParam({
+		id: props.id,
+	});
+	// Should maybe be a link, but then it plays less nicely with `nuqs`
```
tefkah, line 22: this is less of a "best practices" question and more a philosophical question. Is this a button that open a dialogue that happens to use a queryparam to keep state, or is it a link to a page that happens to open a dialog that covers the current page?

you could argue both, but i think having it be a button is fine. if we wanted it to be a link, we would have to manually construct the url ourselves again (find current url, construct the correct query param, add it to the already existing query params etc), which negates some of the benefits of using `nuqs` (it does all of that for us)

PR #678, opened 2024-10-02
Add detailed component configuration

### LINKED ISSUE #551: Add label, help text, and optional field to form field editor panel
## Motivation
- So users can add custom label and helper text to form elements where needed
- So forms fields have labels by default rather than being blank

## Requirements
- Input
  - Supports: Email, URL, String
  - Label
  - Placeholder
  - Help Text
  - Required
- Date Picker
  - Supports: DateTime
  - Label
  - Help text
  - Required
  - No placeholder, no default value, automatically defaults to today's date
- Member Picker
  - Supports: Member
  - Label
  - Help text
  - No placeholder
- ComboSlider
  - Supports: Vector(3)
  - Label
  - Help text
  - No placeholder
  - Required
  - No default value so that it can be made required -- if possible, starting position for the handles should be: 0, 50, 100 but should require users to set them to meet required validation
- Textarea
   - Supports: String
   - Label
   - Help text
   - Placeholder
   - MaxLength: number input
   - MinLength: number input
   - Defaults to resizable
   - Hard-code columns and rows to match design
   - Required
- Checkbox
   - Supports: boolean
   - Group label, as described
   - Checkbox label
   - Default is configurable as true or false, default default is false
   - No required

## Acceptance Criteria
- Migrate labels from old to new JSON column
- Changes to form rendering for new labels and required behavior

## Designs
This was in the designs, but seems to have dropped off the form field requirements (unintentionally?). See comment on #444.

Image

Image

kalilsn:
## Issue(s) Resolved
#551 
- Modifies FormElement to once again use the component rather than the schema to choose what to render
- Adds configuration forms for all existing component types
- Renders labels etc. based on that configuration

Let me know if you think the structure should be improved to make it easier to add new elements. Adding a new component requires a bit more boilerplate than I would have liked, but I don't think it's enough overhead that we need to do anything about it yet.

## Test Plan
- Visit the seeded form in the croccroc community and configure the various fields
- Construct the form fill url by getting an existing pubId, then make sure the fields appear and render alright, and that validation, saving, and submission work. 
http://localhost:3000/c/croccroc/public/forms/review/fill?pubId={pubId}
- Make sure the pub update and pub create forms aren't broken


core/app/components/FormBuilder/ElementPanel/ComponentConfig/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { InputComponent } from "db/public";
+import { Skeleton } from "ui/skeleton";
+
+import type { ComponentConfigFormProps, InnerFormProps } from "./types.ts";
+
+export const ComponentConfig = ({ component, ...props }: ComponentConfigFormProps) => {
+	let ConfigForm: ComponentType<InnerFormProps>;
+	switch (component) {
+		case InputComponent.checkbox:
+			ConfigForm = dynamic(() => import("./Checkbox.tsx"), {
```
kalilsn, line 16: I think there's no real way to make this less cumbersome, since template strings don't work here. LMK if there's something I'm missing though!

core/app/components/forms/elements/FIleUploadElement.tsx
```diff
@@ -1,10 +1,12 @@
 "use client";
 
 import dynamic from "next/dynamic";
+import { Value } from "@sinclair/typebox/value";
```
allisonking, line 4: the file name has a capital I `FIle`

core/app/components/forms/types.ts
```diff
@@ -1,4 +1,7 @@
+import type { CoreSchemaType } from "db/public";
+
 export interface ElementProps {
-	label: string;
 	name: string;
+	config: any;
```
allisonking, line 5: is it possible to type this? since it seems like it gets used downstream quite a bit

core/prisma/schema.prisma
```diff
@@ -192,7 +192,6 @@ model PubValue {
 
   // Add userId and integrationId to track who created the value
 
-  @@unique([pubId, relatedPubId, fieldId])
```
allisonking, line 195: 🙏

core/app/components/FormBuilder/types.ts
```diff
@@ -27,7 +27,7 @@ export type InputElement = baseElement & {
 	type: ElementType.pubfield;
 	fieldId: PubFieldsId;
 	required: boolean | null;
-	label?: string | null;
+	label: never;
```
allisonking, line 30: is `label` used at all anymore now that we have `config`? should we remove it as part of the migration?

core/prisma/migrations/20241002220742_remove_label_column_from_inputs/migration.sql
```diff
@@ -0,0 +1,15 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `label` on the `form_elements` table. All the data in the column will be lost.
+  - A unique constraint covering the columns `[pubId,relatedPubId,fieldId]` on the table `pub_values` will be added. If there are existing duplicate values, this will fail.
```
allisonking, line 5: can probably remove this comment about the unique constraint?

core/prisma/migrations/20241002220742_remove_label_column_from_inputs/migration.sql
```diff
@@ -0,0 +1,15 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `label` on the `form_elements` table. All the data in the column will be lost.
```
allisonking, line 4: hm is this true? I don't think the column `label` was dropped?

allisonking (approved):
awesome job, I tested locally and it seems to work really well!!

core/app/components/forms/types.ts
kalilsn: i could type it as unknown (probably?) but i'm not sure we can do better than that since it's a json object whose structure varies depending on the component. this is kind of a big drawback, so i am definitely not sure this is the right approach. the other big issue i foresee is that we'll need to be quite cautious about breaking changes to these schemas.

core/app/components/FormBuilder/types.ts
kalilsn: i had initially removed it, but then i realized the buttons do use the label. i could make buttons use config instead and then we could drop the label. the unique constraint preventing duplicate button names would need some reworking too

core/prisma/migrations/20241002220742_remove_label_column_from_inputs/migration.sql
kalilsn: oops yes i manually edited this and left some bad comments in, thanks!

core/app/components/forms/elements/FIleUploadElement.tsx
kalilsn: shockingly hard problem to solve on a mac lol

core/app/components/FormBuilder/types.ts
allisonking: ah right, I forgot about the buttons! hmm if it makes the unique constraint not work then I think we can keep it as-is (though maybe with a comment that it's only used for labels?)

kalilsn: will do! i also updated the explicit types in the FormBuilder/types.ts file so that label is `never` on input elements

kalilsn: arguably we could make the unique constraint work (just put it on the config field), but it would also mean that buttons also have to be subject to all the annoying caveats of storing config in a json file (weak types/validation required, risky updates). so i'm going to leave it as is for now because of that mainly!

Issue #681, opened 2024-10-03
Moving slider inputs on form crashes page


gabestein:
### What went wrong, step-by-step?

1. Create a form with a slider input. 
2. Visit the form
3. Drag the slider input 
4. Page crashes

### What did you expect to happen?
Page stays up and slider updates to new value as expected. 

### Who reported this? (If not you)
Khalil/Dave

PR #682, opened 2024-10-03
Fix onChange of slider component

allisonking:
## Issue(s) Resolved

Closes https://github.com/pubpub/platform/issues/681

I don't know why this didn't happen before, but it appears that we need to be careful not to pass `onChange` to `Confidence` if we are passing `onValueChange`. This PR just makes sure we don't pass `onChange` along with `onValueChange` since that results in a bad set state

## Test Plan

See repro steps in https://github.com/pubpub/platform/issues/681 only the page shouldn't crash anymore!

https://github.com/user-attachments/assets/886044f5-9de9-4879-b8a1-748918e9084b

## Notes

~~There is also a forwardRef error which I was able to fix but not without being TS compliant, so I'm still looking into it...~~

Fixed! There was a forwardRef error showing up in the console which was a result of dynamically importing the slider component. Fixed using the strategy described in https://github.com/vercel/next.js/issues/4957#issuecomment-413841689

### DIFF
--- a/core/app/components/forms/elements/ConfidenceElement.tsx
+++ b/core/app/components/forms/elements/ConfidenceElement.tsx
@@ -27,22 +27,27 @@ export const Vector3Element = ({ label, name }: ElementProps) => {
 			control={control}
 			name={name}
 			defaultValue={[0, 0, 0]}
-			render={({ field }) => (
-				<FormItem className="mb-6">
-					<FormLabel className="text-[0.9em]">{label}</FormLabel>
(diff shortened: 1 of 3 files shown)


tefkah:
nice catch!

in react 19 well never run into this problem again hopefully, as you can just use `ref` as a prop then
https://react.dev/blog/2024/04/25/react-19#ref-as-a-prop

kalilsn (approved):
💯

PR #683, opened 2024-10-03
fix: restore base image build on pr/main 

tefkah:
- **fix: restore building the base image, as we need it for migrations**
- **feat: make ECS deploy task wait until service is stable**

## Issue(s) Resolved
Core was not deploying correctly

For the integration tests PR, i commented out the building of the base image (up until the `monorepo` step in `Dockerfile`), because I thought it was unnecessary.
Turns out we use it for migrations, and core needs it to be able to run. Thus, it is restored.

I added the `wait-for-service-stability: true` flag on the ECS deploy task, which should hopefully fail CD (show us an x) if the service does not start succesfully in the future.

## Test Plan
1. Make sure base image builds and is pushed during CI.
2. Send it.

### DIFF
--- a/.github/workflows/deploy-template.yml
+++ b/.github/workflows/deploy-template.yml
@@ -123,7 +123,7 @@ jobs:
                   task-definition: ${{ steps.task-def-service.outputs.task-definition }}
                   service: ${{ env.ECS_SERVICE }}
                   cluster: ${{ env.ECS_CLUSTER }}
-                  wait-for-service-stability: false
+                  wait-for-service-stability: true
 
             - name: Deploy Amazon ECS task definition including migrations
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-template.yml
```diff
@@ -134,3 +134,4 @@ jobs:
                   task-definition: ${{ steps.task-def-migration.outputs.task-definition }}
                   service: ${{ env.ECS_SERVICE }}
                   cluster: ${{ env.ECS_CLUSTER }}
+                  wait-for-service-stability: true
```
kalilsn, line 137: nice!!

kalilsn (commented):
Looks good!

PR #684, opened 2024-10-03
fix: set DATABASE_URL and HOSTNAME correctly for next services

tefkah:
- **fix: restore building the base image, as we need it for migrations**
- **feat: make ECS deploy task wait until service is stable**
- **fix: manually set the hostname to 0.0.0.0 to avoid weird errors when deploying to ecs**
- **fix: copy over docker env file to set DATABASE_URL correctly when deploying**

## Issue(s) Resolved
`core` wasn't starting on ECS

When outputting in standalone mode, next manually starts the server at 
```js
const currentPort = parseInt(process.env.PORT, 10) || 3000
const hostname = process.env.HOSTNAME || '0.0.0.0'
```

By default, the hostname is the internal hostname of the docker container, which is something like `e03a171cfe70`

This works fine when running locally, but ECS probably does something weird and custom which causes that to fail. 

I'm now manually setting `ENV HOSTNAME "0.0.0.0" in the Dockerfile, which should hopefully maybe fix this?

### DIFF
--- a/.github/workflows/deploy-template.yml
+++ b/.github/workflows/deploy-template.yml
@@ -123,7 +123,7 @@ jobs:
                   task-definition: ${{ steps.task-def-service.outputs.task-definition }}
                   service: ${{ env.ECS_SERVICE }}
                   cluster: ${{ env.ECS_CLUSTER }}
-                  wait-for-service-stability: false
+                  wait-for-service-stability: true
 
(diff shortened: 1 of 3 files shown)


kalilsn (approved):
seems like a reasonable idea, let's try it!

PR #685, opened 2024-10-03
Implement optional form fields with toggle

### LINKED ISSUE #679: Make field toggles work on form fill page
## Motivation

So that form elements can actually be optional, regardless of whether the underlying element supports an indeterminate state (e.g. checkboxes, which don't usually). 

## Requirements
- Make the pub toggles functional on the form fill page
- Only show a toggle if the field is not "required"
- Add "required" option to checkbox component type (since it will work now)

kalilsn:
## Issue(s) Resolved
#679 
- Adds toggle provider to form fill page
- Filters out disabled fields from submit payload and form schema
- Fixes a bug where the uppy dashboard would cover the toggle for file upload fields

## Test Plan
Configure some fields on the croccroc review form to be required, and others optional
Create an empty pub and construct the form fill url for it
Make sure the required fields don't have toggles
Only fill out the required fields
Make sure the form can be submitted by toggling off the optional fields

## Notes
Right now this doesn't play super nicely with autosaving. If you enter data in an optional field and then autosave it, this won't remove that data. @3mcd and I discussed and this seems good enough for now. In the future, we may modify the pub update form to use these toggles as _delete_ buttons for a value, which we could also do on this page to avoid autosave issues. Alternatively, we will probably eventually put autosaved data somewhere other than pub values pre-submission, at which point this problem will resolve itself.


core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx
```diff
@@ -5,9 +5,9 @@
  * and the other is a form as in react-hook-form.
  */
 import type { ReactNode } from "react";
-import type { FieldValues } from "react-hook-form";
+import type { FieldPath, FieldValues } from "react-hook-form";
```
Comment (allisonking) [inline core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx:8]: nit: doesn't look like `FieldPath` is used

allisonking (approved):
woot this works great locally! nice change 💯

PR #686, opened 2024-10-03
Em/hostname

3mcd:
Taking another stab at setting the hostname at the task level to get the standalone Next deployment working!


infrastructure/terraform/modules/core-services/main.tf
```diff
@@ -48,51 +48,54 @@ resource "aws_secretsmanager_secret_version" "password" {
...
 module "assets_bucket" {
   source  = "terraform-aws-modules/s3-bucket/aws"
   version = "~> 4.0"
 
-  block_public_acls = false
-  bucket        = var.assets_bucket_url_name
+  block_public_acls        = false
+  block_public_policy      = false
+  ignore_public_acls       = false
+  restrict_public_buckets  = false
```
3mcd, line 94: Set these options to what was manually set in S3

infrastructure/terraform/modules/deployment/variables.tf
```diff
@@ -1,69 +1,73 @@
...
 }
 variable "NEXT_PUBLIC_SUPABASE_PUBLIC_KEY" {
   description = "Supabase anon public key"
-  type = string
+  type        = string
+}
+
+variable "HOSTNAME" {
+  description = "Hostname used by standalone Next app"
 }
```
3mcd, line 67: Added this variable to override the default HOSTNAME variable provided by ECS so that the Next app can bind to the proper hostname when starting.

kalilsn (approved):
approved and noting that we already applied these changes without issue!

Issue #669, opened 2024-09-30
Allow forms to create pubs


3mcd:
## Motivation
So communities can collect external submissions.

## Requirements

- When submitting a form without a pubId search parameter, a new pub should be created instead of updated.
- Adds button to copy a link to create pub form on form editor and in forms list page

## Acceptance Criteria

- Autosave functionality should be disabled when the `pubId` search param is missing.
- Invoking the server action without a pub ID should create the pub rather than update an existing one.

allisonking:
Currently we require a stage ID when creating a pub, but we don't have a place in the form builder where the user is required to specify a stage on pub create. The closest we have is the post-submission flow, but stage isn't required there.

PR #687, opened 2024-10-03
Create pubs from forms

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/669

On a page like `/croccroc/public/forms/{form-slug}/fill`, you no longer need the `pubId` query parameter. Now, if there is no `pubId`, the form will have a "create" behavior. Instead of querying for an existing pub, we will create a pub.

## Test Plan
* Create a form
* Visit the form at `/{community}/public/forms/{form-slug}/fill`
* You should see a blank form and you should be able to submit it **once you fill all fields out** (see note)
* Once submitted, you should see the post-submission behavior (and if you used any pub specific markdown, it should still render ok)
* You should be able to see created pubs

### Other things to test
* File upload works and uploads to the proper pub ID
* Configure submit button with
  * a stage (the pub should be created in this stage)
  * special markdown, i.e. with `:value{field="croccroc:title"}`

### Filling out a form without a pub id

https://github.com/user-attachments/assets/11c09b07-9a49-4c11-adb7-96e68054429c

### Copying link to form

https://github.com/user-attachments/assets/f2dc9032-df6f-4f70-8436-037c5718bfa1

## Notes
Right now the form lets you submit even if not all fields are filled out, but there will be an error because the pub fails validation on create. I think we should come back to this after https://github.com/pubpub/platform/pull/685, which might make it not a problem at all!


core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx
```diff
@@ -67,19 +68,22 @@ const preparePayload = ({
 };
 
 /**
- * Date pubValues need to be transformed to a Date type to pass validation
+ * Set all default values
+ * Special case: date pubValues need to be transformed to a Date type to pass validation
```
Comment (allisonking) [inline core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx:72]: Before, `defaultValues` would be based on the existing pubId. Now that we create, this was being set to `{}` which is _okay_ but not great, since that gives errors like switching from uncontrolled to controlled states when going from `undefined` to a `string` for many fields. This sets (hopefully) reasonably typed defaults for each type of schema field

core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx
```diff
@@ -214,7 +248,7 @@ export const ExternalFormWrapper = ({
 			<form
 				onChange={formInstance.handleSubmit(handleAutoSave)}
 				onSubmit={formInstance.handleSubmit(handleSubmit)}
-				className={cn("relative flex flex-col gap-6", className)}
+				className={cn("relative isolate flex flex-col gap-6", className)}
```
Comment (allisonking) [inline core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx:251]: fun CSS: avoids z-index shenanigans where the little form element enablers appear above the header
image

core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/SubmitButtons.tsx
```diff
@@ -16,14 +16,11 @@ export const SubmitButtons = ({
...
 		return (
-			<Button
-				id="submit-button-default"
-				type="submit"
-				disabled={isDisabled}
-				className={className}
-			>
-				Submit
-			</Button>
+			<div className={className}>
```
Comment (allisonking) [inline core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/SubmitButtons.tsx:19]: the default button was unintentionally filling the whole width

Before:
image


After:
image

core/app/components/MemberSelect/MemberSelectClient.tsx
```diff
@@ -96,8 +96,13 @@ export function MemberSelectClient({
 
 	const onInputValueChange = useDebouncedCallback((value: string) => {
 		const newParams = new URLSearchParams(params);
+		const oldParams = newParams.toString();
 		newParams.set(queryParamName, value);
-		router.replace(`${pathname}?${newParams.toString()}`, { scroll: false });
+		// Only change params when they are different, otherwise can cause race conditions
```
allisonking, line 101: this is only a workaround, and I'm not sure what the 'proper' solution is. basically it seems like this `onInputValueChange` was triggering on form submit, and therefore overwriting the parameters I was trying to set on form submit (namely, that the the form was completed). instead, this seemed to trigger, and there'd be a race condition in setting the URL params, so sometimes the form would never get marked as completed.

I don't know why this func is triggered when the form submits, so the workaround here is just to not set the new params if they aren't different. It seems like the deeper problem might be that this func is firing too often though

core/lib/server/pub.ts
```diff
@@ -395,19 +395,17 @@ export const createPubRecursiveNew = async <Body extends CreatePubRequestBodyWit
 	const parentId = parent?.id ?? body.parentId;
 	const stageId = body.stageId;
 
-	const pubFieldsForPubTypeObject = await getPubFields({
```
allisonking, line 398: We don't necessarily want to restrict our pub values to only those of the pub type, since our forms might add different values. So now we query for pub fields by community instead (which is likely a cached query)

packages/schemas/src/index.ts
```diff
@@ -47,6 +47,32 @@ export function getJsonSchemaByCoreSchemaType(coreSchemaType: CoreSchemaType) {
...
+		case CoreSchemaType.Email:
+			return "";
+		case CoreSchemaType.URL:
+			return "";
+		case CoreSchemaType.MemberId:
+			return "";
+		case CoreSchemaType.FileUpload:
+			return undefined;
+		case CoreSchemaType.Null:
+			return undefined;
```
kalilsn, line 69: Looks like this needs cases for the new elements that I merged in yesterday! I would have added them but I wasn't totally sure what was best for number!

core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx
Comment (kalilsn) [reply core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx:76]: thanks for the explanation! i had seen those errors and wasn't sure where they were coming from!

kalilsn (commented):
nice job! this looks great, works well with the toggles, and that copy button is gonna save me so much time! ready to merge IMO other than those missing defaults

packages/schemas/src/index.ts
```diff
@@ -49,24 +49,30 @@ export function getJsonSchemaByCoreSchemaType(coreSchemaType: CoreSchemaType) {
...
+		case CoreSchemaType.NumericArray:
+			return [];
+		case CoreSchemaType.String:
+			return "";
+		case CoreSchemaType.StringArray:
+			return [];
+		case CoreSchemaType.URL:
+			return "";
+		case CoreSchemaType.Vector3:
+			return [0, 50, 100];
```
kalilsn, line 75: good catch!

kalilsn (approved):

packages/schemas/src/index.ts
allisonking: ah yeah good question... I experimented a bit and it seems like `undefined` actually works okay here! also I got things mixed up in my [earlier comment](https://github.com/pubpub/platform/pull/687#discussion_r1790483864)—we're actually okay on controlled vs uncontrolled state here, but this helps such that when we validate, we don't get a whole sea of red because our string fields have undefined values

(if we don't have a default value for String here, then "title" and "content" immediately get marked as invalid which is alarming, as shown in this vid)


https://github.com/user-attachments/assets/8d7f4537-09aa-451d-af67-dba5c811e8e0

kalilsn: ah right, makes sense! i've seen that on some of our other forms!

PR #691, opened 2024-10-03
Make vscode auto imports use 'import type' when appropriate

kalilsn:
- VSCode will now use `import type` when generating imports, so you won't have to fix it with eslint anymore
- `pnpm fix` runs eslint --fix / prettier -w now


tefkah:
Great idea! I didn't even know that setting existed, great find.
I wanted something easy to do on a Monday morning, so I took the liberty to already fix all the typescript errors this generated if you don't mind!

tefkah (approved):

tefkah:
I think we could now disable this ESLint rule no?

kalilsn:
@tefkah if someone manually adds an import that violates this, would we still be able to autofix it without the eslint rule? 

And thanks for fixing the type errors! I noticed those as I was signing off and was not that excited to fix them lol

PR #694, opened 2024-10-07
fix: also add hostname to other deployments

tefkah:
## Issue(s) Resolved
Integrations and bastion ECS tasks failing.

This adds the `HOSTNAME` env var to the task definition of the integrations and `bastion`.

For the integrations I'm relatively confident this was the same issue as for core, as the changes made were the same and the error in ECS is the same.

For `bastion` i'm less sure, as I'm a bit out of the loop as to what it does and how it is deployed!

Do we need to first run `terraform` locally before this will affect deployment?

## Test Plan
1. send it? Test it locally?


infrastructure/terraform/environments/blake/main.tf
```diff
@@ -78,4 +79,5 @@ module "deployment" {
   NEXT_PUBLIC_SUPABASE_URL        = local.NEXT_PUBLIC_SUPABASE_URL
   NEXT_PUBLIC_SUPABASE_PUBLIC_KEY = local.NEXT_PUBLIC_SUPABASE_PUBLIC_KEY
   ASSETS_BUCKET_NAME              = local.ASSETS_BUCKET_NAME
+  HOSTNAME                        = local.HOSTNAME
```
tefkah, line 82: I think we should also add it here, partially because the `terraform`  linting tool was complaining, and partiallly because of what eve wrote here

https://github.com/pubpub/platform/blob/b31d3b0918ab2e5a7777f7a36e7808bbb4582f7a/infrastructure/terraform/README.md#L30

kalilsn:
@tefkah IMO the procedure for PRs with terraform changes should be that you post the output of `mask tf plan` in the PR for the reviewer.

After approval, then you run `mask tf apply` either immediately before or after merge to main. Before could be risky if the branch isn't up to date with main, after could be risky if `terraform apply` fails, so I don't feel too strongly about which we should do (just be aware that the branch should be up to date, and that terraform apply can fail).

PR #696, opened 2024-10-07
fix: make bastion not exit

tefkah:
- **fix: make bastion idle instead of exiting**
- **fix: use array of strings for command, as intended**
- **fix: make bastion not exit fr**

## Issue(s) Resolved
Bastion was constantly exiting and restarting, because i removed the default idling command in #658. 

I restored this command in the task definition only. I already tested this locally, and CI now succeeds.


tefkah:
Already tested this locally, got go ahead from @kalilsn , am merging immediately

PR #701, opened 2024-10-08
dev: almost enable strict mode fully for `core`

tefkah:
## Issue(s) Resolved
Turns out we weren't even using `strict: true` in `core`! Wack! 

Enables `strict: true`, which enforces more strict typechecking. Follow up to #646, as `strict: true` includes setting `noImplicitAny: true`.

One issue this e.g. helped catch is that we often do not set the type of the form when we use `useForm`.
This can lead to differences in the expected arguments passed to `form.handleSubmit`.

I did chicken out a bit and set `useUknownInCatchVariables: false` (which is set to `true` by `strict: true`), mostly because it introduced a bunch of errors that I would rather fix in a generalized way, which I did not feel like doing rn.

### Logic changes

I tried to make as few logic changes as necessary, but it was necessary for the `Edit/CreateField` form. I think `strict: true` helped make behavior a bit more clear here, as it was somewhat ambiguous what we were doing with legacy pub fields, e.g. ones with `schemaName: null` and `isRelation: false`. I solved this by just not allowing you to edit legacy pub fields.
If this is not desirable let me know, and we can solve this some other way. Would especially like your thoughts @allisonking as I think you worked on this the most. 

## Test Plan
1. Check that you can still edit pubfields.


core/app/c/[communitySlug]/fields/FieldForm.tsx
```diff
@@ -61,21 +62,13 @@ type FormValues = z.infer<typeof schema>;
 const DEFAULT_VALUES = {
 	id: "",
 	name: "",
-	schemaName: null,
+	// can't be `null` and `isRelation: false` by default so should be undefined
+	schemaName: undefined,
 	slug: "",
 	isRelation: false,
-};
+} as const;
```
Comment (tefkah) [inline core/app/c/[communitySlug]/fields/FieldForm.tsx:69]: this is an example of undefined behavior that `strict: true` now helped catch.

Because I needed to do `useForm<FormValues>` below, Typescript now yelled at me that you cannot set the default values as `schemaName: null` and `isRelation: false`.

The `as const` is necessary here to treat `isRelation` as `false`, rather than `boolean`.

core/app/c/[communitySlug]/fields/FieldsTable.tsx
```diff
@@ -34,6 +53,11 @@ export const FieldsTable = ({ fields }: { fields: PubField[] }) => {
 
 	const columns = getFieldTableColumns();
 	const handleRowClick = (row: Row<TableData>) => {
+		// logic change: Legacy fields can no longer be edited
+		if (!isNotLegacyField(row.original)) {
+			return;
+		}
+
 		setEditField(row.original);
```
Comment (tefkah) [inline core/app/c/[communitySlug]/fields/FieldsTable.tsx:61]: this is the logic change.

core/actions/types.ts
```diff
@@ -58,7 +65,7 @@ export type Action<
 	P extends ActionPubType = ActionPubType,
 	C extends ZodObjectOrWrapped = ZodObjectOrWrapped,
 	A extends ZodObjectOrWrappedOrOptional = ZodObjectOrWrappedOrOptional,
-	N extends string = string,
+	N extends ActionName = ActionName,
```
tefkah, line 68: Because of `strict: true` enabling [`strictFunctionTypes`](https://www.typescriptlang.org/tsconfig/#strictFunctionTypes), we often ran into the issue that we were doing something like this:

```ts
import type { Action as ActionName } from 'db/public'

// newly defined function which uses the Enum to narrow action names
const addActionFn = (action: ActionName) => { /* ... */ }

// needs to be string
const onAdd = (action: string) => { /*...*/ }

export function AddButton = ({action, onAdd}: {
	action: Action // the whole thing, not just the name
	onAdd: (actionName: string) => void // needs to take actionname: string bc Action['name'] was `string`
}) => {
		return <Button onClick={()=>onAdd(action.name)} > Add {action.name} </Button>
}
```

Then, when you wanted to use the more correct `actionFn` which uses the `ActionName` enum as arg, like so

```tsx
<AddButton onAdd={addActionFn} action={action} />
```
You  would get an error with `strict: true`, bc with `strictFunctionTypes` the function arguments need to match more closely.

I think this restriction is in general good, see the linked Typescript page for examples of errors.

To solve this, I thought it was nicest to just require actions to use the `Action` enum generated from the `db`, as those are the only allowed options anyway.

core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx
```diff
@@ -151,15 +159,15 @@ export const ExternalFormWrapper = ({
...
+			evt: React.BaseSyntheticEvent | undefined,
 			autoSave = false
 		) => {
 			const pubValues = preparePayload({
 				formElements,
 				formValues,
 				toggleContext,
 			});
-			const submitButtonId = evt?.nativeEvent.submitter?.id;
+			const submitButtonId = isSubmitEvent(evt) ? evt?.nativeEvent.submitter?.id : null;
```
Comment (tefkah) [inline core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx:170]: This is admittedly kind of annoying to have to do, but it's because `form.handleSubmit` only takes `React.BaseSyntheticEvent`, and again bc `strict: true` enables `strictFunctionTypes`, you cannot assign a more strict function type (using `evt: React.BaseSyntheticEvent<SubmitEvent>` where `evt: React.BaseSyntheticEvent` is expected).

core/lib/server/render/pub/renderMarkdownWithPub.ts
```diff
@@ -317,16 +326,12 @@ export const renderMarkdownWithPub = async (
...
-		processor = processorBase.use(rehypeStringify);
+		const processor = processorBase.use(rehypeRemark).use(remarkStringify);
+		return process(processor, text);
 	}
 
-	const result = await processor.process(text);
-
-	return result.toString().trim();
+	const processor = processorBase.use(rehypeStringify);
+	return process(processor, text);
```
tefkah, line 336: this just solved that conditional assignment a bit nicer. This lead to an error, as the fourth `Node` in `Processor` was a more specific thing after using `remarkStringify`, so you couldn't do that.

integrations/evaluations/tsconfig.json
```diff
@@ -7,6 +7,7 @@
 		},
 		"strictNullChecks": true,
 		"noImplicitAny": null,
+		"strict": false,
```
tefkah, line 10: I didn't feel like solving these errors for the integrations tbh.

core/app/c/[communitySlug]/fields/FieldForm.tsx
Comment (tefkah) [reply core/app/c/[communitySlug]/fields/FieldForm.tsx]: Scratch the above, i needed to revert that bc it would crash the app

PR #702, opened 2024-10-08
Try using exec when playwright installs browser

allisonking:
## Issue(s) Resolved

Call `exec` from `e2e.yml` so that we use the installed version of playwright in `node_modules`. To try not to get an error where we are using two different playwright versions!


tefkah:
Brilliant! Makes sense that that would work, but very annoying that the way `playwright` recommends you to do this leads to such annoying errors.

tefkah (approved):

Issue #613, opened 2024-09-16
Dropdown Field Input


gabestein:
## Motivation
So admins can add single-select inputs to forms for large lists of options, e.g. U.S. states.
So users can select from dropdown options on forms.

## Requirements
- For string and number field types, adds a new "Dropdown" input option.
- Field options:
  - Group description -- @ktyjscott is this the same as Help Text?
  - Ability to add individual options. Just the option value for now.
     - Option value (required) - the value stored in the pub field
     - Option name (optional) - if filled out, shows on the form instead of the value
  - Add a setting for placeholder text
  - If the input is required, the form should fail validation if the placeholder text is selected.
  - If the input is not required, no value should be created/updated for the input's corresponding pub field.
- When input field is added to form, users can select an option from a dropdown.

## Acceptance Criteria
- Add the new input component type for "Dropdown"
- Make it allowed for strings and numbers
- Add the schema for the "Dropdown" input component config form.
- Add the "Dropdown" input component config form. This form should use the multi-value component introduced in #667 for the "Dropdown Values" field.
- Add a condition to the `FormElement` component that renders a dropdown component with the options configured in the form settings.
- The options should be validated against the backing pub field's schema. For example, if the number schema is selected, the user should see an error if they try to add an alphanumeric option.
- Renders a basic dropdown on the form. Design does not need to be highly polished, can be "good enough"™ for now.

## Mockups

> Note: This mockup is a little misleading because this ticket is for fields with String and Number schemas. The mockup includes a field with a String Array schema, which is out of scope for this ticket.

[Figma Link](https://www.figma.com/design/jJIpEL3bBAHPFCHYoEPRzY/Platform-Components?node-id=118-1310&t=t2wMJw37do66Tut0-1)

![Image](https://github.com/user-attachments/assets/133877a1-2f84-4a08-9f07-8975d1a15b97)

allisonking:
@ktyjscott we noticed that the Select dropdown has "Help Text" instead of "Description", which the checkbox group and radio group have. Is this intentional, or should they all be the same?

![Image](https://github.com/user-attachments/assets/f34303fb-82ca-46f0-b853-897bcbcdbf22)

Issue #612, opened 2024-09-16
Radio Button Field Input


gabestein:
## Motivation
So admins can add radio inputs for single-select inputs to forms for constrained sets of options, e.g. likert scales.

## Requirements
- For string and number field types, adds an input option for radio buttons.
- Field options:
  - Group description
  - Ability to add individual options
     - Option value (required) - the value stored in the pub field
     - Option name (optional) - if filled out, shows on the form instead of the value
  - Ability to include an "other" field, which allows a user to write in their own value (if input is string only)
  - Required
  - Defaults to unselected
- When added to the form, users can select from the radio options. Design does not need to be highly polished, can be "good enough"™ for now.

## Mockups

[Figma Link](https://www.figma.com/design/jJIpEL3bBAHPFCHYoEPRzY/Platform-Components?node-id=118-1310&t=t2wMJw37do66Tut0-1)

![Image](https://github.com/user-attachments/assets/59f4c2c1-310d-4f27-9526-8d685558aa07)

PR #703, opened 2024-10-08
Checkbox, Radio, and Select dropdown field input on form builder

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/614
Closes https://github.com/pubpub/platform/issues/612
Closes https://github.com/pubpub/platform/issues/613

Adds the checkbox group, radio group, and select dropdown field as options for StringArray and NumericArray fields in the form builder.

## Test Plan
1. Add a StringArray field
2. Add a NumericArray field
3. Add these to your form. You should see the option to add them either as a checkbox group, a radio group, or a select dropdown. You can click between these and the form should change a bit between components. [Design](https://www.figma.com/design/jJIpEL3bBAHPFCHYoEPRzY/Platform-Components?node-id=118-1310&node-type=frame&t=0TuVdrpViDCi3FCa-0)
4. You should be able to fill things out and then check in your db that the `config` obj of your form element saved properly.

https://github.com/user-attachments/assets/b0dad953-e29d-4179-b52d-44df28194b8e

## Notes

I had to redo the RadioGroupCard that holds the InputComponent options since that renders as a button. This causes a hydration error since html doesn't allow buttons inside of buttons—in this case, we want to render our checkbox group/radio group/select dropdown inside the RadioGroupCard, and those components all render as buttons with our shadcn implementation. So I changed that bit of code to use an old school `<input type="radio"/>` instead to avoid rendering buttons in buttons.

This PR will conflict with https://github.com/pubpub/platform/pull/704 but #704 should make things better here!


core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
```diff
@@ -136,7 +220,15 @@ export const InputComponentConfigurationForm = ({ index }: Props) => {
 	const component = form.watch("component");
 
 	const onSubmit = (values: ConfigFormData) => {
-		update(index, { ...selectedElement, ...values, updated: true, configured: true });
+		// Some `config` schemas have extra values which persist if we don't Clean first
+		const cleanedConfig = Value.Clean(componentConfigSchemas[values.component], values.config);
```
allisonking, line 224: for example, the RadioGroup has a field `includeOther` which the SelectDropdown does not have. without this line, if we had saved a component as a RadioGroup, then switched it to a SelectDropdown and saved that, the `includeOther` would still persist in `config`, when really we'd like `config` to look like what we expect of a `SelectDropdown` config

core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
```diff
@@ -84,24 +153,39 @@ const ComponentSelect = ({
...
 			{components.map((c) => {
 				const { name, demoComponent: Component } = componentInfo[c];
 				return (
-					<RadioGroupCard key={c} value={c} className="flex h-[124px] w-full flex-col">
-						<div className="flex h-[88px] w-full items-center justify-center p-3">
-							{Component && <Component element={element} />}
+					<div key={c}>
+						{/* We use regular input instead of a RadioGroup here because the RadioGroup
+						renders buttons, and we cannot render buttons inside of buttons. Some of the demo components
+						need to render buttons. */}
```
kalilsn, line 163: good catch! this seems to work about as well as the cards did although there's a small regression where clicking inside the demo component's inputs doesn't bubble up to select a different radio input.

i think it might be ok to merge as is but if so we should add a ticket to improve these a bit by making it harder to interact with the demo component inputs through some or all of:
- removing them from the tab sequencing
- maybe disabling the inputs when they're not in a currently selected component?
- bubbling click events up

core/app/components/FormBuilder/ElementPanel/ComponentConfig/MultivalueBase.tsx
```diff
@@ -0,0 +1,71 @@
...
+							<MultiValueInput
+								type={isNumeric ? "number" : undefined}
+								onChange={(e: string[]) => {
+									if (isNumeric) {
+										return field.onChange(e.map((v) => +v));
+									}
+									field.onChange(e);
+								}}
+								value={field.value ?? []}
+							/>
```
kalilsn, line 62: Clicking the drag handle on an added option causes the form panel to close. I'm guessing it's a button without a type? It's not so bad since it's saving but probably worth fixing!

https://github.com/user-attachments/assets/c05d736e-5a90-4ea2-8937-061a1295baae

kalilsn (commented):
Nice work overall! I have a few UX notes/nits, but they don't have to be blockers if you think they'd be better handled in the future.

core/app/components/FormBuilder/ElementPanel/ComponentConfig/MultivalueBase.tsx
kalilsn: also, given how much this comes up, maybe we should default our Button component to `type="button"`? I doubt we're relying on that defaulting to `submit` anywhere

tefkah: I'd be in favor of setting `type="button"` to be the default, `type="submit"` as the default (in general) was always confusing to me. I get that that was the only way you could use a button in ye olden days, but I think specifically saying "This button is used to submit a form" is obviously the case you would want to call out by setting the `type`.

allisonking: oh great catch! I've confirmed that changing the Button component's `type` to `button` fixes things, so I'll make that the default 👍

core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
allisonking: I learned about [inert](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert) today which seems like a pretty good use case for this—we don't _really_ want those demo components interactive, but they can be if they're currently selected, so that the user can get a feel for them I guess?

here's the behavior now. so they are only part of the tab order if the item is selected. maybe they shouldn't be part of tab order at all? 🤔 but at least clicking on a radio button selects the component selector, if it wasn't selected already, and only if you click it again can you really click the radio

https://github.com/user-attachments/assets/b95da185-ec94-4660-af6f-04cf0dad395a

kalilsn: very nice, I'm also just learning about inert for the first time! seems like that is exactly what we need!

kalilsn: taking them out of the tab order entirely would be great but i think this is fine for now!

kalilsn (approved):
great job!

core/app/components/FormBuilder/ElementPanel/ComponentConfig/MultivalueBase.tsx
kalilsn: well i guess we were relying on that in some places. but i'm glad we're not now!

PR #704, opened 2024-10-09
refactor: improve dynamic import ux and typesafety for form element InputComponents

tefkah:
- **refactor: make loading component config more efficient and slightly easier to read**
- **refactor: almost make the form config components more typesafe**
- **fix: simplify the types and remove all type errors**

This is a follow up to #678, it includes some stuff I intended to leave as feedback but forgot to do! 

This PR includes two changes, which I explain in the videos below (partially because I wanted to show off the difference in the dynamic import, partially because i'm lazy)

### Improved dynamic import

I made it so the dynamic imports for the config only have to happen once, rather than every time the component is rendered.

https://github.com/user-attachments/assets/c0a959ae-c609-49ab-b3a0-d53b174876a7

### Improved typesafety

I made the `config` property of the form field typesafe by doing some generic shenanigans.

https://github.com/user-attachments/assets/20ff7439-db1c-4c15-8269-db093a39a89a

## Test Plan

1. Create Form elements with every component type, make sure they still work.


packages/schemas/src/schemaComponents.ts
```diff
@@ -57,12 +57,12 @@ export const confidenceIntervalConfigSchema = Type.Object({
...
+export const componentConfigSchemas = {
 	[InputComponent.checkbox]: checkboxConfigSchema,
 	[InputComponent.textArea]: textAreaConfigSchema,
 	[InputComponent.textInput]: textInputConfigSchema,
 	[InputComponent.datePicker]: datePickerConfigSchema,
 	[InputComponent.fileUpload]: fileUploadConfigSchema,
 	[InputComponent.memberSelect]: memberSelectConfigSchema,
 	[InputComponent.confidenceInterval]: confidenceIntervalConfigSchema,
-} as const;
+} as const satisfies Record<InputComponent, TObject>;
```
tefkah, line 68: This is the change that actually allows the type inference to work, typescript now recognizes the unique type of each key, rather than them all being `TObject`

packages/schemas/src/schemaComponents.ts
```diff
@@ -19,7 +19,7 @@ export const componentsBySchema: Record<CoreSchemaType, InputComponent[]> = {
 	[CoreSchemaType.MemberId]: [InputComponent.memberSelect],
 	[CoreSchemaType.Vector3]: [InputComponent.confidenceInterval],
 	[CoreSchemaType.Null]: [],
-} as const;
+} as const satisfies Record<CoreSchemaType, InputComponent[]>;
```
tefkah, line 22: I changed this because I experimented with finding the `CoreSchemaType` by the `InputComponent`.
I realized this wasn't worth it in the end, so I didn't do anything with it, but this type is more `correct` so I say we keep it.

core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
```diff
@@ -119,7 +119,7 @@ export const InputComponentConfigurationForm = ({ index }: Props) => {
 	const { schemaName } = selectedElement;
 	const allowedComponents = componentsBySchema[schemaName];
 
-	const form = useForm<ConfigFormData>({
+	const form = useForm<ConfigFormData<(typeof allowedComponents)[number]>>({
```
tefkah, line 122: We could also solve this by making typing `ConfigFormData` like so

```ts
type ConfigFormData<I extends InputComponent = InputComponent> = /*...*/
```

However this can quickly become kind of messy as when you forget to pass in an `InputComponent` when necessary you get an incorrect type.

Alternatively, this line could be

```suggestion
	const form = useForm<ConfigFormData<InputComponent>({
```

Might be more efficient for the compiler.

packages/schemas/src/schemaComponents.ts
allisonking: for my own understanding, what makes this type more `correct`?

allisonking:
this is amaaaazing! esp the videos :)

packages/schemas/src/schemaComponents.ts
kalilsn: I think (though @tefkah should validate this) that using `satisfies` rather than a type assertion here will still make sure that this is a `Record<CoreSchemaType, InputComponent[]>`, but also allow the compiler to keep the more specific information about what each property maps to. Previously the typescript compiler would decide the expression `componentsBySchema[CoreSchemaType.Vector3]` is `InputComponent[]`, but with this change it will infer the type `[InputComponent.confidenceInterval]`

core/app/components/FormBuilder/ElementPanel/ComponentConfig/index.tsx
```diff
@@ -1,61 +1,40 @@
...
-			return null;
-	}
-	return <ConfigForm {...props} />;
+import type { ComponentConfigFormProps } from "./types.ts";
+
+const toDynamic = (path: string) =>
+	// this dynamic import path needs to provide enough information for webpack/turbopack
+	// to be able to find it. The relative path and the extension are enough, but something like
+	// `import(path)` will not work.
+	dynamic(() => import(`./${path}.tsx`), {
```
kalilsn, line 16: honestly I didn't spend that much time trying to get to this version, because of this warning i saw on the next/dynamic docs!
image

and then i read a few issues where people had various hacks to get around this and I just thought most of them looked like they probably didn't work. thanks for showing me the way!

core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
kalilsn: to be perfectly honest, I don't really understand what that first type means. any relevant reading material or docs you can link me to?

and for your suggested change, how are those equivalent? `(typeof allowedComponents)[number]` is more specific than `InputComponent` right?

kalilsn (approved):
thanks so much for these changes and the helpful videos explaining! I had noticed that unfortunate reloading and wasn't sure what to do about it, and the typesafety is very nice here in the config forms.

core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
tefkah: Sorry, that was kind of vague.

Currently `ConfigFormData` is typed like this

```ts
type ConfigFormData<I extends InputComponent> = /*...*/
```
It thus requires an `InputComponent` argument.

By instead adding the `= InputComponent`
```ts
type ConfigFormData<I extends InputComponent> = /*...*/
```

we can use it like 

```ts
const form = useForm<ConfigFormData>( // ...
```
instead, as we don't need to pass in a type argument anymore.


As for the suggested change, it is currently equivalent because `(typeof allowedComponents)[number]` is the same as `InputComponent`, as all possible `InputComponent`s are allowed.

tefkah: You may notice that this doesn't feel that strict. Ideally I would like a situation like this:

```ts

const component = InputComponent.Text

const form = useForm<ConfigFormData<typeof component>>(/*...*/)
```

But we can't really do that, because the `form` is the thing setting the `component` as well. 


So sadly we end up with a kind of "incorrect" `ConfigFormData` that looks like so

```ts
// as a reminder, this is what ConfigFormData looks like
export type ConfigFormData<I extends InputComponent> = {
	required: boolean | null;
	component: I;
	config: Static<(typeof componentConfigSchemas)[I]>;
};

declare const config: ConfigFormData<InputComponent>
// type looks like so
typeof config = {
	required: boolean | null;
	component: InputComponent.TextArea | InputComponent.Number | /*...*/;
	config: Static<(typeof componentConfigSchemas)[InputComponent.TextArea | InputComponent.Number | /*...*/]>;
}
```

This is sadly the unfortunate way that Typescript handles unions in generics: it doesn't really do any "spreading"/"iterating" per se, it just fills it in in every place you put the generic.

You can force typescript to do this spreading, by doing something like this

```ts
export type ConfigFormDataIterated<I extends InputComponent> =  I extends I ? {
	required: boolean | null;
	component: I;
	config: Static<(typeof componentConfigSchemas)[I]>;
} : never;
```
As a refresher, `X extends Y` is roughly "is X a subtype of Y". This is a conditional (the only way to do them in typescript) which is always true, so we will always go to the part after `?`. The false condition, `: never`, never happens, so we just put `never` there.

"Ok cool thomas but why the hell are we doing that?" Doing `I extends I` (or `I extends any`) here is kind of like "for each element in the union I". This is really annoying, I wish there was another way to iterate over unions in typescript, but there is not.

Now the result type will look like this

```ts
declare const config: ConfigFormDataIterated<InputComponent>
// type looks like so
typeof configIterated = 
{
	required: boolean | null;
	component: InputComponent.TextArea;
	config: Static<(typeof componentConfigSchemas)[InputComponent.TextArea>;
} | {
	required: boolean | null;
	component: InputComponent.Checkbox;
	config: Static<(typeof componentConfigSchemas)[InputComponent.Checkbox>;
} | /*...*/

```

This is a much more accurate discriminated union. Now you could theoretically do

```ts
if(configIterated.component === InputComponent.TextArea){
	config.config // <-- this would now be the correct config. Before it would just be the union of all configs.
}
```

So why not do this? I did try this, but if you want this to work properly, you also need to do it for `ComponentConfigFormProps`. When I did though, everything got realllly messy, and I couldn't figure it out anymore.

So I thought: we mostly just need this to be very accurate when defining the InputComponents themselves. Everywhere else we aren't doing anything like

```ts
if(component === InputComponent.TextArea){
  // do something with the config 
}
```

Instead we are just having each Component figure that out for itself. And at the component level we can be fully sure what the type is, since we only define the component for one component anyway.

https://github.com/pubpub/platform/blob/38afca31b59988735895e003dadcdc7bb004defe/core/app/components/FormBuilder/ElementPanel/ComponentConfig/Checkbox.tsx#L8

So I thought it best not to introduce more esoteric typescript into the codebase if we aren't going to benefit from it!

packages/schemas/src/schemaComponents.ts
allisonking: thank you both, this makes a lot of sense! 🧠

Issue #705, opened 2024-10-09
Form rich text not rendering


gabestein:
## What went wrong, step-by-step?
1. Loaded a form
2. Did not render HTML parts of forms. Instead, rendered tags e.g. `<p>This is my text</p>.`

## What did you expect to happen?
Render as HTML

We had previously fixed this in #660, seems to have regressed.

## What URL can we use to see the issue?
https://app.pubpub.org/c/ipop/public/forms/invited-review/fill

It seems possible this is related only to existing forms? I recreated the paragraph block and it still had the issue. Have not tried creating a new form yet.

### Who reported this? (If not you)

PR #707, opened 2024-10-09
Fix markdown rendering when form is creating a pub

allisonking:
## Issue(s) Resolved

Closes https://github.com/pubpub/platform/issues/705

On form create, there is no pub ID. Before, the form just wouldn't render any structural markdown. Now, it renders it, but it won't be able to handle the directives that reference a pub.

## Test Plan
* Create a form with structural paragraph and put markdown in it
* Visit the form's fill page i.e. http://localhost:3000/c/croccroc/public/forms/review/fill
* The markdown should render
* The markdown should also render if you pass a pubId param
* Adding a directive like `:link[view]{field="croccroc:url"}` should work in the URL with the pub id, but should render nothing in the URL without the pub id.

Without pub ID
image

With pub ID
image

### DIFF
--- a/core/app/c/[communitySlug]/forms/[formSlug]/edit/page.tsx
+++ b/core/app/c/[communitySlug]/forms/[formSlug]/edit/page.tsx
@@ -1,3 +1,4 @@
+import dynamic from "next/dynamic";
 import { notFound } from "next/navigation";
 
 import type { CommunitiesId } from "db/public";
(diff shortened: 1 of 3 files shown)


core/app/c/[communitySlug]/forms/[formSlug]/edit/page.tsx
```diff
@@ -14,7 +15,11 @@ import { getForm } from "~/lib/server/form";
 import { getPubFields } from "~/lib/server/pubFields";
 import { ContentLayout } from "../../../ContentLayout";
 import { EditFormTitleButton } from "./EditFormTitleButton";
-import { FormCopyButton } from "./FormCopyButton";
+
+const FormCopyButton = dynamic(
```
Comment (allisonking) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/page.tsx:19]: unrelated to the issue this PR closes, but this was giving errors in the logs because `FormCopyButton` checks for `window` and the server doesn't like that

3mcd (approved):
Looks good, and impressive turnaround time on the fix!

PR #710, opened 2024-10-10
dev: run e2e tests before deploying

tefkah:
- **dev: make e2e tests run on promotion to main**
- **fix: make deploy wait for e2e, duh**

We should probably be good children and eat our vegetables and just run e2e tests on promotion to main even though we require you to run be up to speed with main anyway.


kalilsn (approved):
hate to say it but i think you are right!

PR #711, opened 2024-10-10
Run autosave even when there are invalid values

### LINKED ISSUE #692: External user form not saving new or updated progress

1. External user enters content into a form
3. External user refreshes tab, not saving/submitting
4. Previously entered content is not saved

Progress would save

## Optional

It looks like this was disabled by something committed last week - previously entered data from _before_ whatever the commit is saved, but new data (or updating previously saved content) isn't.

https://github.com/user-attachments/assets/57f10a88-60c8-472e-bf85-c97db79c18eb

David/Unjournal

kalilsn:
## Issue(s) Resolved
#692 

Previously our autosave handler wouldn't run if there were form errors. I've now modified the update fill form to run that handler on errors, but exclude all invalid values. For extra caution, I've also made our update function on the backend do the same thing (conditionally) so that valid pub values will still be updated even if some fail backend validation.


core/lib/server/pub.ts
```diff
@@ -410,14 +410,14 @@ export const createPubRecursiveNew = async <Body extends CreatePubRequestBodyWit
 		return Boolean(value);
 	});
 
-	const validated = validatePubValuesBySchemaName({
+	const validationErrors = validatePubValuesBySchemaName({
 		fields: filteredFields,
 		values: body.values,
 	});
 
 	// TODO: this should throw instead, aborting the transaction
```
kalilsn, line 418: @tefkah this comment was already taken care of, right? i can delete it?

3mcd (approved):
Looks and works great!

Issue #445, opened 2024-07-23
Form builder form keeps track of whether there are changes to save


kalilsn:
## Motivation

- So users don't accidentally lose lots of work on forms by navigating away before saving.

## Requirements
- Save button is disabled when no changes have been made to the form (will require refactor to put contentlayout below FormBuilderContext)
- Add dialog warning you before navigation if there are unsaved changes
- Optional: Only the dirty fields are sent to the backend, so we can minimize the db changes

## Mockups (if available)
Mockups exist for both the dialogue and button states
https://www.figma.com/design/SqOTMElJn7fuPljfQL1gIY/PubPub-Platform?node-id=408-15402&m=dev
https://www.figma.com/design/SqOTMElJn7fuPljfQL1gIY/PubPub-Platform?node-id=440-40641&m=dev

PR #714, opened 2024-10-10
Add unsaved changes warning to form

3mcd:
## Issue(s) Resolved

Kind of related to #692. These changes would at least notify users if they would lose progress.

Also partially related to #445, in that it introduces a general-purpose React hook for warning about unsaved changes. The general strategy here can be used on the form builder as well.

This PR changes the external form fill page to notify the user about unsaved changes when navigating. It uses the standard `beforeunload` method of notifying the user in a blocking manner, giving them the opportunity to cancel navigation.

The PR accounts for autosave by clearing the "dirty" flag from the form after an auto-save occurs.

I also change the form to only submit _changed_ values which can potentially make Pub editing w/ forms a bit snappier.

## Test Plan

1. Open a form fill page.
2. Change a value. Attempt to navigate or reload the page. You should be prompted by the browser that you have unsaved changes. Hit cancel to stay on the current page.
3. Wait for an auto-save to trigger. Navigate or refresh again. You should not be prompted.

### DIFF
--- a/core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx
+++ b/core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx
@@ -6,7 +6,7 @@
  */
 import type { Static } from "@sinclair/typebox";
 import type { ReactNode } from "react";
-import type { FieldValues, SubmitErrorHandler } from "react-hook-form";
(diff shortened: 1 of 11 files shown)


allisonking:
Hm it looks like I'm still getting the prompt even after an autosave (see vid) 

https://github.com/user-attachments/assets/b51882f0-85e4-407c-9a44-ca001753b6f4

allisonking (approved):
Tried again on a reset db and things work for me now! 🎉

PR #716, opened 2024-10-15
Add role to permission and community-level permission

### LINKED ISSUE #672: Add role column to permission model
## Motivation

Based on the requirements defined in #594 and #595, a community member's role does not imply the level of access to resources/actions granted by permissions. Instead the permission itself conveys a role, which is not something that our permission model captures right now.

## Requirements

- Adds permissions to community, stage, and pub
- Each permission should carry a role of contributor, editor, or admin
- Includes proposal of how to move forward on migrating existing role structure iteratively into new permissions structure

## Acceptance Criteria

- Add a "role" column to the permissions table.
- Similarly to the members table, this new column stores a value of `"contributor" | "editor" | "admin"`

kalilsn:
## Issue(s) Resolved
#672
This is quite a simple change to the schema for now: adding the role and connecting permissions to communities so we can have a community level permission. I've made it optional because there are existing permissions (we use them for forms only at the moment). In the next migration, we can set an appropriate role for those permissions and make it non-nullable.

As a next step, I'd like to suggest that we delete the memberships table, and replace it with a community membership table (or just rename it). I also believe we can get away with no permissions table at all, and just join directly from the users table to our more granular permissions tables for form permissions, community permissions, stage permissions, and pub permissions. I'll add a diagram soon for a clearer visualization of this.


kalilsn:
Closing this since I'm going to do a larger migration which will overwrite all of this anyways: #718

Issue #656, opened 2024-09-25
Community Membership updates


gabestein:
## Motivation

- So admins can see all members, even if stage/pub specific, in one place
- So community permissions are consistent with 

## Requirements

- Update community Member UI to show global community members as well as members who have permissions only in certain stages/pubs/groups.
- Permissions behavior:
   - Contributor: 
      - Can see, edit, and delete Pubs they have been directly added to (e.g. not children/related, unless they have direct or stage access).
      - Can see Stages they have been directly added to, and the name of stages any Pubs they have been added to are in.
      - Can create Pubs in Stages they have been directly added to
      - Can create child/related Pubs of Pubs they have been directly added to
   - Editor: 
      - Can see, edit, and delete all Pubs, including children/siblings
      - Can see all Stages
      - Can move, assign, and run actions on all Pubs in all Stages
   - Admin: same as editor, plus: 
      - Can edit community settings
      - Can see/manage all stage/action settings
      - Can see/manage workflow settings
      - Can add/remove members to/from the community, pubs, or stages. (If the user does not exist, admin can invite them via email. The user receives an invitation to join PubPub as a result, with a message specifying the stage they have been invited to and a redirect to the stage configuration they have been added to.) 
- Email behavior
   - When a user is added to a pub from the UI, they receive an email notifying them that they have been added to that pub, with a link to visit it.

ktyjscott:
Member Landing page and Member detail mocks: https://www.figma.com/design/SqOTMElJn7fuPljfQL1gIY/PubPub-Platform?node-id=1763-28204&t=k6TrYIzBG2F4KzWs-1

One lingering question on these mocks from dev: 1. Should admins be able to edit user info? e.g. name, email.

Issue #595, opened 2024-09-09
Add basic Stage membership UI


gabestein:
## Motivation

So admins can add, remove, and change permissions of Members in stages

## Requirements

- Ability to add/remove existing community members to a stage as an admin, contributor, or editor in the stage member tab, and invite new ones.
  - Add "add member" button to member tab of stage management, identical to "members" page add member button
  - Only admins should see the button
  - Should repurpose the existing member add UI, e.g.:
    - When clicked, shown a popup to search by email
    - If user found, can select permission for the stage and add it
    - If no user found, sends an invitation. That invitation should use slightly different language, e.g.: `You have been invited to be a [role name] of the [stage name] stage in [community name].`
- Ability to see existing members and their role, and delete them in the member tab of the side panel
  - Only admins should be able to see the delete button
  - Can just be the current member list, plus their role underneath their name and an x to delete them, similar to the current pub type editor

## Acceptance Criteria
- If user is not already a member of the community, only adds a membership to the stage with the selected role — not an additional community-level role
- Does not implement permissions/querying, that's separate cards, but should add/remove the correct permissions
- Does not need to show cascading permissions yet, we can add that later

## Mockups (if available)
![Image](https://github.com/user-attachments/assets/a37ee106-563a-494b-af4b-c7d6cfdb5dd6)

<details>
  <summary>Old Requirements rolled into other tickets</summary>
   
- Permissions behavior:
   - Contributor: 
      - Can see and edit Pubs directly in the stage (e.g. not child/related pubs not in the stage they don't have direct access to)
      - Can create new Pubs in the stage (they become direct members of these pubs)
      - Can create stageless child/related Pubs of Pubs in the stage (they become direct members of these pubs)
      - Can run actions/move/assign on Pubs in the stage
   - Editor: same as contributor, plus: 
      - Can see and edit children/related pubs of Pubs in the stage, even if child/related Pubs are not in the stage
   - Admin: same as editor, plus: 
      - Can edit stage configuration/actions/etc.
      - Can delete pubs in the stage
      - Can add members to stage as contributor, editor, or admin. (If the user does not exist already, they can be invited via email. The user receives an invitation to join PubPub as a result, with a message specifying the stage they have been invited to and a redirect to the stage configuration they have been added to.)
- Email behavior: 
  - When a user is added to a stage from the UI, they receive an email notifying them that they have been added to that stage, with a link to visit it.

## Acceptance Criteria

[...]

PR #723, opened 2024-10-15
Add privileges enum

### LINKED ISSUE #670: Define Pub and Stage membership privileges
## Motivation

To have a clear definition of all of the privileges that can be granted to users at the pub and stage levels, independent of role.

## Requirements

- Using the requirements from #594 and #595, define (in code) the complete set of actions that can be taken for each object.

## Acceptance Criteria

- Create a set of constants, e.g. one or more enums, that describes all of the authorized actions that can be taken on pubs and stages.

For example, an enum for pubs might look something like:

```ts
enum PubPrivilege {
  move
  create
  addPubField
  removePubField
  updatePubValue
  ...
}
```

kalilsn:
## Issue(s) Resolved
#670 
Adds an enum with every action in platform that we might need to check a user's authorization for before completing. Some of these actions will probably be condensed into one as we build this out, such as `createPubType` and `editPubType` and `removePubType` becoming just `managePubType`. The one place I haven't done this is for `updatePubValues` because we aren't yet consistently showing the difference between an existing but empty value, and a null row in the values table.


core/lib/permissions.ts
```diff
@@ -18,3 +18,39 @@ export const getPubUsers = (permissions: PermissionPayload[]) => {
...
+	createPubWithForm,
+	editPubWithForm,
+	createPubField,
+	archivePubField,
+	editPubField,
+	createPubType,
+	editPubType,
+	deletePubType,
+	createStage,
+	deleteStage,
```
tefkah, line 37: should we also have `editStage`, for changing the name? (and possibly other stage level configuration in the future?)

core/lib/permissions.ts
```diff
@@ -18,3 +18,39 @@ export const getPubUsers = (permissions: PermissionPayload[]) => {
...
+	addActionToStage,
+	editAction,
+	removeActionFromStage,
+	addPubMember,
+	addStageMember,
+	addFormMember,
+	addCommunityMember,
+	createMemberGroup,
+	addMemberToGroup,
+	removeMemberFromGroup,
```
tefkah, line 49: should we have `editMemberGroup` as well, for changing e.g. the name of a group after it's been created?

core/lib/permissions.ts
```diff
@@ -18,3 +18,39 @@ export const getPubUsers = (permissions: PermissionPayload[]) => {
...
+	deleteStage,
+	addMoveConstraint,
+	removeMoveConstraint,
+	addActionToStage,
+	editAction,
+	removeActionFromStage,
+	addPubMember,
+	addStageMember,
+	addFormMember,
+	addCommunityMember,
```
tefkah, line 46: should these have their `removeXMember` counterpart as well?

core/lib/permissions.ts
```diff
@@ -18,3 +18,39 @@ export const getPubUsers = (permissions: PermissionPayload[]) => {
 	}
 	return users;
 };
+
+enum Privileges {
+	movePub,
+	createPub,
+	viewPub,
```
tefkah, line 25: `archive/deletePub` as well right?
also thanks for including a separate `viewPub` privilege. do you think we should also add this to other places? for instance, in #656 @gabestein  describes the `contributor` role for a stage as 

> "Can see Stages they have been directly added to, and the name of stages any Pubs they have been added to are in".

this implies that they can only see the stage, not edit it (which i think is good). So would we need a separate `viewStage` privilege, or do you that is already captured some other way?

core/lib/permissions.ts
```diff
@@ -18,3 +18,39 @@ export const getPubUsers = (permissions: PermissionPayload[]) => {
...
+	editAction,
+	removeActionFromStage,
+	addPubMember,
+	addStageMember,
+	addFormMember,
+	addCommunityMember,
+	createMemberGroup,
+	addMemberToGroup,
+	removeMemberFromGroup,
+	addCommunity,
```
tefkah, line 50: There is currently no functionality for it yet, but we should probably have `editCommunity` as well right, for changing community settings?

tefkah (commented):
Do you have any thoughts on splitting this up per domain, e.g. instead of creating an enum, creating an object of enums that's something like

```ts
enum PubPrivliges {
	edit = 'edit',
	create = 'create',
	view = 'view',
	archive = 'view'
}

const privileges = {
	pub: PubPrivileges
} as const
```

That way you could do 
```ts
priviliges.pub.edit
```

Might be slightly more ergonomic, but I don't really feel that strongly about it.

core/lib/permissions.ts
```diff
@@ -18,3 +18,39 @@ export const getPubUsers = (permissions: PermissionPayload[]) => {
...
+	createForm,
+	editForm,
+	archiveForm,
+	createApiToken,
+	revokeApiToken,
```
tefkah, line 55: One thing I would like to ask: could you make this enum use strings instead? Since we aren't doing any bit masking or anything, I don't see much value to leaving the enum members as integers, and have found that often kind of confusing.

That way `Privileges.movePub === "movePub"`, rather than `Privileges.movePub === 0`

This would entail this.
```suggestion
    movePub = "movePub",
    createPub = "createPub",
    viewPub = "viewPub",
    updatePubValues = "updatePubValues",
    createRelatedPub = "createRelatedPub",
    createPubWithForm = "createPubWithForm",
    editPubWithForm = "editPubWithForm",
    createPubField = "createPubField",
    archivePubField = "archivePubField",
    editPubField = "editPubField",
    createPubType = "createPubType",
    editPubType = "editPubType",
    deletePubType = "deletePubType",
    createStage = "createStage",
    deleteStage = "deleteStage",
    addMoveConstraint = "addMoveConstraint",
    removeMoveConstraint = "removeMoveConstraint",
    addActionToStage = "addActionToStage",
    editAction = "editAction",
    removeActionFromStage = "removeActionFromStage",
    addPubMember = "addPubMember",
    addStageMember = "addStageMember",
    addFormMember = "addFormMember",
    addCommunityMember = "addCommunityMember",
    createMemberGroup = "createMemberGroup",
    addMemberToGroup = "addMemberToGroup",
    removeMemberFromGroup = "removeMemberFromGroup",
    addCommunity = "addCommunity",
    createForm = "createForm",
    editForm = "editForm",
    archiveForm = "archiveForm",
    createApiToken = "createApiToken",
    revokeApiToken = "revokeApiToken",
```

core/lib/permissions.ts
kalilsn: Yes to deletePub (maybe archive in the future if we build that).

For stage viewing, I had been thinking that a `viewPub` privilege would be enough, since it's the pubs in the stage you want to see, not the stage, but I think you're right!

core/lib/permissions.ts
```diff
@@ -18,3 +18,39 @@ export const getPubUsers = (permissions: PermissionPayload[]) => {
 	}
 	return users;
 };
+
+enum Privileges {
+	movePub,
+	createPub,
+	viewPub,
+	updatePubValues,
+	createRelatedPub,
```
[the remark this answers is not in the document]
kalilsn: It was the former, based on https://github.com/pubpub/platform/issues/595 saying that editors can create/edit related pubs. I think connecting an existing pub would fall under `updatePubValues`. But tbh, depending on how we implement it, it might not be different from `createPub` + `updatePubValues`. I expect to make some changes to this list pretty quickly once we have the actually write the queries for these.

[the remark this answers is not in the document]
tefkah: I think it makes sense to have separate capabilities for that then, i can see allowing someone to create a child pub/related pub, but not create any pub.

core/lib/permissions.ts
kalilsn: good point. and i think that can encompass all of the actions/rules config too.

core/lib/permissions.ts
kalilsn: good point!

kalilsn:
> Do you have any thoughts on splitting this up per domain, e.g. instead of creating an enum, creating an object of enums that's something like
> 
> ```ts
> enum PubPrivliges {
> 	edit = 'edit',
> 	create = 'create',
> 	view = 'view',
> 	archive = 'view'
> }
> 
> const privileges = {
> 	pub: PubPrivileges
> } as const
> ```
> 
> That way you could do
> 
> ```ts
> priviliges.pub.edit
> ```
> 
> Might be slightly more ergonomic, but I don't really feel that strongly about it.

I considered this a bit but I decided it was premature and that we would figure out the best way to split it up once we were actually implementing the different privileges and cascading. I will keep this in mind though!

3mcd (approved):
Looks good to me. I like that they are granular even though we could condense them. I think keeping them granular will make the code a bit easier to understand and maintain, as `createPubType && editPubType && removePubType` is much more obvious than `managePubType`

PR #725, opened 2024-10-16
Add unsaved changes warning to form builder

### LINKED ISSUE #445: Form builder form keeps track of whether there are changes to save

3mcd:
## Issue(s) Resolved

Resolves #445 

This PR changes the form builder to prompt the user (in a blocking manner) about unsaved changes when navigation occurs.

The top-level form builder, structural element editor, and input component editor all received a `useUnsavedChangesWarning` hook call that triggers the default browser unsaved changes prompt when the form state is dirty (modified).

This means that the warning prompt is triggered:
- if you modify a child (structural/input) form without saving the changes
- if you modify a child (structural/input) form and save the changes, but do not save the form
- if you modify the form itself (e.g. add/delete/reorder elements) without saving it

The "save" button within the structural element editor, input component editor, and top-level form builder were changed to become disabled/uninteractive when its respective form is dirty.

When a save button is pressed, its form state is no longer dirty, so it goes back to the disabled/uninteractive state.

## Test Plan

1. Open the form builder.
2. Refresh the page. No prompt.
3. Change the form structure by adding a field, deleting a field, or reordering existing fields.
  - The top-level save button (top right) should become interactive.
  - Refresh the page without saving the form. The prompt should appear.
4. Re-configure an input component or structural element.
  - The input component or structural element form's save button should become interactive.
  - The top-level save button should remain uninteractive.
  - Refresh the page without saving the child form. The prompt should appear.
5. Repeat step 4, but save the child form without refreshing the page.
  - The input component or structural element form's save button should become uninteractive.
  - The top-level save button should become interactive.
  - Refresh the page. The prompt should appear.

Note that the top-level "save" button (top right) is disabled because the user has not applied the changes from the inner form to the outer form yet:

image

### DIFF
--- a/core/app/c/[communitySlug]/forms/[formSlug]/edit/page.tsx
+++ b/core/app/c/[communitySlug]/forms/[formSlug]/edit/page.tsx
@@ -26,10 +26,14 @@ const getCommunityStages = (communityId: CommunitiesId) =>
 
 export default async function Page({
 	params: { formSlug, communitySlug },
+	searchParams: { isDirty },
 }: {
 	params: {
 		formSlug: string;
 		communitySlug: string;
+	};
+	searchParams: {
+		isDirty: boolean;
 	};
 }) {
 	const { user } = await getPageLoginData();
@@ -70,7 +74,7 @@ export default async function Page({
 				<div className="flex items-center gap-2">
 					<FormCopyButton formSlug={formSlug} />
 					{/* <ArchiveFormButton id={form.id} className="border border-slate-950 px-4" />{" "} */}
-					<SaveFormButton form={formBuilderId} />
+					<SaveFormButton form={formBuilderId} disabled={!isDirty} />
 				</div>
 			}
 		>
--- a/core/app/c/[communitySlug]/forms/[formSlug]/edit/page.tsx
+++ b/core/app/c/[communitySlug]/forms/[formSlug]/edit/page.tsx
@@ -26,14 +26,14 @@ const getCommunityStages = (communityId: CommunitiesId) =>
 
 export default async function Page({
 	params: { formSlug, communitySlug },
-	searchParams: { isDirty },
+	searchParams: { unsavedChanges },
 }: {
 	params: {
 		formSlug: string;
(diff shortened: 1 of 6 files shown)


kalilsn:
This is definitely an improvement, so I'm comfortable merging it as is if you want to. But I have noticed a few things that we may want to address:
- Clicking "Delete" on an element and then restoring it doesn't disable the save button
- Changing the form's access type doesn't disable the save button
- Reloading/closing/navigating by editing the URL directly all trigger the warning. But clicking a link in the left nav doesn't!
- Even locally there is a small but noticeable delay after clicking the form panel save or delete on an element when that updates the search params
- Extremely minor: If you edit an element, click the panel save button (not the form save button), then edit the element back to its original contents and then save again, the unsaved changes warning will still be triggered.

3mcd:
@kalilsn Nice catches. The first and fifth points are definitely surprising. Perhaps it has something to do with `useFieldArray` marking fields/the form as "dirty" even if the values match. Maybe they are comparing the default and current arrays by reference to determine "dirtiness"?

W.r.t the sidebar navigation issue -- we discussed this and decided to handle in a follow-up PR.

The search parameter delay is also expected, even if undesirable, because a solution that doesn't rely on search parameters would require a bigger refactor due to the way the save button is rendered outside of the form builder context.

I think I will merge this PR as-is, then address the access type issue in a commit in #714.

Issue #706, opened 2024-10-09
Add multivalue input option to forms


gabestein:
## Motivation

- So admins can ask users to fill out multiple, unconstrained choices in a form (e.g. a list of keywords describing a journal article)
- So community members can edit pubs in the (internal) default pub editor

## Requirements

- Allows the multivalue component to be selected as an option for array(string) and array(number) fields in the form builder
- Renders the multivalue component in public forms, when selected
- Uses the multivalue component for editing array(string) and array(number) fields in the (internal) default pub editor

## Acceptance Criteria

For the form builder:
- Label
- Description
- Default placeholder for the multivalue entry input: "Type the value and hit enter"
- No options

For form renderer:
- If field type is array(number), rendered input type should be number for browser-side validation

PR #727, opened 2024-10-17
Multivalue input in form builder, pub editor, and form fill page

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/706

Adds the multivalue input component to various parts of the app:
* Form builder for `StringArray` and `NumericArray` fields
* Form fill page if the multivalue input component was chosen
* The pub update/create form

## Test Plan
* Add a pub field that is either of type `StringArray` or `NumericArray`
* Add to a form and select the multivalue input field. You should be able to add a label/description if you want.
* Check that the field renders and works as you'd expect it to in the form fill page
* Add the same field to a pub type
* Try to create/edit a pub of this type. You should be able to use the multivalue input component for the field you added

### Form builder
image

### Fill page
image

With field disabled
image

### Pub create/edit
image
image

### DIFF
--- a/core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
+++ b/core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
@@ -19,6 +19,7 @@ import { useUnsavedChangesWarning } from "ui/hooks";
 import { ImagePlus } from "ui/icon";
 import { Input } from "ui/input";
 import { Label } from "ui/label";
+import { MultiValueInput } from "ui/multivalue-input";
 import { RadioGroup, RadioGroupItem } from "ui/radio-group";
 import {
 	Select,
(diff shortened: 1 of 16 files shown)


core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
```diff
@@ -172,7 +188,7 @@ const ComponentSelect = ({
 								onChange(c);
 							}}
 						/>
-						<div className="flex h-[124px] w-full flex-col justify-between rounded-lg border bg-card text-card-foreground shadow-sm peer-checked:border-2 peer-checked:border-ring peer-checked:outline-none">
+						<div className="flex h-32 w-full flex-col justify-between rounded-lg border bg-card text-card-foreground shadow-sm peer-checked:border-2 peer-checked:border-ring peer-checked:outline-none">
```
allisonking, line 191: made these boxes a little bigger since otherwise the multivalue input looked super squished inside

PR #728, opened 2024-10-17
Add new membership tables

kalilsn:


core/prisma/migrations/20241017223037_add_new_membership_tables/migration.sql
```diff
@@ -0,0 +1,123 @@
...
+CREATE TABLE "community_memberships" (
+    "id" TEXT NOT NULL DEFAULT gen_random_uuid(),
+    "role" "MemberRole" NOT NULL,
+    "communityId" TEXT NOT NULL,
+    "userId" TEXT,
+    "memberGroupId" TEXT,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CHECK ( num_nonnulls("userId", "memberGroupId") = 1 )
```
kalilsn, line 20: We can't express this constraint in prisma, but it should probably be documented somewhere more discoverable than this specific migration

core/prisma/schema.prisma
```diff
@@ -280,19 +287,90 @@ enum MemberRole {
...
+  formMemberships      FormMembership[]
+  communityMemberships CommunityMembership[]
+  pubMemberships       PubMembership[]
+  stageMemberships     StageMembership[]
 
   @@map(name: "member_groups")
 }
 
+model CommunityMembership {
+  id            String       @id @default(dbgenerated("gen_random_uuid()"))
```
kalilsn, line 307: Even though we shouldn't need to use the id ever, we still need one because composite ids don't allow nullable columns and two of the three columns that would make up our composite id (communityId, userId, memberGroupId) are nullable.

kalilsn: Alternatively we could make userId and memberGroupId not null, and set them to `""` by default. But that feels worse than an unused column to me

core/actions/_lib/runActionInstance.ts
```diff
@@ -107,10 +107,19 @@ const _runActionInstance = async (
...
 		};
+		if (args.actionInstanceArgs) {
+			// Check if the args passed can substitute for missing or invalid config
+			const argsParsedAsConfig = action.config.schema.safeParse(args.actionInstanceArgs);
+			if (!argsParsedAsConfig.success) {
+				return err;
+			}
+		} else {
+			return err;
+		}
```
kalilsn, line 122: Before this change we were validating the config for every action run, even when parameters are supplied, meaning actions would fail even if all the required data was provided. Now you should be able to run an action with invalid config if you supply all the parameters it would need. This came up here because in my test, I create an email action and then run it from the pub details page, without configuring it. 

LMK if you think this is a bad change. I mostly did it because it was easier than configuring the action in my test. But I also think it's an improvement for users, since we don't actually prevent you from saving an action with invalid config, so it's easy to encounter this situation.

kalilsn: I also think we should change how actions work a bit so that they're not individually responsible for implementing the config/parameters merging, which this would be part of

core/prisma/migrations/20241017223037_add_new_membership_tables/migration.sql
```diff
@@ -0,0 +1,123 @@
...
+CREATE TABLE "community_memberships" (
+    "id" TEXT NOT NULL DEFAULT gen_random_uuid(),
+    "role" "MemberRole" NOT NULL,
+    "communityId" TEXT NOT NULL,
+    "userId" TEXT,
+    "memberGroupId" TEXT,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CHECK ( num_nonnulls("userId", "memberGroupId") = 1 ),
```
kalilsn, line 20: These check constraints ensure that we set exactly one of `userId` or `memberGroupId`. We can't express this constraint in the prisma schema, but it should probably be documented somewhere more discoverable than this specific migration.

core/app/(user)/signup/page.tsx
```diff
@@ -12,7 +12,7 @@ export default async function Page() {
 		return (
 			<div>
 				You are not allowed to signup for an account, or the link that you clicked is
-				invalid expired. Please try again
+				invalid or expired. Please try again
```
kalilsn, line 15: should they really try again?

core/prisma/seed/seedCommunity.ts
```diff
@@ -824,7 +827,24 @@ export async function seedCommunity<
 		});
 
 	const createdMembers = possibleMembers?.length
-		? await trx.insertInto("members").values(possibleMembers).returningAll().execute()
+		? await trx
+				.with("community_membership", (db) =>
+					db.insertInto("community_memberships").values(
+						possibleMembers.map((m) => ({
+							...m,
+							id: m.id as unknown as CommunityMembershipsId,
```
kalilsn, line 835: Here, and everywhere else that I've modified the membership write queries, I am ensuring that the IDs will be the same between the tables, which will make inserting data from `members` into `community_memberships` easier. In other places I just select the id from the first CTE to use in the later query, but since this code will all be deleted soon I took the easy route of generating the ids on the server and assuming there won't be a collision

core/app/(user)/communities/actions.ts
```diff
@@ -46,237 +47,49 @@ export const createCommunity = defineServerAction(async function createCommunity
...
-						{
-							pubId: eb.selectFrom("new_pubs").select("new_pubs.id"),
-							fieldId: fields.find((field) => field.slug === `${slug}:content`)!.id,
-							value: '"LONG LIVE THE SLUGS"',
-						},
-					]),
-				{ communitySlug: slug }
-			).execute();
-			await Promise.all([permissionPromise, moveConstraintPromise, createPubPromise]);
-		}
```
kalilsn, line 277: Adding all this seed data made sense when we had no way to create a lot of these things in the app. But now that we're actually using the button in prod it's kind of annoying, especially because some of the things created here can't be deleted. So I removed it for now, with the assumption that we'll build a few specific templates of seed data into this page in the future.

That does slow down the E2E tests a bit, but I also think it's good for them to exercise the pub/field/type/stage crud interfaces

core/prisma/migrations/20241017223037_add_new_membership_tables/migration.sql
tefkah: yeah shame prisma just cant do that. I think a comment on the table pointing to this file would be good enough!

`drizzle` seems to have added check constraints recently... (https://orm.drizzle.team/docs/indexes-constraints#check)

May be worth considering using them for our migrations instead at some point, i've grown a bit annoyed at the constraints (hah) of prismas model.

core/app/(user)/communities/actions.ts
```diff
@@ -46,238 +47,49 @@ export const createCommunity = defineServerAction(async function createCommunity
...
+					userId: user.id,
+					communityId: eb.selectFrom("new_community").select("new_community.id"),
+					role: MemberRole.admin,
+				}))
+				.returning("id"),
+			{ communitySlug: slug }
+		).executeTakeFirstOrThrow();
 		revalidatePath("/");
 	} catch (error) {
+		if (isUniqueConstraintError(error) && error.constraint === "communities_slug_key") {
```
kalilsn, line 86: It would be nice to eventually generate these constraint names with kanel so that this line of code would cause a type error if that constraint were ever deleted or renamed. The same thing would also be useful for queries were we refer to the constraint name in an `on conflict (<constraint name>)` expression

core/prisma/migrations/20241017223037_add_new_membership_tables/migration.sql
kalilsn: added the comment! 

i'd love to replace prisma with something that lets us express all of our constraints in the schema. do you know if drizzle can generate/manage partial indices (an index with a where clause)?

core/prisma/schema.prisma
tefkah: I see absolutely no problem with giving them ids

core/app/(user)/signup/page.tsx
tefkah: No, I guess not. I wanted to capture something along the lines of "click here to send another email"/"ask the person who invited you to try again", but that's rather complex

core/app/components/MemberSelect/MemberSelectAddUserForm.tsx
```diff
@@ -48,21 +53,35 @@ export const MemberSelectAddUserForm = ({ email, community }: Props) => {
...
-					});
-
-					if (didSucceed(result)) {
-						toast({
-							title: "Success",
-							description: "User successfully invited",
+				return form.handleSubmit(
+					async ({ email, firstName, lastName }) => {
+						// eslint-disable-next-line no-console
+						console.error("running create user");
```
tefkah, line 59: i think these should be removed?

core/app/components/MemberSelect/MemberSelectAddUserForm.tsx
```diff
@@ -48,21 +53,35 @@ export const MemberSelectAddUserForm = ({ email, community }: Props) => {
...
+						const result = await runCreateUserWithMembership({
+							email,
+							firstName,
+							lastName,
+							role: MemberRole.contributor,
 						});
+
+						if (didSucceed(result)) {
+							// eslint-disable-next-line no-console
+							console.error("create user succeeded");
```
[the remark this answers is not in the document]
tefkah: maybe fine to leave in this one, but it probably shouldn't be an error

core/app/components/MemberSelect/MemberSelectClient.tsx
```diff
@@ -103,9 +103,13 @@ export function MemberSelectClient({
...
 			router.replace(`${pathname}?${newParams.toString()}`, { scroll: false });
 		}
-		setInputValue(value);
 	}, 400);
 
+	const onInputValueChange = (value: string) => {
+		setInputValue(value);
+		updateSearchParams(value);
+	};
+
```
tefkah, line 112: is neater yeah

core/lib/server/member.ts
```diff
@@ -68,17 +76,50 @@ export const getMembers = ({ communityId }: { communityId: CommunitiesId }, trx
...
+						userId: props.userId,
+						communityId: props.communityId,
+						role: props.role ?? MemberRole.editor,
+					})
+					.returning("community_memberships.id")
+			)
 			.insertInto("members")
-			.values({
+			.values((eb) => ({
+				id: eb.selectFrom("community_membership").select("id") as unknown as MembersId,
```
tefkah, line 91: im assuming this is a stopgap solution for later right

core/prisma/schema.prisma
```diff
@@ -11,6 +11,11 @@ generator client {
   previewFeatures = ["fullTextSearch", "tracing"]
 }
 
+generator dbml {
+  provider = "prisma-dbml-generator"
+  output   = "."
+}
```
tefkah, line 17: really cool!

core/lib/server/member.ts
kalilsn: writing to both tables is a stopgap solution that is removed in my later PRs! but if you mean there's a better way to make the types work here, i would love to know!

core/prisma/schema.prisma
```diff
@@ -295,19 +306,98 @@ enum MemberRole {
...
 
+model CommunityMembership {
+  id            String       @id @default(dbgenerated("gen_random_uuid()"))
+  role          MemberRole
+  community     Community    @relation(fields: [communityId], references: [id], onDelete: Cascade)
+  communityId   String
+  user          User?        @relation(fields: [userId], references: [id], onDelete: Cascade)
+  userId        String?
+  memberGroup   MemberGroup? @relation(fields: [memberGroupId], references: [id], onDelete: Cascade)
+  memberGroupId String?
```
tefkah, line 333: ideally the type generated by `kanel` would be a discriminated union having either the `userId` or the `memberGroupId`. I'm not sure that's possible without a custom prerender hook.
 Working with these mysterious nullable columns is always kind of annoying to me, so i'm very glad you added the `check` constraint!

core/app/components/MemberSelect/MemberSelectAddUserForm.tsx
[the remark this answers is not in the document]
kalilsn: oops, yes i'll leave this one but remove the others

core/prisma/seed/seedCommunity.ts
tefkah: nice solution to the problem of keeping the value of the `MemberSelect` field accurate!

packages/ui/src/editors/LexicalEditor.tsx
```diff
@@ -92,22 +92,29 @@ export const LexicalEditor = (
...
-				<FormControl>
-					<LexicalComposer initialConfig={initialConfig}>
-						<RichTextPlugin
-							contentEditable={
+				<LexicalComposer initialConfig={initialConfig}>
+					<RichTextPlugin
+						contentEditable={
+							<FormControl>
 								<ContentEditable
+									ariaLabelledBy={labelId}
```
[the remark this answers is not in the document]
tefkah: i feel like this knowledge is earned through some very painful debugging work hahah

tefkah (approved):
Even though this PR is quite large, I couldn't find a single thing to comment on to really change! All the caveats seem thoroughly thought out, sets up the following PRs very nicely, the visualization is very nice to have. Fantastic job Kalil.

The only thing I have to admit: I did not look at your integration test changes very thoroughly. I'm trusting you that the tests still roughly test all the same stuff, and as long as they pass im happy! If you feel unsure about it or would like me to take a closer look please let me know and i'll do that, but if you feel good about it I say send it!

core/prisma/schema.prisma
kalilsn: yeah it would be great if the generated type had a discriminated union. at the very least, i think we should be able to manually construct a type from the generated one that enforces our constraint. for now, i've just been omitting `memberGroupId` from the generated type since we aren't using it anywhere yet

core/app/(user)/signup/page.tsx
kalilsn: ok that makes sense! sorry if that sounded critical, I just meant that I didn't actually remember what they're supposed to do at that point. something to revisit as we build out the permissions!

PR #729, opened 2024-10-21
dev: add typesafe seed function and complex arcadia seed

### LINKED ISSUE #644: Spike: Journal Article Pub Type
## Motivation

Model an ideal "journal" Pub Type that looks like for Platform from the point of view of using Travis's editor demo to create the Pub and outputting a Pub to HTML, JATS, and PDF. Encapsulated in this motivation is the desire to avoid the shortcomings of the Legacy pub model, which put too much structure into "content," making it incredibly difficult to customize PDF and JATS output in particular. Instead, per the [Journal Article Type roadmap issue](https://github.com/orgs/pubpub/projects/46/views/1?pane=issue&itemId=72099324), we aim to map most legacy fields that could be used as metadata (and some blocks) to child/related pubs.

## Requirements

- A basic description of the structure, similar to the link roadmap issue
- A working example of a Pub in Platform that uses the proposed structure, including all needed pub types and fields, based off of [this pub](https://research.arcadiascience.com/pub/result-venom-capsids/release/1)
- An idea of how to transform Legacy prosemirror structure into new pubs, or a decision that it's too hard and we should use the existing Legacy prosemirror structure

## Acceptance Criteria

- Must be compatible with Travis' [editor demo](https://github.com/pubpub/platform/tree/tr/context-editor).
- Should start with the Journal Article Type roadmap [issue structure](https://github.com/orgs/pubpub/projects/46/views/1?pane=issue&itemId=72099324) as a jumping-off point, and justify major deviations from this structure
- Idea of what we have no concept of how to import/represent currenlty (e.g. node labels)

## Extra work

- Does not need to account for versions or release notes for now, but a good bonus if it attempts to. The default assumption is that we'll likely stuff versions into a child "version" pub whenever the pub is saved, in this way preserving all changes in a way that could be diffed, but there may be better options.
- Does not need to use actions to do any import/export. Can just use the API with local scripts that do all field mapping in code.
- Does not need to be renderable/editable in Travis's editor demo as long as Travis says the way we're conceiving of it can be compatible with it.
- To see if we can automatically import Pubs from Legacy PubPub into this structure reliably, or if we would need to flatten it some to accommodate Legacy Pubs
- A basic example of converting a Pub from the proposed structure to JATS
- An example of importing [this Pub](https://research.arcadiascience.com/pub/result-venom-capsids/release/1) from Legacy to Platform
- JATS output should validate against the [JATS4R recommendations](https://jats4r.niso.org/jats4r-validator/) for the [NISO JATS Article Publishing DTD v1.3](https://jats.nlm.nih.gov/publishing/1.3/)
- Legacy to Platform import should use the PubPub Legacy API to retrieve content from PubPub Legacy.

tefkah:
## Issue(s) Resolved

#644

This PR

- Adds a way to seed communities in a relatively easy and type-safe manner.
- Adds a complex example community seed for Arcadia Research.
- Adds the ability to create related pubs in `createPubRecursiveNew`

### To quickly get started

- `pnpm i`
- `pnpm dev:setup`
- `pnpm dev`

To run the tests you need to do

- `pnpm test:setup`
- `pnpm --filter core test`

### The Arcadia Community

I tried to make a community that matches a Legacy community quite closely. The reason I did the whole community rather than just the Article/Legacy Pub is that that was specified in the issue (https://github.com/orgs/pubpub/projects/46/views/1?pane=issue&itemId=72099324)

> Collections
>
> - Issue Collection (DOI, volume, number, title, etc.)
> - Tags
> - Journal (ISSN, etc.)
> - Site

Here I will go over all the items mentioned in the above issue, and comment how they are represented in the Arcadia community seed, and my opinion on how why I chose to model them in that way.

##### Basic pub fields

I propose that a Journal Article PubType should have at least the following fields:

- Title: Field of type `String`
- Slug: Field of type `String`. I think `slug`s should be something we build in the future however.
- PublicationDate: Field of type `Date`
- CreationDate: Field of type `Date`. We should make this part of core.
- Last Edited Date: Field of type `Date`. We should also make this part of core.
- PreviewImage: = Avatar in the seed, to model the name of the field in Legacy. Field of type `Image`.
- Description: Field of type `String`. Should probably be of type `HTML` or something, as it can be rich text in Legacy.
- Abstract: Field of type `String`. Should probably also be of type `HTML`.
- LatestContent: = PubContent in the seed. Field of type `String` for now, will be `RichText` in the future.
- License: Field of type `String`. Should be an enum in the future.
- BackgroundImage: Left out for now, can just be a field of type `Image`.
- BackgroundColor: Left out for now, can just be a field of type `String` for now, but we should have a `Color` type in the future.
- TextStyle: Not added here. Can be Field of type `String`. Should be an enum in the future.
- CitationStyle (including inline): Field of type `String`. Should be an enum in the future.
- ItemLabels: Left out for now. We could add something like "TableLabels", "FigureLabels", etc. If we really wanted to, but there's still too much uncertainty about how we want to handle rendering of Pubs for me to say "We should do it like this".
- ConnectionSettings: Left out for now. Could be a relation Field
- DOI: Field of type `URL`.

##### Relations to other pubs/children

> [!NOTE]
>  A note on relations and children.
> Currently most things that are modeled as a Pub are modeled as both a relation and a child.
> This is mostly because we cannot see relations in the UI yet, so they are added as children as well as a workaround.
> I think we should just use relations for everything in the future, see recommendation below.

- Downloads: A PubField referencing a `DownloadablePub` PubType with a field of type `File`. Currently I just implemented it for PDFs. The relationship does not really need to specify metadata.
- Versions: Left out for now. I think version should be a built in feature of PubPub, I think it's too tricky to get right, and it's something that almost any community would want for almost any pubtype. 
- Images: A PubType with an `Image` field of type `File`. The relationship does not really need to specify metadata.
- Citations: Citations can be relationships to a `Reference` PubType. I have currently modelled it as an `ExternalBookReference` and an `ExternalArticleReference`, both of which have some metadata.
    - The `Citations` field then contains the metadata to the `*Reference` field, currently just chapter information or 
- Footnotes: Should be part of the editor!
- LaTeX blocks: I think they should be part of the editor. @isTravis is a proponent of this being a PubType. I think this would lead to a wild proliferation of Pubs, and it's something we should "just" have in the editor imo.
- Code blocks: Should be part of the editor.
- Contributors: A PubField, which references an "Author" (or something else) PubType. The metadata is the role of the author in the context of the Pub.

- Issue Collection: 
- Tags: I think tags should probably be referenced from the thing that is tagging, rather than the other way around. That makes most sense from a user perspective imo (you want to add a tag to the thing you are editing, not go create a "Tag" pub and add "Journal Articles", "Issues", etc fields.)
- Journal (ISSN, etc.): Stragistforward, same as issue.
- Site: See the diagram below or the seed.
-  Connections: Just a pubfield with a reference to another "Journal Article" Pub. We should probably dissallow self-references in general. For non-trivial self-reference, we should probably cap the depth that we fetch relationships in and manually check during that fetch if we are referencing ourselves.

See here a hastily AI generated graph that may or may not help. The nodes are Pubs/PubTypes, in the body of the nodes it shows the non-relationship fields of the PubType and their Schema. The edges are the fields that are relationships, sometimes with some metadata.

```mermaid
graph TD
    Site["Site
    ---
    Title: String
    Description: String
    Slug: String"]
    Header["Header
    ---
    Background Color: String
    Text Color: String"]
    Banner["Banner"]
    HeaderNav["Header Navigation
    ---
    Title"]
    Footer["Footer
    ---
    Title: String
    Background Color: String
    Text Color: String"]
    FooterNav["Footer Navigation
    ---
    Title"]
    Journal["Journal
    ---
    Title
    DOI: URL
    ISSN: String
    Slug: String"]
    Issue["Issue
    ---
    Title: String
    ISSN: String
    DOI: URL
    Description: String"]
    Article["Journal Article
    ---
    Title
    Publication Date: Date
    Creation Date: Date
    Last Edited: Date
    Avatar: File
    Description: String
    Abstract: String
    License: String
    PubContent: RichText
    DOI: URL
    Inline Citation Style: String
    Citation Style"]
    Tag["Tag
    ---
    Title: String"]
    Author["Author
    ---
    Name: String
    ORCiD: URL
    Affiliation: String"]
    PDFDownload["PDF Download
    ---
    PDF: File"]
    Table1["Table 1
    ---
    Caption: String
    CSV: File"]
    Table2["Table 2
    ---
    Caption: String
    CSV: File"]
    Image["Pub Image
    ---
    Pub Image: File
    Caption: String"]
    ExternalBook["External Book
    ---
    Title: String
    DOI: URL
    Year: String"]
    ExternalArticle["External Journal Article
    ---
    Title: String
    DOI: URL
    Year: String"]
    HomeItem["Navigation Item: Home
    ---
    Title: Home
    External Link: URL"]
    LegalItem["Navigation Item: Legal
    ---
    Title: String
    External Link: URL
    Open In New Tab: Boolean"]
    CommentArticle["Journal Article: A Comment on Capsid Identification
    ---
    Title"]

    Site -->|Header| Header
    Site -->|Footer| Footer
    Site -->|Journals| Journal
    Header -->|Banner| Banner
    Header -->|Navigation| HeaderNav
    Footer -->|Navigation| FooterNav
    Journal -->|Issues| Issue
    Issue -->|Articles| Article

    subgraph article
    Article -->|Tag| Tag
    Article -->|Contributors: Original Draft Preparation| Author
    Article -->|Downloads| PDFDownload
    Article -->|Tables| Table1
    Article -->|Tables| Table2
    Article -->|Images| Image
    Article -->|Citations: Chapter 1| ExternalBook
    Article -->|Citations: pp. 34-54| ExternalArticle
    Article -->|ConnectedPubs| CommentArticle
    end
    HeaderNav -->|Navigation Items| HomeItem
    FooterNav -->|Navigation Items| LegalItem

    classDef default stroke:#333,stroke-width:2px;
```

### Issues we should tackle

- [ ] Add ability to fetch relations recursively from the DB and the API. I have a working branch for this, but it's only for relations, not both relations and children.
- [ ] We need the ability to designate certain fields, either on the Pub or PubType level (or both) as being the "Title" field.
- [ ] Similarly, we need the ability, in some way, to either designate a field as the "Main" field, meaning that it will be the one that is shown on the pub page, while treating the rest as metadata, or to specify how a pub is rendered on a Pub.
- [ ] Add the ability to create, modify, and see relations in the UI.
- [ ] We probably need some way to signal, on a PubType (and PubField), to the editor that certain Fields or PubTypes are "Citation Fields", or have other editor specific data. I don't think we can get away with Duck Typing things like "Math" or "Citations".
- [ ] In addition to the above, we probably want to have some core, built in PubTypes/Fields that we, in the editor, can rely on existing.

### Opinions about data structures moving forward

You can skip this if you are not interested in rants.

<details>
<summary> <h4> ✊ We should get rid of children</h3>
<p>I'm not really sure how to treat children. I personally really would only like us to have one kind of relationship, and since we need fields,
I think we should only have fields. I do see the use of a general bucket we can put Pubs into for the editor.
Below I have some very rough thoughts on children and relationships. Feel free to skim, because it's not very well written, my apologies.
</p>
</summary>

##### Children as the "default" container for relating Pubs in the editor 

The most important way add complex things to a Pub in the editor at the moment is by creating children.
By typing e.g. `/Image`, you can create a child of PubType `Image` and see it (and possibly edit it) in the editor.

This is nice UX, and I can see users making sense of this quite quickly. "Oh, to add an image, I just need to create an `Image` PubType first".
I can see that, say, not being able to add an `Image` to a Pub in the editor until you create both an `Images` Field and a `Image` PubType would be a lesser UX. You would also need to be able to specify where the `Image` would go, so you would need to do something like `/Images/Image`, which is kind of cringe.

I do think this makes sense though as we are editing the pub and it's fields using the context editor. The context editor I think shouldn't be that "magical", in the sense that it should work very differently from how you would normally edit a pub. I think the context editor should "just" be a very nice and convenient way to both edit the Pub and its field and integrate other Pubs into the text.
That to me means that if the editor is intented to allow you to create any kind of Pub inline, and you want to model your `Journal Article` to put your `Authors` in a `Contributors` field (which makes a lot of sense I think), then the editor should work with that and allow you to do something like `/Contributors/Author`. The editor should not force you to only be able to create `Contributors` or `Images` as children in the editor. I don't think it should treat children specially.

As a compromise, maybe we could always create a `children` field on a Pub, and by default, in the editor, if you create a Pub using `/Image` say, it would add it to the `children` field. We could also call it something different, like `Inline Pubs`.
That way, in other parts of the app, we could still treat all relations the same, which I think is important, see the below section.

##### Parent-Child is a just a default one-to-many relationship with some yet-to-be-elsewhere-implemented functionality

The only difference between parent-child and our other relationships is that we have predefined logic and UI to handle this special case, and that
other relationships could be many-to-many.
Otherwise, there is very little difference between our current children implementation and us creating a `children` field.

For the little benefit that children offer, they mostly have downsides compared to relations.

- We need sibling relations, therefore we are forced to have two models of doing intra-pub relationships.
  - This is confusing for users, annoying for us.
  - Since we would probably want to fetch both relations and children recursively in most cases, having two fields to do this for will almost certainly be much less performant, not to mention more complex.
- Children cannot hold any metadata, whereas relations can.
- Children can basically be anything, whereas relationships can indicate what kind of data is likely to be on a Pub.
  - "Children" can be any kind of Pub, hard to reason about a structure.
  - Relationships are at the very least a certain `PubField` with a `name`. If you see a field with the name `Contributors`, you have a pretty good idea of what kind of data is likely to be on a Pub in that field.
  - For relationships, we could add a `pubTypeId` field to the `PubField`, which could give users and us more information about what kind of data is likely to be on a Pub.
- The "relationship model" is likely to be a lot more familiar to users as that is how almost every other CMS works.

The only other usecase we have discussed for children is that children are "scoped" to a certain Pub.
What does this mean? Does it mean that Pubs which are children of a Pub cannot be references by any other Pub?
This might sound useful, as you might not want to pollute the global "scope" with a bunch of small "Math" pubs for instance (if we were to have Math blocks as Pubs).

However, I think this will ultimately just cause problems. Say you start your Community and create a `Monograph` pubtype.
You will give Pubs of this PubType children of PubType `Chapter` (you cannot communicate this properly in the first place, everyone working with `Monograph`s will just have to know that they will have children of type `Chapter`). You go on like this for a while, all is great.

Some time later you expand your offerings and create an `Edited Volume` pubtype. You realize it's possible a chapter could come be in multiple edited volumes, so you create a `Chapters` field on `Edited Volume`.
You are creating an `Edited Volume` and want to reuse one of the chapters you created for your `Monograph`. But, _SHOCK, GASP_, you cannot do this! Because that chapter is scoped to your `Monograph`!

This will also happen the other way around, i.e. you would not be able to add a `Chapter` to your `Monograph` if it's already referenced by an `Edited Volume`.

"But Thomas", I hear you say, "let's just not add that restriction." Then the only restriction is that a Child can only be a Child of one Pub, but can still be freely referenced by other Pubs.
But what does that even give us? We are then just left with an improverished relationship field with a special exception. If we really want such a special exception, we can just add that to our relationship model!
We could easily add a `one-to-one` relationship type, and then add the `ParentBook` field to `Chapter` with a `one-to-one` relationship to `Monograph`.

In conclusion, I think there is almost no benefit to having children. It creates a more complicated mental model for the user (as they need to remember which children go with which PubType), and could easily be replaced by a more sophisticated relationship model.

</details>

#### Math blocks, just like footnotes and code blocks, should not be Pubs, but be part of the editor

My thinking here is as follows: For "special" editor functionality, like being able to do `$x^2$` inline, we probably want that to be part of the editor, not a Pub. 
Consider the alternative: Inline math, and shortcuts for math, are done by referencing/inlining Pubs.

Say you have a blank community, with just one PubType and one field for `RichText`. You start typing. How is the editor to know that you are able to do math? If math editing is in any way accessible through some shortcut like typing `$` (which I think it ought to be), how is the editor to know that you are able to do that? How is the editor to know which `Math`-y PubType it should create? 

Options:
- We do duck typing, and you need to create a `Math` PubType which has some special `Math` field.
    - When typing `$`, which PubType is the editor creating? Where do you specify that?
    - We could specify this if we had typed PubFields. We could create a `MathBlocks` PubField, give it some property called `EditorProperty.Math` or something, and maybe only allow one of these to exist per PubType?
    - But PubTypes do not constrain the values a Pub can have, so you could still have a bunch of these on the same Pub.
    - So either we add _another_ field to the Pub, which points to the field it should use for math. Then the editor reads _that_ field, and then know what to do when you type `$`.
    - OR every single time you type `$`, you have to specify which Field/PubType you want to use for Math.

#### Relationship data should be stored on the value as relationship metadata. We should not use intermediate "Relationship" Pubs for this.

Take a `Contributor` field on a `Journal Article` PubType. My suggestion is that we point the `Contributor` field to an `Author` PubType, and we store things like `role`, `order`, and any other data that is not part of the `Author` PubType on the `Contributor` field as metadata on the relation.

We could do this differently. We could also create a `Contributor` PubType, which would have fields like `Role`, `Order`, etc, and an `Author` field which would point to an `Author` PubType. On the `Journal Article` PubType we would then have a `Contributors` field, which would be a relation to the `Contributor` PubType. 

| Approach | Pros | Cons |
| --- | --- | --- |
| Store relationship data on the value as metadata | - Relationship data is stored on the relationship value, not in an intermediate Pub.<br>- Minimal proliferation of Pubs. | - Would have to repeat the schema of the metadata if you create multiple fields pointing to the same PubType. |
| Use a separate Pub to store relationship data | - Easy have an understanding what metadata is needed.<br>- Would keep our data model quite minimal. | - Requires defining a lot of PubTypes.<br>- Requires creating a ton of Pubs that do not are not useful in their own right, might cause confusion in the I and slowdowns in the DB | 

I think the first approach is the better one.

## Community Seed functionality

The one defined here:
https://github.com/pubpub/platform/blob/9d823ac8c620f4d56488f2a3a846dfe231a1a520/core/prisma/seed/seedCommunity.ts#L358-L385

This function allows you to create a community with a set of pre-defined pub fields, pub types, users, stages, stage connections, pubs, and forms, in a type-safe manner.

### What was added to make this work?

I added the possibilty to create related pubs in `createPubRecursiveNew`, both like children, and both like values. See my comments in the PR about that.

### Usage

See the Arcadia example seed for an example of how to use the function.

https://github.com/pubpub/platform/blob/9d823ac8c620f4d56488f2a3a846dfe231a1a520/core/prisma/exampleCommunitySeeds/arcadia.ts

The crocroc one is slightly less complex:

https://github.com/pubpub/platform/blob/9d823ac8c620f4d56488f2a3a846dfe231a1a520/core/prisma/exampleCommunitySeeds/croccroc.ts

#### Kysely queries/steps

Do you think the Kysely queries/steps to generate the community are easy enough to understand?

#### Type complexity

Do you think the types are too complex? They are not very easy to read, but other than extracting some complex inline type manipulation out to separate types and adding comments, I'm not sure how to improve it.

By extracting complex inline types I mean something like turning this

https://github.com/pubpub/platform/blob/9d823ac8c620f4d56488f2a3a846dfe231a1a520/core/prisma/seed/seedCommunity.ts#L129-L131

into something like this

```ts
type PubTypeFieldNames<PT extends PubTypesInitializer<any>, PubTypeName extends keyof PT> = keyof PT[PubTypeName]

// ...
{
    [FieldName in PubTypeFieldNames<PT, PubTypeName>]?: /*..*/
}
```

Does not really make the logic that much easier, but at least it's slightly easier to read.

Please let me know if you think I should spend a bit more time on that!

Type tests might help as well. See for an example here:

https://github.com/pubpub/platform/blob/77d31e8541ef9394e5e85a7d0f2dacd83569ceee/core/lib/server/cache/cache.test.ts#L227-L238

#### TypeScript performance

Is using the function fast enough in VS Code? I.e. does typescript not slow down to a crawl?

## Notes


core/prisma/seed.ts
```diff
@@ -30,71 +33,82 @@ async function createUserMembers({
...
+		.with("new_users", (db) => db.insertInto("users").values(values).returningAll())
+		.insertInto("members")
+		.values((eb) =>
+			prismaCommunityIds.map((id) => ({
+				userId: eb.selectFrom("new_users").select("new_users.id").where("slug", "=", slug),
+				communityId: id as CommunitiesId,
+			}))
+		)
+		.returningAll()
+		.executeTakeFirstOrThrow();
```
tefkah, line 55: begone prisma

core/prisma/seed/register-stub.js
tefkah, line 1: these annoying stubs are necessary in order to be able to use `createRecursivePubNew` in the seed script. This stub is only called when running the seed script. This is really hacky and annoying, if you know of a better way to stub things let me know. 

For ref, we need to stub because `React.cache` and `unstable_cache` and `cookies` etc are not available outside of `next`.

packages/ui/src/icon.tsx
```diff
@@ -38,6 +38,8 @@ export {
 	Heading2,
 	Heading3,
 	History,
+	TextQuote,
+	BookOpenText,
```
tefkah, line 42:
```suggestion
```

packages/schemas/src/CoreSchemaWithIcons.ts
```diff
@@ -1,6 +1,7 @@
 import type { LucideIcon } from "lucide-react";
 
 import {
+	BookOpenText,
```
tefkah, line 4:
```suggestion
```

packages/schemas/src/CoreSchemaWithIcons.ts
```diff
@@ -9,6 +10,7 @@ import {
 	ImagePlus,
 	Link,
 	Mail,
+	TextQuote,
```
tefkah, line 13:
```suggestion
```

core/prisma/seed/seedCommunity.ts
```diff
@@ -0,0 +1,1102 @@
...
+	string,
+	{
+		relation?: true;
+		schemaName: CoreSchemaType;
+	}
+>;
+
+type PubTypeInitializer<PF extends PubFieldsInitializer> = Record<
+	string,
+	Partial<Record<keyof PF, true>>
```
allisonking, line 55: I don't really understand why this is `<Record<keyof PF, true>>`? what is the `true` for? I noticed the 

```
"Navigation Item": {
  Title: true,
  "External Link": true,
  "Open In New Tab": true,
  // nested navigation
  "Navigation Items": true,
},
```
in the seed script, is the `true` significant?

allisonking (commented):
the `seedCommunity` func is super nice!!! 

I tried running things locally but am running into some things. If I try to do a `pnpm run reset` in core, I get an error

```
Running seed command `tsx --require ./prisma/seed/register-stub.js prisma/seed.ts` ...
[16:03:11.221] ERROR: crypto is not defined
```

I get a similar error when trying to run the tests (about `crypto` not being defined). I also had to manually remove my supabase containers to not have things conflict... maybe I was supposed to do that a while ago 😬

core/prisma/seed/seedCommunity.ts
tefkah: Good question! No, it's not really significant, this just defines which fields belong to which pubtype.
This could have also been an array I guess, but I find in general that it's easier for typescript to do things with Objects. I also left it like this because maybe in the future we would add some configuration to a PubField in a PubType, for instance marking a PubField as the "title" field. This would not be that hard to add in this way.

core/prisma/seed/register-stub.js
```diff
@@ -0,0 +1,12 @@
+const Module = require("module");
+const path = require("path");
+
+const originalResolveFilename = Module._resolveFilename;
+
+Module._resolveFilename = function (request, parent, isMain, options) {
```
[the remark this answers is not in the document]
tefkah: Yes! This is necessary in order to stub out `React.cache`, `unstable_cache` and `server-only`. Basically before the script is run, this patches the way node resolves files, so that when it tries to load `next/cache`, or `react`, or any of the other things mentioned in that `if` statement, it gets the content of `stubs.js` instead, which defines most of those things as functions which return functions.

tefkah:
> the `seedCommunity` func is super nice!!!
> 
> I tried running things locally but am running into some things. If I try to do a `pnpm run reset` in core, I get an error
> 
> ```
> Running seed command `tsx --require ./prisma/seed/register-stub.js prisma/seed.ts` ...
> [16:03:11.221] ERROR: crypto is not defined
> ```
> 
> I get a similar error when trying to run the tests (about `crypto` not being defined). I also had to manually remove my supabase containers to not have things conflict... maybe I was supposed to do that a while ago 😬
> 
> the TS performance in vscode is doing just fine on my computer at least!

My guess is that you are running node 18? The problem is probably with me calling `crypto.randomUUID`, which was only added in [Node 19](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility). You should probably be running node 20!

allisonking:
> @allisonking My guess is that you are running node 18? The problem is probably with me calling crypto.randomUUID, which was only added in [Node 19](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility). You should probably be running node 20!

ah yup you are right! upgrading to 20 fixes that 🎉

allisonking (approved):
🤩 amazing!

Issue #720, opened 2024-10-15
Create Rich Text Pub Field type


gabestein:
## Motivation

- So users can add rich text (name tbd?) fields to their pubs, for use with our editor

## Requirements

- In the Field editor, users can create a Pub field with a Rich Text type

## Acceptance Criteria

- No typebox checking required
- Instead, field should be special cased such that it validates against prosemirror with the included schema for the editor — if too slow, could validate only against a top-level prosemirror structure -- e.g., that it is an object that contains a `doc` array.

Issue #722, opened 2024-10-15
Render rich text editor


gabestein:
## Motivation
- So users can edit in rich text

## Requirements 
- In the form builder and default form editor, rich text fields render the prosemirror editor
- Support for support only non-child/non-parent fields for now (e.g. bold, italic, links, etc.)

## Acceptance Criteria
- "Design" for the rich text input can be extremely basic at this point, more or less porting what Travis has already with few modifications (less the ability to change stylesheets. We can use just one default stylesheet.)
- Acceptable if the editor still includes all options travis demo'd but they just don't work, e.g. provides stubs for functions that request fields/pub types/attributes/etc.

allisonking:
### Approach
I'm making as few changes as possible within the `context-editor` package. I did have to add a simple `index.ts` in order for preconstruct to build the editor as a package that I could import elsewhere.

* I'm working on a separate branch https://github.com/pubpub/platform/tree/aking/722/import-richtext-editor which is the latest `main` with Travis' context editor branch merged in. The diff is tough because the editor branch is behind main, but you can see my changes [here](https://github.com/pubpub/platform/compare/main...aking/722/import-richtext-editor) and then click my commits at the bottom

### Log

1. Merging the two branches seems to break storybook for some reason. I thought it might be because we have, since the branch was branched, added our own storybook, so there might be different versions of storybook. I'm not sure though, but this does seem to be contained to storybook from what I can tell
![Image](https://github.com/user-attachments/assets/2b1b46e1-730b-48a8-999f-cdcbc31bb051)

3. We'll need to pass some sort of `getPubs` that can filter by a string. I've mocked this out for now, not sure how we would pass a server call with a filter to a client component?
4. `getByPubId` not used yet, though it's a required prop. leaving it out for now!
5. There are some logs about `window` not being defined which I'm guessing is because we are using the component in an app with server side rendering, not quite sure the right way to work around this
6. There are parts of the editor that are hard coded to use the `rd` community, so for now I'm making a local `rd` community just to smooth things along.
  * For example, getting the name of the pub is hard coded in the context editor as `rd:title` right now, probably need to use something like [`getPubTitle`](https://github.com/pubpub/platform/blob/main/core/lib/pubs.ts#L7)

7. The pop up menu renders kind of far away from the cursor here, but not in storybook

For now, things kind of work!
![Image](https://github.com/user-attachments/assets/79f09821-c60b-4c31-ab3b-ec7e38b3cd34)
[...]

Issue #721, opened 2024-10-15
Add rich text input component to forms (custom and internal default)


gabestein:
## Motivation
- So users can add and configure rich text fields in the form builder
- So rich text fields display by default in the internal default editor

## Requirements
- In the form builder, rich text fields can only use the "rich text" input component.
- In the default form editor, rich text fields automatically use the rich text input component.

## Acceptance Criteria
Options for form builder: 
- Label option
- Help text option (in this case, should show above the field)
- Required option

gabestein:
@tefkah -- mind estimating when you get online tomorrow?

tefkah:
I put 2 days. It might be faster, but I haven't really touched the form stuff much at all, so I'm expecting to run into a snag or two.

PR #730, opened 2024-10-21
Rich text pub field type

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/720

And some progress on https://github.com/pubpub/platform/issues/721

This makes two enum changes to add rich text to both `CoreSchemaType` and `InputComponent`. Many of the changes made here were what follows from adding these enum types (we already had good type checking in place that makes sure we handle cases for every enum!). 

Now you can add a pub field of "Rich Text" type, and you can add that to forms or pub types. You can save a form configuration for the rich text field, but we don't display it anywhere yet (that's the rest of https://github.com/pubpub/platform/issues/721, and is blocked by https://github.com/pubpub/platform/issues/722)

There is also my guess at validating a prose mirror field, though I don't have a way to test it out yet and am guessing it probably won't just work as-is!

## Test Plan
* Add a new pub field of RichText type
* You should be able to add it to a pub type, but creating a pub of this pub type won't render the rich text field yet
* You should be able to add it to a form and configure a "label" and "help text" but this won't render the rich text field in the fill form page yet

image
image


core/actions/_lib/validateFields.ts
```diff
@@ -28,8 +30,21 @@ export const validatePubValuesBySchemaName = ({
...
 		}
-		const jsonSchema = getJsonSchemaByCoreSchemaType(field.schemaName);
-		const result = Value.Check(jsonSchema, value);
+
+		let result = false;
+		// Rich text fields are a special case where we use prosemirror to validate
+		// as opposed to typebox
+		if (field.schemaName === CoreSchemaType.RichText) {
+			try {
+				(value as Node).check();
```
allisonking, line 39: I really don't know if this will work yet...

tefkah: I think, as the `value` will be a `JSON` value, we will first need to call `nodeFromJSON` on the `schema` @isTravis defines here: https://github.com/pubpub/platform/blob/tr/context-editor/packages/context-editor/src/schemas/index.ts

Not sure how you would do this in this PR though 😅. Maybe you extract the prosemirror `nodeFromJSON` + `Node.check` validation to a different file/function, and create a stub schema there for now with a comment indicating we will import it from the editor?

core/package.json
```diff
@@ -91,6 +91,7 @@
 		"openapi3-ts": "^4.1.2",
 		"pg": "^8.11.3",
 		"prosemirror-markdown": "^1.12.0",
```
tefkah, line 93: why did we already have `prosemirror-markdown` ??

allisonking: hmm looks like for the `pushToV6` action! 

https://github.com/pubpub/platform/blob/1f3217c7319d29356a160726076fc1a1e6a87778/core/actions/pushToV6/run.ts#L136-L138

core/lib/__tests__/contextEditor.test.ts
```diff
@@ -0,0 +1,28 @@
...
+	],
+	type: "doc",
+};
+
+test.each([
+	{ value: "", expected: false },
+	{ value: 1, expected: false },
+	{ value: {}, expected: false },
+	{ value: JSON_EXAMPLE, expected: true },
+])("validateAgainstContextEditorSchema", ({ value, expected }) => {
```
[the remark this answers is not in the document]
allisonking: never mind—let me make a better test!

core/package.json
3mcd: yea i used it to convert markdown to a prosemirror doc before creating the pub in v6

tefkah: ahh that makes sense

PR #731, opened 2024-10-22
fix: copy integration data to the correct folder

tefkah:
## Issue(s) Resolved

Static assets not properly resolved for integrations because they were copied to the wrong place in the dockerfile.

## Test Plan
Send it.


kalilsn (approved):
👍 let's test in prod

PR #732, opened 2024-10-23
Add the context editor to forms

allisonking:
## Issue(s) Resolved

Closes https://github.com/pubpub/platform/pull/732
Closes https://github.com/pubpub/platform/issues/722

TODO
* [x] Validate schema from `ContextEditor` schema
* [x] Properly pass in pubs, pub types, pubId, pubTypeId, etc.
* [x] what's up with all the re-rendering? I put a workaround in the form flow, but not yet for the create/update pub flow

## Test Plan
* Add a rich text field to a pub type and to a form
* You should be able to choose the rich text component
* Fill out a form and the rich text component. Currently, only a simple paragraph will validate since I haven't imported the actual schema yet
* Save
* The editor obj should persist to the db
* Visit the newly created pub. It'll just show `[object, object]` on the pub page since that doesn't know how to render things yet, but the Update pub form should show things properly
* Creating a pub from the create pub button _should_ work too but doesn't yet

image
image
image


core/app/components/forms/elements/ContextEditorElement.tsx
```diff
@@ -0,0 +1,77 @@
...
+	help,
+	onChange,
+	initialValue,
+}: {
+	label: string;
+	help?: string;
+	onChange: (state: any) => void;
+	initialValue?: Node;
+}) => {
+	const memoEditor = useMemo(() => {
```
allisonking, line 27: I'm memoizing liberally—I've been falling into infinite render traps and this seems to help, but I couldn't exactly tell you why

core/app/components/forms/elements/ContextEditorElement.tsx
```diff
@@ -0,0 +1,77 @@
...
+		<FormField
+			control={control}
+			name={name}
+			render={({ field }) => {
+				return (
+					<EditorFormElement
+						label={config.label ?? name}
+						help={config.help}
+						// I think this should be state.doc.toJSON() but that causes infinite rendering :(
+						onChange={(state) => field.onChange(state.doc)}
```
allisonking, line 70: I could use some help getting this to work! not sure if there's anything that can be done external to the context editor to get this to work. but this basically makes it so that someone else needs to convert to JSON on submit. I added that workaround on the form flow, but not yet for the pub create/update flow, since ideally this part would work

tefkah: I think maybe we don't want this for the Editor, I think that's a huge performance hit for the user to have to serialize the entire tree every render. Instead we probably only should serialize this on "save" (don't know what that looks like) or submit, mayyybe on "blur". 

I'm not really sure what that looks like here. Maybe what you have for the Form flow is fine.

Don't really know whether we need to update the state of the of the form anytime something is changed in the editor really. Is it possible to just leave this `onChange` out and only get the state of the field on submit, or is that not how `react-hook-form` works?

As for the infinite rerendering: I'm guessing it has something to do with the fact you would be creating a new object every time `toJSON()` is called. But still I don't know what makes that happen, maybe the `initialValue` is used in the editor as the value as well?

allisonking: ah got it! that makes sense, and I imagine we'll have to do some other processing with the editor data on save anyway (like modifying the related pubs and all that). I'll look into if `react-hook-form` a bit more, but I think it would be okay to have all the processing on the `onSubmit`

tefkah: btw i think the problem is in here: https://github.com/pubpub/platform/blob/36088913aeec23b447a8fd9d82ef9a00ce02ad16/packages/context-editor/src/ContextEditor.tsx#L136-L139

@isTravis  also left a comment about this, it seems he's doing custom there.
It does seem to have to do with creating a new object, the same thing happens when you put `{}` instead of `state.doc.toJSON()` in the `field.onChange`. I tried for a while to understand it, but I don't really get it tbh!

allisonking:
I exported `baseSchema` from ContextEditor to validate against—this works in the app, though I get an error in my test. I'm guessing this either has something to do with how I'm exporting, or something with how the context editor is being built?

```
Error: require() of ES Module /Users/allisonking/workspace/kf/v7/node_modules/.pnpm/@prosemirror-adapter+react@0.2.6_react-dom@18.3.1_react@18.3.1__react@18.3.1/node_modules/@prosemirror-adapter/react/lib/index.es.js from /Users/allisonking/workspace/kf/v7/packages/context-editor/src/ContextEditor.tsx not supported.
Instead change the require of index.es.js in /Users/allisonking/workspace/kf/v7/packages/context-editor/src/ContextEditor.tsx to a dynamic import() which is available in all CommonJS modules.
```

allisonking:
I made a context provider for the stuff the editor needs (pubs and pub types) so now the context editor actually has... context! This will likely change as we figure out the `getPub` method but at least now we can use the fancy `@` command

image

(the extra Insert [blank]s at the end are because the editor doesn't know how to get the title of existing pubs in my non-`rd` community)

allisonking:
Also, I think because the editor state is changing, currently our `useUnsavedChangesWarning` is activating all the time in the external form (probably because the form is always dirty)

tefkah:
> I exported `baseSchema` from ContextEditor to validate against—this works in the app, though I get an error in my test. I'm guessing this either has something to do with how I'm exporting, or something with how the context editor is being built? perhaps it also needs a `type: "module"`?
> 
> ```
> Error: require() of ES Module /Users/allisonking/workspace/kf/v7/node_modules/.pnpm/@prosemirror-adapter+react@0.2.6_react-dom@18.3.1_react@18.3.1__react@18.3.1/node_modules/@prosemirror-adapter/react/lib/index.es.js from /Users/allisonking/workspace/kf/v7/packages/context-editor/src/ContextEditor.tsx not supported.
> Instead change the require of index.es.js in /Users/allisonking/workspace/kf/v7/packages/context-editor/src/ContextEditor.tsx to a dynamic import() which is available in all CommonJS modules.
> ```


Sorry if you already fixed this @allisonking, but I pushed up some changes on this branch (https://github.com/pubpub/platform/tree/tfk/fix-test-require) that should fix it!

The problem is that `@prosemirror-adapter/react` is an `ESM` only package. Currently our `preconstruct` etc are setup to only output `CommonJS` however. You could solve this maybe by putting `type: module` on `context-editor`, but I think that would only work when running `pnpm p:dev && pnpm test`, not `pnpm p:build && pnpm test`, which is what we are doing in CI. Setting up `preconstruct` to output `ESM` is a bit more involved, which i did a bunch of in this PR (https://github.com/pubpub/platform/pull/510). Moving everything over to ESM is something we might want to revisit in the future.

At the moment, it's enough to just create a separate entrypoint for the schema, as that's the only thing we need when running tests at the moment, which is what I did in the abovementioned branch (as well as fix some depenency issues).

However you might still run into the same issue when trying to test the component itself. If you do, we might need to move over to `ESM`.

tefkah:
There's another issue with our `preconstruct` setup though, which prevents `pnpm p:build` from succeeding: this editor imports its own CSS, but `preconstruct` can't bundle that AFAIK, without some custom setup.
Either:
- We figure out how to make `preconstruct` bundle CSS
- We only export some CSS file with the relevant thing, and have the consumer of the component import the CSS themselves.

I think the latter is probably the correct approach, but I'm not sure how to do it without having to literally import the following like in `src/ContextEditor.tsx`

```ts
import "context-editor/style.css";
import "prosemirror-view/style/prosemirror.css";
import "prosemirror-gapcursor/style/gapcursor.css";
```

Ideally we would only need to import something like 
```ts
import "context-editor/style.css";
```

but that would imply bundling the `prosemirror` CSS into the editor CSS in some way.

Maybe that's not such a big deal, and just importing those three lines of CSS is enough. What do you think?

allisonking:
@tefkah I think this branch is in a good state in terms of editor rendering in core but the build is failing, is that something you could look into? And I'm not sure if we necessarily want to merge into main quite yet— @isTravis what do you think? or we could feature flag it?

allisonking:
Okay I think this is ready for review! 🎉 It's large since it includes all of the editor package, but the integration parts are mostly in `core`, so if reviewers could take a look there and test out using the editor in core, that'd be great!

packages/schemas/src/schemas.ts
```diff
@@ -177,4 +177,5 @@ export const FileUpload = Type.Array(
 
 export const Null = Type.Null({ description: "An empty value" });
 
-export const RichText = Type.String({ description: "A string of rich text" });
+// Rich text is validated via prosemirror, so we allow Any here
```
3mcd, line 180: Could you update this comment to mention that we validate the structure of this field when it changes?

3mcd (changes requested):
Looks good! Just a couple of documentation related things!

core/app/components/forms/elements/ContextEditorElement.tsx
```diff
@@ -0,0 +1,116 @@
...
+					const hasChanged = docHasChanged(initialDoc ?? EMPTY_DOC, state);
+					if (hasChanged) {
+						onChange(state);
+					}
+				}}
+				initialDoc={initialDoc}
+				disabled={disabled}
+			/>
+		);
+	}, [onChange]);
```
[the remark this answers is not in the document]
allisonking: oh it actually looks like it's not necessary anymore now that I'm checking for `hasChanged` before firing the `onChange` (before it would get stuck in a loop). updated!

packages/schemas/src/schemas.ts
allisonking: updated! though it only validates on submit, not on every change

3mcd (approved):
Great work and thanks for addressing those comments!

PR #746, opened 2024-10-28
feat: add new recursive pub fetch query

### LINKED ISSUE #734: Add internal utils for querying pub relationships
## Motivation

So developers can ergonomically query for pub relationships

## Requirements

- Ability to query for all fields of all relationships on a pub
- Ability to, optionally, query recursively for relationships (e.g. show me related pubs AND their relationships)

## Acceptance Criteria

- Should be able to optionally get relationships or child fields
- Should be able to limit levels of recursion
- Discuss specific implementation with team, currently leaning against master `getPubs` function in favor of specific functions (e.g. `getPubsWithRelations`) that share code

## Extra work

- Thoughts about performance for pubs (and trees of pubs) with lots of relationships, e.g., a "collection" type pub that might contain 1000s of Pubs.
- Ability to, optionally, specify which fields of relationships to return (e.g. show me all relationship but only return their ids and the `title` field)
- Developers can, optionally, filter the query to get relationships defined by specific fields (e.g. show me this pub's "image" relationships)
- Developers can, optionally, add pagination and limits to relation sub-queries

tefkah:
## Issue(s) Resolved

Resolves #734 
I want to do another pass which includes using this function in various parts of the app/api, ideally in a bunch of separate PRs.

Adds a new "god" function for fetching pubs, which is able to fetch related pubs and children, fetch all the same information as the old `getPub(s)` functions, plus a little more.

I do not intend us to use this function most of the time in components, but instead have some more specific functions like `getPubsForCommunity`, or `getPub` or something which use this function with some default settings. 

You can still fetch pubs either by their `id`, `pubTypeId`, `stageId`, or `communityId`, and limit, offset, order by the results. There are a number of intersting differences however, which I will go through below.

#### SQL

This query should be somewhat more efficient than the old `getPubs` function, specifically the `withChildren` part.

As we can see here, the old function fetches a bunch of columns, the `values`, and the `pubType` in the recursive part of the query from the `pubs` table

https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L171-L205

However, in `getPubBase`, instead of retrieving this information from the `children` CTE we just created, we do another look through the `pubs` table, only getting the `children` from the `children` CTE

https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L236-L274

This is kind of inefficient, as this requires an extra excursion through the `pubs` table.

Instead in the new query, we already retrieve all values and most other data in the recursive part of the query, and then `select` into the CTE we just fetched instead

https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L982-L992

While solely gut-feeling based, I feel like this is probably more efficient. We fetch (almost) all the columns we need in the recursive part, and then just kind of clean that up at the end, rather than doing two big fetching steps.

#### Input

You can no longer specify specifc columns of the `pubs` table to fetch. I did not think this functionality was particularly useful, as most of those colmuns you either almost always want, or almost never want.

#### Output

Support for the `assignee` is dropped, as we want to transition away from this. Let me know if this is still required.

The output is mostly the same. The main difference is that I opted to fetch the `pubValues` as an array of values, rather than as an object of slugs with values. This is because we now fetch `relatedPub`s, which would force some ugly modifications to the current object in a number of ways:

1. We would need a `{ value: 'relationshipValue', relatedPub: { ... } }` value for slugs of fields which are relationships, as we cannot simply either attach the `value` of the relationship field or just the relatedPub, as both constitute the relationship.
2. Relationship values would need to be arrays, as a single relationship field can be attached to multiple pubs.

So the `values` object would have to be typed in any of the following ways

```ts
// keeping the normal values the same
type Values = {
	[slug in string]: unknown | { value: unknown, relatedPub: Pubs }[]
}

// also making the normal values an object, only relationship values an array
type Values = {
	[slug in string]: { value: unknown } | { value: unknown, relatedPub: Pubs }[]
}

// make everything an array, even though normal pub values are always singular
type Values = {
	[slug in string]: { value: unknown, relatedPub?: Pubs }[]
}
```

I thought neither of these options very good, and since we almost always convert the `values` object to an array anyway, and we already talked about making the values an array previously, I decided to go ahead with it.

Currently the values have some information about the field they belong to on the `field` property, like so (ignore the `MaybeWithRelatedPub` thingy)

https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L749-L766

Maybe this does not need to be a nested object, e.g. `schemaName` and `slug` could just be properties on this `value` object like so

```ts
	values: ({
		id: PubValuesId;
		fieldId: PubFieldsId;
		value: unknown;
		relatedPubId: PubsId | null;
		createdAt: Date;
		updatedAt: Date;
		/**
		 * Information about the field that the value belongs to.
		 */
		schemaName: CoreSchemaType;
		slug: string;
	} & MaybeWithRelatedPub<Options>)[];
 ```
Let me know what you prefer!

#### Types

I made it so that the output type of the new function depends on the input parameters in a number of ways.

1. If `pubId` is used the "where" clause, return a single Pub. Otherwise return an array. This is accomplished through function overloads here.

https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L789-L811

Theoretically this could also be accomplished in the generic return type, as the others are. I implemented the function overloads before the generic return type, hence why it's like this. If you'd rather me make all type info be defined through the generic return type, let me know. That is a bit cleaner.

2. The `withX` options. I think a table would be most useful. Here the type in the `true` and `false` column represent what the output type matches with when the option is true or false. The `default` column the default setting for the option.

| Option | `true` | `false` | default |
|--------|--------|---------|---------|
| `withChildren` | `{ children: ProcessedPub<Options>[] }` | `{ children?: undefined }` | `true` |
| `withRelatedPubs` | `{ values: { value: unknown, relatedPub: ProcessedPub<Options> }[] }` | `{ values: { value: unknown }[] }` | `true` |
| `withPubType` | `{ pubType: ProcessedPubType }` | `{ pubType?: undefined }` | `false` |
| `withStage` | `{ stage: ProcessedPubStage }` | `{ stage?: undefined }` | `false` |

I think this is nice, as it makes it harder to make mistakes when fetching stuff. However you could argue it makes things more complicated and that this specific type information should be left to more specific functions which use this "god" function. Let me know what you think!

##### Max depth

This function is able to fetch a pub and all its children/related pubs up to a given depth. By default it is set to 2, so it will fetch the pub and all its children and related pubs, but not their children or related pubs.

This is done by incrementing a `depth` column in the CTE here

https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L962-L963

 and then limiting the results with a `where` clause here

 https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L973

##### Cycle detection

In addition to the depth setting, this query also has built in cycle detection, and will not get stuck in an infinite loop even if there are cycles in the pub tree, even if `depth` is set to `Infinity` (which we probably should not allow in the first place).

This is done by keeping track of the path we've taken so far in the `path` column and setting a flag if we see a pub that is already in the path.

https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L964-L971

We then filter those out here:

https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L973-L974

The query also allows you to specify whether the cycle detection should be inclusive or exclusive, meaning whether the first pub that is detected to be circular should be included or excluded.

This allows you to e.g. show the user that the child of this pubs child is the same as this pub more easily, as you can just look at the `path` or `isCycle` flag on the pub. This is the default behavior. If it's set to `exclude`, you will not see the first pub that is part of a cycle. 

##### Field slugs

This query allows you to specify a list of field slugs to fetch, and will only fetch the values of the fields that match the slugs.

This is done here:

https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L944-L946

Note that this currently does not make any distinction between relationship fields and other fields. So if you set `fieldSlugs` to `['title', 'description']`, you will get all the values of the `title` and `description` fields, but will not get the values or related pubs of the relationship fields.

An improvement here would be to allow the user to specify which fields to exclude, rather than only which to include.

Note that this also applies to the children and related pubs, so if you specify `fieldSlugs` to fetch, you will only get those fields for the pub itself and its children/related pubs. Maybe this could be improved, but I don't really know if we intend to do ever do this, or what a good API would be for specifying this.
E.g. how would you specify "give me all fields for the top level pubs, but only the `title` and `description` fields for the children/related pubs"?

##### Excluding children and related pubs

This query allows you to specify whether you want to exclude children or related pubs. This is done by only running the recursive part of the `CTE` if either `withChildren` or `withRelatedPubs` is true (which are the default settings).

And by not joining the `pubs` table on `parentId` or `relatedPubId` if `withChildren` or `withRelatedPubs` is false respectively.

https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L906-L933

##### Searching

This has been scrapped for now, I'll add this back in when #750  lands.

~I have also included (incredibly primitive) searching functionality on the Pubs.
This is only done on the top level (so not on the children or related pubs).~

https://github.com/pubpub/platform/blob/d7d5ff8eb0b4d7bd36ac226b238efe0396777019/core/lib/server/pub.ts#L902-L904

~This forcebly converts the `value` of the fields to text and does a case-insensitive search.~

~In the future I would want to limit this to only `isTitle` fields of the type @allisonking is implementing in #750, or at least make it so that the user can specify which fields to search on.~

~Likely there's a much better way to do this, e.g. using Postgres' [full text search](https://www.postgresql.org/docs/current/textsearch-intro.html). I have zero experience with this however.~

### Future work

- [ ] We need to add the ability to filter out pubs which the user does not have permission to view. We could do this manually in JS for now, but it would be nice to do it in the SQL, although that may proof extremely inefficient or convoluted, deferring to your judgement @kalilsn.
- [x] ~The `limit` functionality does not work properly at all at the moment. It will limit both the number of root-level pubs and related pubs/children, so you really only get `LIMIT` total pubs in some random order. I will tackle this in a follow up PR, as I would like to get this in now.~ This has been added!

## Test Plan

1. Run the tests! I wrote quite a few extensive tests for this function.
	- Some are kind of large, which sometimes makes it a bit hard to identify why they failed. I could add some more granular tests, let me know if you'd like me to do that.

Vids
### General structure

https://github.com/user-attachments/assets/b52fcf98-8565-4993-85dc-22db7f108618

### Deep dive into the recursive part of the query

https://github.com/user-attachments/assets/854e1d85-fbc7-4d28-aa51-e42dfd173414

https://github.com/user-attachments/assets/a6e5c15e-8170-4825-8864-3dabfe754eca

### CODE REVIEW

File: core/lib/server/pub.ts [hunk 1]
```diff
@@ -589,3 +600,564 @@ export const deletePub = async (pubId: PubsId) =>
...
+									sql<PubsId[]>`array_append(pub_tree."path", pubs.id)`.as(
+										"path"
+									),
+								])
+								.where("pub_tree.depth", "<", depth)
+								.where("pub_tree.isCycle", "=", false)
+								.$if(cycle === "exclude", (qb) =>
+									// this makes sure we don't include the first pub that is part of a cycle
+									qb.where(sql<boolean>`pubs.id = any(pub_tree.path)`, "=", false)
+								)
```
tefkah, line 978: this part of the query handles detecting cycles and limiting the recursed depth

core/lib/server/pub.ts
```diff
@@ -589,3 +600,574 @@ export const deletePub = async (pubId: PubsId) =>
...
```
tefkah, line 965: this part of the query handles detecting cycles and limiting the recursed depth

core/lib/server/pub.ts
```diff
@@ -589,3 +600,574 @@ export const deletePub = async (pubId: PubsId) =>
...
+					)
+					.$if(Boolean(props.communityId), (qb) =>
+						qb.where("pubs.communityId", "=", props.communityId!)
+					)
+					.$if(Boolean(props.pubTypeId), (qb) =>
+						qb.where("pubs.pubTypeId", "=", props.pubTypeId!)
+					)
+					.$if(Boolean(limit), (qb) => qb.limit(limit!))
+					.$if(Boolean(offset), (qb) => qb.offset(offset!))
+			)
```
tefkah, line 991: the reason we split up finding the root pubs here is because we want to be able to properly limit them. If we did it in the `withRecursive` part (which doesn't really work due to https://github.com/kysely-org/kysely/issues/646), where we `leftJoin` with `pub_values`, this wouldn't work, as `pubs.leftJoin('pub_values')` yields `#pubs * #pubValuesPerPub` rows. 

Ex.

If we have two pubs, one with `croccroc:title` and `croccroc:description`, and one with only `croccroc:title`, we would get the following rows (not actual output, but similar)

```ts
[{ 
	id: '1',
	slug: 'croccroc:title',
	value: 'Title of Pub 1'
},
{ 
	id: '1',
	slug: 'croccroc:description',
	value: 'A description of pub 1'
},
{ 
	id: '2',
	slug: 'croccroc:title',
	value: 'Title of Pub 2'
}]

```

If we were to put `limit(2)` near there, we would only get the rows

```ts
[{ 
	id: '1',
	slug: 'croccroc:title',
	value: 'Title of Pub 1'
},
{ 
	id: '1',
	slug: 'croccroc:description',
	value: 'A description of pub 1'
}]

```

which are only the values of a single pub. 

So we need to limit before we `leftJoin`, thus this setup.

core/lib/server/pub.ts
```diff
@@ -589,3 +600,574 @@ export const deletePub = async (pubId: PubsId) =>
...
+			//
+			// we could instead only look for the related and child pubs and ignore the other values
+			// but this would mean we would later need to look up the values for each pub as a subquery
+			// this way, the only subqueries we need below are looking up information already stored in pub_tree,
+			// which is more efficient
+			.withRecursive("pub_tree", (cte) =>
+				cte
+					// we need to do this weird cast, because kysely does not support typing the selecting from a later CTE
+					// which is possible only in a with recursive query
+					.selectFrom("root_pubs_limited as p" as unknown as "pubs as p")
```
tefkah, line 856: looks weird, but it's possible to reference "later" CTEs in recursive CTEs in postgres, see https://www.postgresql.org/docs/current/sql-select.html#SQL-WITH

tefkah: `kysely` obviously doesn't type that correctly though, hence the cast

core/lib/server/pub.ts
```diff
@@ -589,3 +600,574 @@ export const deletePub = async (pubId: PubsId) =>
...
+	 * The `updatedAt` of the latest value, or of the pub if the pub itself has a higher `updatedAt` or if there are no values
+	 *
+	 * We do this because the Pub itself is rarely if ever changed over time.
+	 * TODO: Possibly add the `updatedAt` of `PubsInStages` here as well?
+	 * At time of writing (2024/11/04) I don't think that table has an `updatedAt`.
+	 */
+	updatedAt: Date;
+} & MaybePubChildren<Options> &
+	MaybePubStage<Options> &
+	MaybePubPubType<Options>;
```
tefkah, line 778: maybe a bit ugly, but this is the only way i found to do roughly this pseudo type code

```ts
if( options.withPubType === true ) {
	return { pubType: PubType }
} else {
 	return { pubType?: never }
}
```

This is bc it's not possible to do "conditional optionality" in typescript, ie you cant say "only make this type optional for a certain condition". You need to do some messed up stuff for that.

tefkah: I think optional `never` is preferable over just leaving it out, because it allows you to do 

```ts
if(pub.pubType){
}
```
rather than having to do

```ts
if('pubType' in pub && pub.pubType){
}
```

core/lib/server/pub.ts
```diff
@@ -589,3 +600,574 @@ export const deletePub = async (pubId: PubsId) =>
...
+
+const DEFAULT_OPTIONS = {
+	depth: 2,
+	withChildren: true,
+	withRelatedPubs: true,
+	withPubType: false,
+	withStage: false,
+	cycle: "include",
+} as const satisfies GetPubsWithRelatedValuesAndChildrenOptions;
```
tefkah, line 787: please let me know if you have other preferences for these defaults!

core/lib/server/pub.ts
```diff
@@ -589,3 +600,574 @@ export const deletePub = async (pubId: PubsId) =>
...
+			.withRecursive("pub_tree", (cte) =>
+				cte
+					// we need to do this weird cast, because kysely does not support typing the selecting from a later CTE
+					// which is possible only in a with recursive query
+					.selectFrom("root_pubs_limited as p" as unknown as "pubs as p")
+					.leftJoin("pub_values as pv", (join) =>
+						join.on((eb) =>
+							eb.and([
+								eb("p.id", "=", eb.ref("pv.pubId")),
+								// eb("pv.relatedPubId", "is not", null),
```
allisonking, line 861: I guess this commented line can be removed since we are joining all the fields? I agree it sounds more efficient!

core/lib/server/pub.ts
```diff
@@ -589,3 +600,574 @@ export const deletePub = async (pubId: PubsId) =>
...
+							])
+						)
+					)
+					.innerJoin("pub_fields", "pub_fields.id", "pv.fieldId")
+					.$if(Boolean(fieldSlugs), (qb) =>
+						qb.where("pub_fields.slug", "in", fieldSlugs!)
```
allisonking, line 867: is there somewhere we do this filtering by field slugs, or is that for upcoming functionality?

core/lib/server/pub.ts
```diff
@@ -589,3 +600,574 @@ export const deletePub = async (pubId: PubsId) =>
...
+													]
+												: []),
+										])
+									)
+								)
+								.leftJoin("pub_values", (join) =>
+									join.on((eb) =>
+										eb.and([
+											eb("pubs.id", "=", eb.ref("pub_values.pubId")),
+											// eb("pub_values.relatedPubId", "is not", null),
```
allisonking, line 924: same thing, this can be cleaned up?

core/lib/server/pub.ts
```diff
@@ -589,3 +600,574 @@ export const deletePub = async (pubId: PubsId) =>
...
+							"inner.relatedPubId",
+							"inner.valueCreatedAt as createdAt",
+							"inner.valueUpdatedAt as updatedAt",
+							jsonObjectFrom(
+								eb
+									.selectFrom("pub_tree as inner2")
+									.select(["inner2.schemaName", "inner2.slug"])
+									.whereRef("inner2.valueId", "=", "inner.valueId")
+									.limit(1)
+							).as("field"),
```
allisonking, line 1024: I vote for the flat version, not nested in `field`. would also be good for the upcoming `isTitle` field, I think it just reads a little more naturally being attributes of the value

allisonking (approved):
amazing! great tests, explanation, and videos 💯

core/lib/server/pub.ts
tefkah: that's indeed upcoming functionality, mostly useful for the api i think!

core/instrumentation.ts
```diff
@@ -15,12 +15,12 @@ export async function register() {
...
-				"NEXT_RUNTIME is `nodejs` and NODE_ENV is `development`; skipping OTEL + Sentry registration."
-			);
-			return;
-		}
+		// if (process.env.NODE_ENV === "development") {
+		// 	logger.info(
+		// 		"NEXT_RUNTIME is `nodejs` and NODE_ENV is `development`; skipping OTEL + Sentry registration."
+		// 	);
+		// 	return;
+		// }
```
kalilsn, line 23: is this supposed to be commented out?

core/lib/server/pub.db.test.ts
```diff
@@ -181,3 +202,486 @@ describe("createPubRecursive", () => {
...
+		const pubs = await getPubsWithRelatedValuesAndChildren(
+			{ communityId: newCommunity.community.id },
+			{
+				depth: 10,
+				limit: 1,
+				_debugDontNest: true,
+			}
+		);
+
+		// 1 root pub, 1 child pub, 1 related pub, even tohugh the limit is . This is correct behavior.
```
kalilsn, line 497:
```suggestion
		// 1 root pub, 1 child pub, 1 related pub, even though the limit is 1. This is correct behavior.
```

core/lib/server/pub.db.test.ts
```diff
@@ -181,3 +202,486 @@ describe("createPubRecursive", () => {
...
+		expect(withCycleIncluded.length).toBe(3);
+
+		expect(withCycleIncluded.find((p) => p.isCycle)).toBeDefined();
+
+		expect(withCycleExcluded.length).toBe(2);
+
+		expect(withCycleExcluded.find((p) => p.isCycle)).toBeUndefined();
+	});
+
+	it("it be able to only fetch fields for certain slugs", async () => {
```
kalilsn, line 559:
```suggestion
	it("should be able to only fetch fields for certain slugs", async () => {
```

core/lib/server/pub.ts
```diff
@@ -589,3 +600,551 @@ export const deletePub = async (pubId: PubsId) =>
...
+		schemaName: CoreSchemaType;
+		fieldSlug: string;
+	} & MaybeWithRelatedPub<Options>)[];
+	createdAt: Date;
+	/**
+	 * The `updatedAt` of the latest value, or of the pub if the pub itself has a higher `updatedAt` or if there are no values
+	 *
+	 * We do this because the Pub itself is rarely if ever changed over time.
+	 * TODO: Possibly add the `updatedAt` of `PubsInStages` here as well?
+	 * At time of writing (2024/11/04) I don't think that table has an `updatedAt`.
```
kalilsn, line 769: It seems like we have two options when dealing with a pub's updatedAt: either look at related tables (values) to calculate the user-facing updatedAt during read queries, or update it during write queries that only modify values. I think this is a more general problem with the application, which also affects forms right now, and probably will affect just about anything where we want to show a last modified date!

IMO we should agree on one of these two strategies for simplicity. I'm slightly in favor of doing it on writes rather than reads, because I think those queries will wind up simpler in the long run. So instead of using the most recent pub value's updatedAt here (or even the last time the pub was moved to a new stage) we would instead explicitly update a pub's updatedAt each time we moved it to a new stage or updated its values. But tbh I don't feel that strongly or really want you to change this now, I'd just like to make sure we agree on a strategy!

kalilsn (approved):
Really nice work!

I am thinking about how to make it easy to query whether the user has read/write capabilities for the pub entirely inside the db and I think we can do it. But I think the best way to achieve that is to store the mapping between role/{pub,stage,community,form}_membership and a set of capabilities in a database table, rather than in the codebase somewhere. I'll keep it in mind as I work on the capabilities queries!

core/instrumentation.ts
tefkah: no

tefkah, line 23:
```suggestion
		if (process.env.NODE_ENV === "development") {
			logger.info(
				"NEXT_RUNTIME is `nodejs` and NODE_ENV is `development`; skipping OTEL + Sentry registration."
			);
			return;
		}
```

core/lib/server/pub.ts
tefkah: hmm yeah i agree, that makes more sense. There'll probably some queries where we fetch pubs where we don't want to fetch the values, and needing that to calculate the updatedAt is a lot of extra overhead, plus lowers the cache hit rates. I won't change it now, but we will have to do that in the future.

tefkah:
> Really nice work!
> 
> I am thinking about how to make it easy to query whether the user has read/write capabilities for the pub entirely inside the db and I think we can do it. But I think the best way to achieve that is to store the mapping between role/{pub,stage,community,form}_membership and a set of capabilities in a database table, rather than in the codebase somewhere. I'll keep it in mind as I work on the capabilities queries!

Haven't looked at it at all, but maybe this is something? https://github.com/ben-pr-p/kysely-access-control/tree/main

Issue #735, opened 2024-10-24
Add ability to designate a field as the title of a pubtype, both in the database and in the type editor UI


gabestein:
## Motivation

- So admins can specify which field should serve as the entity name
- So developers can know which field to use as the name in locations throughout the app (e.g. pub page, API, etc.)

## Requirements

For now, we're just gonna rough this in, knowing that a full type editor redesign is planned for early next year.

- New "column" to the right of "fields" called "Name", with helper text that says: "the selected field will be used as the pub's name"
- Radio button that allows user to choose one field as the `name` field
- When creating a type, new checkbox when adding a field that says "use this field as the name field". If a name has already been selected, clicking the checkbox will overwrite the selection

## Acceptance Criteria

- Add a column to (probably?) the PubFieldToPubType table, called something like `isTitle`. If possible, it should only be possible to add one pub with `isTitle` for each PubType.
- Add a uniqueness index that guarantees that only one PubField can be marked as a title for a given PubType.
- Users can only select fields that use string, rich text, or number schemas
- A designated name field is required of all pub types
- One-time data migration that adds the designation to the first existing "title" or "name" fields; if none exists, it adds it to the first-created string, rich text, or number field; if none exists...I don't think there's one in the db, but could just create a "title" field in that case

tefkah:
This should also add a column to (probably?) the PubFieldToPubType table, called something like "isTitle".
If possible, it should only be possible to add one pub with `isTitle` for each `PubType`. 

This might be a separate ticket.

PR #750, opened 2024-10-30
Add `isTitle` to pubtype-pubfield relationships

allisonking:
## Issue(s) Resolved

Closes https://github.com/pubpub/platform/issues/735

* Adds a new `isTitle` column to the join table between PubType and PubFields.
* Users must specify a PubField as the "name" for a pubtype when creating a pub type

## Test Plan
* Migrate your local database. This should run the migration, which designates a field as the "title" for each pub type. Make sure all of your pub types now have a pub field designated as 'isTitle'!
* Create a new pub type. You should be able to specify a field as the title
* Edit a pub type. You should be able to change which field is the title

Pub types
image

Editing pub type
image
* Only two options are available to be chosen as a title because those are the only string/number/richtext fields
* The delete button is disabled if the option is the current title

Creating a pub type
image

If no field is chosen as the name field
image

## Notes

Nothing uses these name fields yet. `PubTitle` would be a good candidate, though I had some trouble with the prisma query there, so figured I'd put this PR up first since it's already fairly chunky


core/prisma/seed/seedCommunity.ts
```diff
@@ -497,7 +499,7 @@ export async function seedCommunity<
 		 * 	},
 		 * 	pubTypes: {
 		 * 		Article: {
-		 * 			Title: "A Great Article",
+		 * 			Title: { isTitle: true },
```
tefkah, line 502: good catch, that comment is indeed wrong.

core/app/c/[communitySlug]/types/actions.ts
```diff
@@ -18,6 +18,22 @@ export const addPubField = defineServerAction(async function addPubField(
...
+	await autoRevalidate(
+		db.updateTable("_PubFieldToPubType").set({ isTitle: false }).where("B", "=", pubTypeId)
+	).execute();
+	await autoRevalidate(
+		db
+			.updateTable("_PubFieldToPubType")
+			.set({ isTitle: true })
+			.where("A", "=", pubFieldId)
+			.where("B", "=", pubTypeId)
+	).execute();
```
Comment (allisonking) [inline core/app/c/[communitySlug]/types/actions.ts:34]: I couldn't figure out how to do this in one query :( I tried using a `with` but that didn't work, which I remember not working for stages and pubs a bit ago too...

core/lib/server/_legacy-integration-queries.ts
```diff
@@ -83,7 +87,8 @@ export const _getPubType = async (pubTypeId: string): Promise<GetPubTypeResponse
 	if (!pubType) {
 		throw new NotFoundError("Pub Type not found");
 	}
-	return pubType;
+	const { fields, ...rest } = pubType;
+	return { ...rest, fields: fields.map((f) => f.pubField) };
```
allisonking, line 91: I'm not great with prisma and wasn't sure how to get the query to not nest `pubField` there—it seems like now that I've explicitly defined the join table, prisma doesn't want to allow the direct `fields` selection anymore. so this part just rejiggers the result from 

```js
fields: [
  {
    pubField: {
      id: '',
      name: '',
      slug: '',
      schema: {...}
    }
  }
]
```

to

```js
fields: [
  {
    id: '',
    name: '',
    slug: '',
    schema: {...}
  }
]
```
which is what the return type expects

core/lib/server/pubtype.ts
```diff
@@ -92,12 +92,22 @@ export const getAllPubTypesForCommunity = (communitySlug: string) => {
 						.selectFrom("_PubFieldToPubType")
 						.whereRef("B", "=", "pub_types.id")
 						.select((eb) =>
-							eb.fn.coalesce(eb.fn.agg("array_agg", ["A"]), sql`'{}'`).as("fields")
+							eb.fn
```
allisonking, line 95: instead of returning a list of field Ids `PubFieldsId[]`, we now return a list of `{id: PubFieldsId, isTitle: boolean}`

core/prisma/exampleCommunitySeeds/unjournal.ts
```diff
@@ -288,101 +288,133 @@ export default async function main(prisma: PrismaClient, communityUUID: string)
...
+				slug: "unjournal:title",
+				communityId: communityUUID,
+				schemaName: CoreSchemaType.String,
+			},
 			{
 				id: fieldIds[1],
 				name: "Description",
 				slug: "unjournal:description",
 				communityId: communityUUID,
+				schemaName: CoreSchemaType.String,
```
allisonking, line 303: not really related to this PR, but our prod data now has schemaNames for unjournal, so just updating our seed data to match. not sure I chose the right schema for each type, but did my best!

core/prisma/exampleCommunitySeeds/unjournal.ts
```diff
@@ -393,64 +425,108 @@ export default async function main(prisma: PrismaClient, communityUUID: string)
 			id: submissionTypeId,
 			name: "Submission",
 			communityId: communityUUID,
-			fields: {
```
allisonking, line 396: I couldn't figure out how to do this connect in prisma now that we've defined our own join table, so went with kysely 🥲

core/prisma/migrations/20241030175617_update_comments/migration.sql
```diff
@@ -0,0 +1,178 @@
+-- generator-version: 1.0.0
```
allisonking, line 1: I'm not really sure why this generated

core/prisma/schema.prisma
```diff
@@ -204,13 +204,27 @@ model PubType {
...
 
-  fields PubField[]
+  fields PubFieldToPubType[]
   pubs   Pub[]
   Form   Form[]
 
   @@map(name: "pub_types")
 }
 
+model PubFieldToPubType {
```
allisonking, line 214: this defines the join table explicitly. previously, from what I understand, prisma had made a join table behind the scenes (hence names like "A" and "B") but since we now want to add our own column, I defined it here, but tried to keep it the same as the existing table. 

this seems to work fine for kysely, but causes some trouble with some of the prisma queries

core/prisma/migrations/20241030175616_add_is_title_to_pubfield_pubtype_join_table/migration.sql
```diff
@@ -0,0 +1,97 @@
...
+    LIMIT 1
+  ) AS "newField" ON true
+WHERE
+  NOT EXISTS (
+    SELECT 1
+    FROM "_PubFieldToPubType"
+    WHERE "_PubFieldToPubType"."B" = pub_types.id
+    AND "_PubFieldToPubType"."isTitle" = true
+  )
+  AND "newField".id IS NOT NULL;
```
tefkah, line 95: what does this part intend to do?

from what i gather, it tries to add a new field to a pubtype that looks like `%title%` or `%name%` to pubtypes that do not have a `title` field yet, and sets it to have `isTitle: true` right? I see that that is described in the issue, but is that what we want to happen?

I also see that we are requiring a title field to be selected when creating a pubtype, so that's inline with this.

I am generally in favor of this approach: I think all pubtypes should have titles. But I'm not sure everyone agrees. I will put it in the slack as a discussion point!

core/prisma/exampleCommunitySeeds/unjournal.ts
tefkah: you'll never see me complain about replacing prisma with kysely haha

core/app/c/[communitySlug]/types/TypeEditor.tsx
```diff
@@ -24,12 +36,23 @@ type Props = {
...
+		.array(
+			z.object({
+				fieldId: z.string(),
+				name: z.string(),
+				slug: z.string(),
+				schemaName: z.nativeEnum(CoreSchemaType).nullable(),
+			})
+		)
+		.min(1, { message: "Add at least one field" }),
+	titleField: z.string().min(1, { message: "Designate a title field" }),
```
Comment (tefkah) [inline core/app/c/[communitySlug]/types/TypeEditor.tsx:49]: I asked about this in Slack. I think this is a good constraint, but I'm not sure everyone agrees!

tefkah (approved):
I think it looks and works great! Very impressed at how quickly you managed to put this together!

I would hold off on merging it until we get some feedback from the rest about always requiring a title field. 

Otherwise I have tested creating and editing pubtypes with title fields, all works great. I also checked whether updating/creating pubs still worked, and it did.

For good measure I also ran the unjournal evaluation flow once more, and that still worked.

There was one small issue i found, which you don't need to fix, which is that if you select another field as the title while updating a PubType, it like moves around 

https://github.com/user-attachments/assets/f41b06c1-1f0a-4271-bcea-8031ae68bfcf

Again, does not need to be fixed as this UI is very temporary, but it did bring me to a thought i'd like your take on. The above issue could be fixed by sorting the pubfields in a certain order. Ideally it'd be something like "when pubfield was added to pubtype", but we can't really track that as the `_PubFieldsToPubType` table does not have a createdAt column (yet). 

We may want to add that, but maybe in the nearish future (as a follow up PR) i think the best solution might be to always put the `Title` field first. This also makes sense when editing/creating pubs, as you would probably want the title field on top.

Just a thought, again does not need to be done rn!

core/prisma/migrations/20241030175616_add_is_title_to_pubfield_pubtype_join_table/migration.sql
allisonking: oops I meant to add a comment—this was meant to get at the last part of:

> One-time data migration that adds the designation to the first existing "title" or "name" fields; if none exists, it adds it to the first-created string, rich text, or number field; if none exists...I don't think there's one in the db, but could just create a "title" field in that case

I was hoping this wouldn't come up, but at least locally it comes up with the Arcadia seed—for instance, the pub type "PDF Download" only has a field "PDF" which is of type `FileUpload`, so this bit of the migration will detect that "PDF Download" doesn't have an `isTitle` field, and add `arcadia:title` to this pub type. It seems a little heavy handed, but we would need something like it if we do in fact want to require that all pub types have a title field

core/app/c/[communitySlug]/types/TypeEditor.tsx
Comment (allisonking) [reply core/app/c/[communitySlug]/types/TypeEditor.tsx:49]: as discussed in slack, I'll remove this constraint! also removing the "disabled" prop on `RemoveFieldButton` when the field was the title

core/app/c/[communitySlug]/types/utils.ts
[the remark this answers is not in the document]
Comment (allisonking) [reply core/app/c/[communitySlug]/types/utils.ts:1]: makes sense to me, and easy to change!

core/prisma/migrations/20241030175616_add_is_title_to_pubfield_pubtype_join_table/migration.sql
allisonking: removing these last two steps! migration becomes simpler :)

allisonking:
> Again, does not need to be fixed as this UI is very temporary, but it did bring me to a thought i'd like your take on. The above issue could be fixed by sorting the pubfields in a certain order. Ideally it'd be something like "when pubfield was added to pubtype", but we can't really track that as the _PubFieldsToPubType table does not have a createdAt column (yet).
> 
> We may want to add that, but maybe in the nearish future (as a follow up PR) i think the best solution might be to always put the Title field first. This also makes sense when editing/creating pubs, as you would probably want the title field on top.

I think that's a good idea! I just changed it to sort the title first, and then alphabetically from there. This is just for the edit field, as I think the create field makes more sense as-is (where they are appended to a list as you add them). You still get some jumping around, but it's a little more expected I think


https://github.com/user-attachments/assets/7a3bd178-d126-4fcc-97e2-fc7e641d23cb

PR #752, opened 2024-11-05
feat: shadcn sidebar

tefkah:

Move to new `@shadcn` sidebar component.

#### New sidebar

- Easier to maintain
- Better mobile experience

#### Move `base` color from `slate` to `gray`, so from a bluish gray to a more neutral one

this is bc our header and eg form config sidebar already have bg `gray-50`, so it makes sense to make the sidebar also `gray-50`. However, our default accent color `slate-100` does not look great on the `gray-50` bg, therefore @ktyjscott and I decided to move our main gray color to `gray` over `slate`. 
These are most of the changes in `ui`, and can be ignored.

#### Authorization checks

I changed the way authorization checks are done on the main page. Each link now has its own check, many of which are the same and cached.

Ideally, all these checks should be the same as the checks on the page itself, ideally-ideally we would define these checks in a separate place and use the same one for the page and the link to the page. ~I have not yet implemented this~. I have started working on this here, would love your thoughts https://github.com/pubpub/platform/pull/908

## Test Plan

1. Click on things, see if you can still get everywhere.
2. Log in as an editor (some@pubpub.org), check to see that you can't see that many things. (contributor will still fail)

Demo

https://github.com/user-attachments/assets/635f6d96-9a0e-4752-9a61-99ef4e4fb988


ktyjscott (changes requested):
Posting a screenshot of the suggested changes along with the figma link:

Screenshot 2025-01-14 at 12 11 08 PM

https://www.figma.com/design/jJIpEL3bBAHPFCHYoEPRzY/Platform-Components?node-id=828-20541&t=xWMBWQSCOyojdTBQ-1


1. Update community dropdown icon to utilize `chevrons-up-down`
2. Break up links into 3 distinct sections: 'Views', 'Manage', and 'Admin'. These should be styled at 12px, bold, all caps, with a color of slate-500
3. All Pubs now gets organized under 'Views', all active state colors should also be set to gray-200
4. 'All Workflows' gets added - should take the user to the stages page (https://app.pubpub.org/c/unjournal-demo/stages)
5. Under the 'Manage' section, links should be organized as shown in the designs, with the 'Workflows' link taking the user to the stage editor experience. Also make sure this link's icon is utilizing the `layers` icon
6. Settings receives a new icon: `settings-2`
7. Add new dropdown section for Docs
8. Utilize the new base 'User' segment as shown in the shadcn docs. I think this design with the menu helps clean up the options a lot, and i like how the icons are consistent with the community switcher up top

Please lmk if you have any questions or feedback on this that we may have overlooked :)

EDIT: I did in fact overlook the mobile version. I think the layout you have right now looks good - I would just say let's switch the icon to utilize `menu` as I think that will be a lot more recognizable/understandable to our users.

core/app/(user)/forgot/ForgotForm.tsx
```diff
@@ -21,6 +21,10 @@ const forgotPasswordSchema = z.object({
 export default function ForgotForm() {
 	const form = useForm<z.infer<typeof forgotPasswordSchema>>({
 		resolver: zodResolver(forgotPasswordSchema),
+		defaultValues: {
+			// in order to prevent "Form changed from uncontrolled to controlled" React errors
+			email: "",
+		},
```
tefkah, line 27: i added a number of these things in order to get rid of some errors, which were making tests fail in dev mode bc the error popup would overlap with the loginswitcher 😅

core/app/(user)/login/LoginForm.tsx
```diff
@@ -55,7 +60,7 @@ export default function LoginForm() {
 									<FormItem>
 										<FormLabel>Email</FormLabel>
 										<FormControl>
-											<Input placeholder="m@example.com" {...field} />
+											<Input placeholder="name@example.com" {...field} />
```
tefkah, line 63: about time we changed that, could also be `mail` probably

core/app/(user)/reset/ResetForm.tsx
```diff
@@ -84,11 +88,9 @@ export default function ResetForm() {
...
 						<DialogTitle>Success</DialogTitle>
-						<DialogDescription>
-							<p className="flex flex-col gap-2">
-								<span className="text-green-700">Success - password reset!</span>
-								Redirecting in 5 seconds...
-							</p>
+						<DialogDescription className="flex flex-col gap-2">
+							<span className="text-green-700">Success - password reset!</span>
+							Redirecting in 5 seconds...
 						</DialogDescription>
```
tefkah, line 94: `<DialogDescription>` already renders a `<p>`, so this was giving hydration errors bc you should not render `<p><p/></p>`

core/app/c/[communitySlug]/NavLinkSubMenu.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+	useEffect(() => {
+		let mounted = true;
+		if (mounted) {
+			setActuallyOpen(open ?? false);
+		}
+		return () => {
+			mounted = false;
+		};
+	}, [open]);
```
Comment (3mcd) [inline core/app/c/[communitySlug]/NavLinkSubMenu.tsx:42]: This effect is breaking my brain a bit. Shouldn't `mounted` always be `true` when the effect is called? What does setting it to `false` within the closure do?

3mcd (approved):
Looks good. Just one question about a `useEffect` that I don't quite understand.

core/app/c/[communitySlug]/NavLinkSubMenu.tsx
Comment (tefkah) [reply core/app/c/[communitySlug]/NavLinkSubMenu.tsx:35]: ugh, sorry that comment took a while, i always accidentally add a "review" and then don't submit it 😅

PR #755, opened 2024-11-05
Read from new memberships tables

### LINKED ISSUE #718: Modify membership tables
## Motivation
- So developers have a simpler way of checking permissions

## Requirements
- Write tests for code paths expected to change as a result
- Change permission model to have memberships to communities, stages, or pubs with a role, userID "foreign key'd" to userID, rather than permissions as described in ACs.
- Update functions that user memberID to use userID first and then fall back to memberID
- Migrate PubValues that contain memberID to userID

## Acceptance Criteria
- Add an initial PR to update code paths that query memberIds to make them fall back to userId
- Create a migration that does the following:
	- Alter form_to_permission table
		- rename to form_permissions
		- replace permissionId with nullable userId (migrate data)
		- add nullable memberGroupId
		- add role, set to contributor for now (maybe even default? we won't be using these roles in code for now)
	- Drop permissions table (this has to happen before updating the members table, since it references the primary key members.id that we are dropping)
	- Add pub_permissions(pubId, memberGroupId, userId), stage_permissions(stageId, memberGroupId, userId) tables with composite keys
	- Alter members table:
		- Rename to community_permissions
		- Update primary key to be composite of (userId, memberGroupId, communityId)
		- Drop id column
	- Add constraints to prevent both userId and memberGroupId on one row
- Rename tables in queries/update prisma queries that reference old names
- After the first two PRs are merged and migrations successfully completed, change memberId functions to only query on userId

kalilsn:
## Issue(s) Resolved
#718 
This is part 2 of our expand and contract migration series, to be merged only after #1 is successfully deployed.

Now that we've made sure the application is writing to the new tables (in addition to the old ones) we can copy data into the new tables and be certain that they will be in sync with the old tables as part of the migration. Since the code changes in this PR will be deployed only after the migration succeeds, we can then immediately start reading exclusively from the new tables, which is what the rest of the changes in this PR accomplish:
- There should be no more queries reading from the `members`, `form_to_permissions`, and `permissions` (writing is still ok)
- Explicit typings using the generated `Members` types have been replaced with `CommunityMemberships` equivalents

### DIFF
--- a/core/lib/server/stages.ts
+++ b/core/lib/server/stages.ts
@@ -1,7 +1,6 @@
 import { sql } from "kysely";
 import { jsonArrayFrom, jsonObjectFrom } from "kysely/helpers/postgres";
 
-import type { JsonValue } from "contracts";
 import type {
 	CommunitiesId,
 	NewMoveConstraint,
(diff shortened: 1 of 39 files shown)


core/lib/server/stages.ts
```diff
@@ -15,134 +14,6 @@ import type { AutoReturnType } from "../types";
...
-					eb
-						.selectFrom("action_instances")
-						.whereRef("action_instances.stageId", "=", "stages.id")
-						.selectAll()
-						.$narrowType<{ config: JsonValue }>()
-				).as("actionInstances"),
-			])
-			.selectAll("stages")
-			.orderBy("order asc")
-	);
```
kalilsn, line 145: this query is unused (superseded by a later function in this file) so I deleted it instead of fixing it

tefkah: I think that's fine, I don't think we intend to use it in the short term

core/prisma/migrations/20241105184158_copy_membership_data/migration.sql
```diff
@@ -0,0 +1,14 @@
...
+    SELECT "id", "role", "communityId", "userId", "createdAt", "updatedAt" FROM "members"
+    ON CONFLICT ("id") DO NOTHING;
+
+INSERT INTO "form_memberships" 
+    SELECT "formId", "userId" FROM "form_to_permissions"
+    JOIN "permissions"
+        ON "permissions"."id" = "form_to_permissions"."permissionId"
+    JOIN "members"
+        ON "members"."id" = "permissions"."memberId";
+
```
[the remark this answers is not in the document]
tefkah: great, good to start from a fresh slate there.

core/prisma/migrations/20241105184158_copy_membership_data/migration.sql
```diff
@@ -0,0 +1,14 @@
+BEGIN;
+
+INSERT INTO "community_memberships"
+    SELECT "id", "role", "communityId", "userId", "createdAt", "updatedAt" FROM "members"
+    ON CONFLICT ("id") DO NOTHING;
```
tefkah, line 5: the conflicts you are anticipating here are community_memberships that would have been created in tandem with "legacy" memberships due to the previous pr, correct?

core/lib/server/stages.ts
```diff
@@ -216,13 +87,11 @@ export const getCommunityStages = (communityId: CommunitiesId) =>
...
-					.selectFrom("permissions")
-					.innerJoin("_PermissionToStage", "permissions.id", "_PermissionToStage.A")
-					.innerJoin("members", "_PermissionToStage.B", "members.id")
+					.selectFrom("stage_memberships")
 					.select((eb) =>
 						eb.fn
-							.count("_PermissionToStage.A")
-							.filterWhereRef("_PermissionToStage.B", "=", "stages.id")
+							.count("stage_memberships.userId")
+							.filterWhereRef("stage_memberships.stageId", "=", "stages.id")
```
tefkah, line 94: so much clearer!

core/prisma/migrations/20241105184158_copy_membership_data/migration.sql
kalilsn: exactly. not a super likely scenario but it could happen!

3mcd (approved):

PR #764, opened 2024-11-07
feat: add functions to interact with relationships

### LINKED ISSUE #733: Add internal utils for creating, updating, and deleting relationship
## Motivation

So developers have a way to easily create relationships on their own (currently only possible when creating a new pub)

## Requirements

- Ability to create a relationship between two pubs
- Ability to update the value of a relationship between two pubs
- Ability to delete a relationship between two pubs
- Should only allow creation of relationships for which a relationship field already exists

## Acceptance Criteria

- Function to create a pub and immediately add it as a relationship to a different pub in one transaction
- Function to create a relationship between two existing pubs, optionally with a specified value
- Function to update the value of a relationship field given a `relatedPubId`.
- Update the existing `updatePub` function to not be able to update relation fields

tefkah:
Wanted to get this in before i left on holiday.

No real time to write a proper PR, if the code is straightforward enough to understand/merge/continue working on it, please do! If not, I can fix all feedback once i'm back. See you in a week!

## Issue(s) Resolved
#733 

## Test Plan
1. see tests.


packages/contracts/src/resources/integrations.ts
```diff
@@ -21,7 +23,7 @@ export const User = SafeUser.and(
 export type User = z.infer<typeof User>;
 
 // Json value types taken from prisma
-export type JsonObject = { [Key in string]?: JsonValue };
```
tefkah, line 24: this was needed bc we cast one to the other sooo frequently.

core/lib/server/pub.ts
```diff
@@ -601,6 +606,383 @@ export const deletePub = async (pubId: PubsId) =>
...
+
+		const relationsWithSlug = slugs.flatMap((slug) =>
+			relations[slug].map((relation) => ({ ...relation, slug }) as AddPubRelationsInput)
+		);
+
+		await upsertPubRelations({ pubId, relations: relationsWithSlug, communityId, trx });
+	});
+};
+
+export const updatePub = async ({
```
tefkah, line 906: i did change how this works, see individual commits/compare with old _updatePub. mainly the upserts are not by id, rely on the unique index

core/lib/server/pub.ts
```diff
@@ -601,6 +606,383 @@ export const deletePub = async (pubId: PubsId) =>
...
+							pubId,
+							fieldId,
+							value: JSON.stringify(value),
+						}))
+					)
+					.onConflict((oc) =>
+						oc
+							// we have a unique index on pubId and fieldId where relatedPubId is null
+							.columns(["pubId", "fieldId"])
+							.where("relatedPubId", "is", null)
```
kalilsn, line 982: very nice, i tried this before but couldn't get it to work since i was specifying the name of the index. i now know that you have to specify the columns and condition for postgres to find a partial index in an on conflict statement!

tefkah: lot of trial & error 😅

core/lib/server/pubFields.ts
```diff
@@ -15,56 +39,11 @@ import { autoCache } from "./cache/autoCache";
...
-		  }
-		| { pubId?: never; pubTypeId?: never; communityId: CommunitiesId; includeRelations?: never }
-		| {
-				pubId?: never;
-				pubTypeId?: never;
-				communityId: CommunitiesId;
-				includeRelations: boolean;
-		  } = ({} = {})
-) => autoCache(_getPubFields(props));
+export const getPubFields = (props: PubFieldsInput) => autoCache(_getPubFields(props));
```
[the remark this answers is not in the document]
tefkah: I agree, I'm not a huge fan of it either. I think you're first interpretation is correct, i.e. a private function that's usually not exported, only done so for testing purposes. But it's indeed somewhat unclear, we've also used it in the past for `_updatePub`, with the idea that the underscored version was the "non server action" version. This was incorrect though, as it was still exported from a file with `use server`, making it a server action as well.

I think a more descriptive name is probably better. We could still use `_` to indicate "this function should probably not be used outside of this module", but the function could be something like `_uncachedGetPubFields`.

core/lib/server/validateFields.ts
```diff
@@ -27,10 +27,42 @@ const validateAgainstContextEditorSchema = (value: unknown) => {
...
+
+	return errors;
+};
+
 /**
  * Validate all `values` against their corresponding field's `schemaName`.
+ * @deprecated
  */
-export const validatePubValuesBySchemaName = ({
+export const __validatePubValuesBySchemaName = ({
```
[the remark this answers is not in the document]
tefkah: i've updated the function name to be more clear about what i meant.
maybe i should just replace the usage of it everywhere instead? It doesn't do the exact same as the `validatePubValuesBySchemaName` though, hence why I hesitated to do this.

This function did a bit too much imo and had clunky usage for the caller. It both validated that the passed values match the schemas of the fields passed, but also that the values are only of the SchemaNames of those in the fields. This means that the caller needs to do a lot of weird matching and preprocessing which is very prone to error.

In the new function, you need to pass the schemaname together with the value it applies to, so this matching problem does not occur. However this does remove some functionality, namely the "making sure the values are only of the schemanames specified in the passed fields". It is on the caller to fix this before calling this function. Maybe we could provide a similar helper function specifically for this purpose, but I didn't want to make this PR even larger, so I just left the old function in and renamed it to signal we should fix this at some point.

core/lib/server/pub.ts
```diff
@@ -601,6 +607,401 @@ export const deletePub = async (pubId: PubsId) =>
 export const getPubStage = async (pubId: PubsId) =>
 	autoCache(db.selectFrom("PubsInStages").select("stageId").where("pubId", "=", pubId));
 
+/**
+ * Consolidates field slugs with their corresponding field IDs and schema names from the community.
+ * Validates that all provided slugs exist in the community.
+ * @throws Error if any slugs don't exist in the community
+ */
+const consolidateFieldSlugs = async <T extends { slug: string }>({
```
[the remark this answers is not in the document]
tefkah: hmm i agree that this function might be doing too much, but i was motivated by trying to avoid the pitfalls of eg the old `validatePubValues` function(s). Because in that function, the field info is sort of decoupled from the value info, it's on the caller to mix and match them appropriately. With this function, you don't need to worry about that as the caller, you just supply the pubvalues and get the thing that you want, namely the hydrated pubvalues (which are necessary for validating said pubvalues).

Do you feel that maybe the new validation fn I wrote (see below) should work differently as well, e.g. it should not take an array of "hydrated" pubvalues with their field info? Because if you (like me) think that this new fn is easier to use, I think it makes sense to use this corresponding "hydrating" function to get your values in the right shape. Otherwise the caller still needs to do a bunch of custom preprocessing every time to "hydrate" the pubvalues, eg by calling a version of this fn that does what you describe, take an array of slugs and get their field info, and then having to "manually" match those hydrated slugs with the pubvalues.

But i agree that this all feels very annoying and not quite right. i feel like our "api" here is not good enough, and we should have a clearer way of "just doing what we want", namely validating that the pubvalues pass validation and exist in the community.

https://github.com/pubpub/platform/blob/8c310efd64d913b0d6c6b6966755552ae0681e72/core/lib/server/validateFields.ts#L26-L55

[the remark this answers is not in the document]
3mcd: And sorry, to respond directly to your question, I don't think there's anything wrong with tacking on the `slug` property to the results returned by `consolidateFieldSlugs` if they would be needed downstream by the validation fn. I just don't see the need to attach all the additional properties from the objects originally provided to `consolidateFieldSlugs`.

core/lib/server/pub.ts
```diff
@@ -406,65 +415,22 @@ export const createPubRecursiveNew = async <Body extends CreatePubRequestBodyWit
...
-
-		const valueMaybeWithRelatedPubId =
-			value && typeof value === "object" && "value" in value
-				? { value: JSON.stringify(value.value), relatedPubId: value.relatedPubId as PubsId }
-				: { value: JSON.stringify(value) };
-
-		return {
-			id: field.id,
-			slug: field.slug,
-			...valueMaybeWithRelatedPubId,
```
tefkah, line 461: i also updated `createPubRecursiveNew` to use the new `validatePubValues` function

core/lib/server/pub.ts
```diff
@@ -541,55 +508,26 @@ export const createPubRecursiveNew = async <Body extends CreatePubRequestBodyWit
...
+			relations: Object.entries(body.relatedPubs).flatMap(([fieldSlug, relatedPubBodies]) =>
+				relatedPubBodies.map(({ pub, value }) => ({
+					slug: fieldSlug,
+					value,
+					relatedPub: pub,
+				}))
+			),
+			communityId,
+			trx,
+		});
```
tefkah, line 523: i made `createPubRecursiveNew` use the new `upsertPubRelation` function, slightly DRYer

Issue #638, opened 2024-09-23
Membership Epic


gabestein:
## Overview
Membership on Platform uses the existing action framework to allow Admins to create complex user roles without needing to manage overlapping sets of permissions. Instead, the idea is that Admins can give Members (sometimes via Groups) the ability to edit fields on Pubs — either directly or by giving them access to all Pubs in a Stage (which also gives them the ability to take actions on Pubs in those stages).

Users can be added in three places: Community-level, Stage-level, Pub-level. In general, these levels govern what they have access to based on the other dimension, their role.

We have three user roles for each level, which have these general features across the Community, Pubs, and Stages:
- Contributor: Can only edit fields/take actions on Pubs/Stages which they have explicit access, even at the community level. For people like submitters, reviewers, etc.
- Editor: Can edit fields on pubs they have access to and their children/siblings, and run available actions on those Pubs (even if not a member of the stage). At the community level, these people can edit, but not configure/invite, everything. Generally for people who need access to groups of pubs like copyeditors, section editors, book editors, etc.
- Admin: Same as editor, plus can configure actions on stages and invite members to the objects they have access to (or all objects, if at the community level).

## Tickets
- [ ] #656 
- [x] #595
- [x] #594
- [ ] #596
- [x] #637
- [ ] #639

Issue #594, opened 2024-09-09
Add basic Pub membership UI


gabestein:
## Motivation

- So admins can assign Members to individual pubs
- So admins can create reviews of pubs as children or siblings that their submitters, as contributors, can't see (relies on conceit that any form/editor that creates children will add the creator as admins of those child pubs -- seems reasonable?)
- So admins can assign editors who are responsible for managing pubs and their children without giving them deletion permissions

## Requirements

- Ability to add/remove existing community members to a Pub as an admin, contributor, or editor in the stage member tab, and invite new ones.
  - Add "add member" button to member section of the sidebar, identical to "members" page add member button
  - Only admins should see the button
  - Should repurpose the existing member add UI, e.g.:
    - When clicked, shown a popup to search by email
    - If user found, can select permission for the stage and add it
    - If no user found, sends an invitation. That invitation should use slightly different language, e.g.: `You have been invited to be a [role name] of the [stage name] stage in [community name].`
- Ability to see existing members and their role, and delete them in the member section of the pub sidebar
  - Only admins should be able to see the delete button
  - Can just be the current member list, plus their role underneath their name and an x to delete them, similar to the current pub type editor

## Acceptance Criteria
- Does not implement permissions, that's a separate card, but should add/remove the correct permissions
- Does not need to show cascading permissions yet, we can add that later

## Mockups (if available)
![Image](https://github.com/user-attachments/assets/fa3224c3-3b85-4119-b987-9a7a56c3dfe9)

<details>
  <summary>Old</summary>

Permissions behavior:
- Contributor:
  - Can see and edit pub fields
  - Can see what stage a pub is in
  - Can create child/related Pubs (they become direct members of these pubs)
  - Cannot take actions, move, or assign
- Editor: same as contributor, plus:
  - Can see and edit children and siblings (even if not directly added)
  - Can move, assign, and run actions on pubs (even if they're not stage members)
- Admin: same as editor, plus:
  - Can delete the pub
  - Can add members to the pub as contributor, editor, or community admin. (If the user does not exist, admin can invite them via email. The user receives an invitation to join PubPub as a result, with a message specifying the stage they have been invited to and a redirect to the stage configuration they have been added to.)

</details>

3mcd:
This all seems right to me, but I had one thought: what about adding fields to and removing fields from a pub via the UI (i.e. not via actions; including relationships)? Maybe we could restrict these types of edits to editor+.

kalilsn:
[...]

PR #766, opened 2024-11-11
Add pub and stages capabilities query

### LINKED ISSUE #671: Write function(s) that query pub or stage privilege
## Motivation
So developers have a way to check if a user can access or perform some action upon a pub or stage.

## Requirements

- Create a method that developers can use to determine if a community member has access to a certain resource or action.
- Function should traverse entire graph of permissions to find the highest permission at the highest level for the user (e.g. if user is a pub contributor but stage admin for the stage the current pub is in, they should have admin permissions)

## Acceptance Criteria

- A function (or functions) should be created that accepts three arguments:
  1. the platform resource (i.e. a pub or stage)
  2. a community member object
  3. some privelege associated with the resource (defined in #670)
- The function should inform the caller (e.g. by returning a boolean) of whether or not the community member has been granted permission that would authorize the requested access or action.

kalilsn:
## Issue(s) Resolved
Resolves #671
This PR adds a new database table and query to manage authorization for every kind of action a user might take on a stage or pub (forms and communities not yet included). It's not used anywhere besides the tests I wrote for it, which focus on the capabilities granted by community membership, since that's the only kind we can actually manage right now. There's no in-app way to add a stage or pub membership yet, so I'll add tests for those when it's possible. And while we can invite users to a form, how those permissions should work still needs some speccing out.

I chose to design the query/db structure the way I did because we know that we'll soon need to use this query to filter a tree of pubs to only include the ones the user is authorized to see. And to do that efficiently, we need to be able to calculate the permissions a user has entirely in the database. As a result, the definition of which capabilities are granted by any kind of membership has to be stored in the database, or else we'd have to serialize it with every query. We do take the [serialization approach](https://github.com/pubpub/platform/blob/main/core/app/c/%5BcommunitySlug%5D/forms/actions.ts#L19-L60) in the createForm query, but that seems less desirable in this query because we'll be running it more frequently and the serialized data would be much larger. Instead, I've opted to move our capabilities enum definition to the database, and add the capabilities mapping to the new `membership_capabilities` table in a migration. This is a little awkward, since we don't really want that data to change and technically one of us could modify the database directly in a way that would change how our permissions work. But that should be pretty easy to avoid! Since this data should never be updated dynamically, I think we should set a convention that when we update it in future migrations we add the entire table (truncating first or using `on conflict`) to make sure there's a single spot in the codebase where we can see what the current state should be.

A few more notes:

- There's a lot of noise in this PR, because I renamed our `auth` module to `authentication` for clarity and that updated a lot of imports. That's really the only thing that should have changed in `core/app`, so I would recommend skimming those changes! The real changes are in `core/lib/authorization`, `core/prisma`, and `packages/db`. 
- I had to do some kind of ugly work with the .kanelrc and the zod compatibility hook to make my new table's types generate correctly. LMK if you think there's a better way to do it
- There's definitely room for some refactoring improvements on the query. But I think leaving it as is until we add pub filtering, community and form capabilities, as well as member groups makes more sense than refactoring now.


core/lib/authorization/capabilities.db.test.ts
```diff
@@ -0,0 +1,113 @@
+import { describe, expect, test } from "vitest";
+
+import { CoreSchemaType, MemberRole } from "db/public";
+import { Capabilities } from "db/src/public/Capabilities";
+import { MembershipType } from "db/src/public/MembershipType";
+
+import { seedCommunity } from "~/prisma/seed/seedCommunity";
+import { mockServerCode } from "../__tests__/utils";
+
+const { createForEachMockedTransaction } = await mockServerCode();
```
kalilsn, line 10: This test currently fails because this doesn't seem to be mocking React.cache, though I have no idea why.
image

kalilsn: The test passes if I put it in the pub.db.test.ts file, so I think I'm missing some setup step here but I have no idea which

core/prisma/migrations/20241107193617_add_capabilities_to_db/migration.sql
```diff
@@ -0,0 +1,147 @@
...
+    ('editor'::"MemberRole", 'community'::"MembershipType", 'movePub'::"Capabilities"),
+    ('editor'::"MemberRole", 'community'::"MembershipType", 'createPub'::"Capabilities"),
+    ('editor'::"MemberRole", 'community'::"MembershipType", 'viewPub'::"Capabilities"),
+    ('editor'::"MemberRole", 'community'::"MembershipType", 'deletePub'::"Capabilities"),
+    ('editor'::"MemberRole", 'community'::"MembershipType", 'updatePubValues'::"Capabilities"),
+    ('editor'::"MemberRole", 'community'::"MembershipType", 'createRelatedPub'::"Capabilities"),
+    ('editor'::"MemberRole", 'community'::"MembershipType", 'createPubWithForm'::"Capabilities"),
+    ('editor'::"MemberRole", 'community'::"MembershipType", 'editPubWithForm'::"Capabilities"),
+    ('editor'::"MemberRole", 'community'::"MembershipType", 'runAction'::"Capabilities"),
+    ('editor'::"MemberRole", 'community'::"MembershipType", 'viewStage'::"Capabilities"),
```
kalilsn, line 111: Probably no need to review below here, since community admins get every capability

core/prisma/migrations/20241107193617_add_capabilities_to_db/migration.sql
```diff
@@ -0,0 +1,147 @@
...
+CREATE TABLE "membership_capabilities" (
+    "role" "MemberRole" NOT NULL,
+    "type" "MembershipType" NOT NULL,
+    "capability" "Capabilities" NOT NULL,
+
+    CONSTRAINT "membership_capabilities_pkey" PRIMARY KEY ("role","type","capability")
+);
+
+INSERT INTO "membership_capabilities"
+VALUES
```
kalilsn, line 53: @gabestein the table below maps role and membership type to capabilities for stages, pubs, and communities. Could you review to make sure there's no missing or extra permissions for various roles? I also think I've missed a few required capabilities, which you're welcome to note here or we can just add them along with the features that require them.

core/lib/authorization/capabilities.db.test.ts
kalilsn: turns out this was because I had imported `mockServerCode` after `seedCommunity` and it has to come first! Probably this means we should be dynamically importing `seedCommunity` in these tests, but for now I just switched the import style from relative to using our ~/lib alias, which makes the mockServerCode import sort first.

core/lib/authorization/capabilities.ts
```diff
@@ -0,0 +1,211 @@
...
+		const capabilitiesQuery = db
+			.with("stage", (db) =>
+				db
+					.selectFrom("PubsInStages")
+					.where("PubsInStages.pubId", "=", target.pubId)
+					.select("PubsInStages.stageId")
+			)
+			.with("community", (db) =>
+				db.selectFrom("pubs").where("pubs.id", "=", target.pubId).select("pubs.communityId")
+			)
```
kalilsn, line 96: For a pub, we first get its stage and community because pub capabilities can be granted by stage or community memberships. Stage capabilities can only be granted by stage or community membership so that query below is slightly simpler (like the community and forms versions will be).

core/lib/authorization/capabilities.ts
```diff
@@ -0,0 +1,211 @@
...
+};
+
+type FormTarget = {
+	type: MembershipType.form;
+	formId: FormsId;
+};
+
+export const userCan = async <T extends Target>(
+	capability: CapabilitiesArg[T["type"]][number],
+	target: T,
```
kalilsn, line 83: There's a distinction between capabilities that's not really captured in the database structure. `updatePubValues` is a pub capability because its target is a pub, meaning we need a pubId to determine if the user has that capability. `viewStage` is a stage capability so it requires a stageId. The types here make sure that you'll get a type error if you call `userCan` with a pub capability but pass a stageId or any other invalid combination.

kalilsn: There's one very misleading thing here, which is my use of MembershipType to express both the type of membership the user has, and the type of target a certain capability requires. At the moment those are both `pub`, `stage`, `community`, or `form`, but that's kind of incidental! Let me know if that makes sense and/or if it's troubling.

packages/db/.kanelrc.js
```diff
@@ -16,6 +22,38 @@ module.exports = {
...
+			type = innerType.name;
+			imports.push(...innerType.typeImports);
+		}
+
+		let typeDefinition = [`${type} & { __brand: '${escapeString(name)}' }`];
+
+		/* this is the custom part */
+		if (column.type.kind === "enum") {
+			typeDefinition = [`${type}`];
+		}
```
kalilsn, line 46: Without this change, kanel generates branded types for the three id columns in `membership_capabilities`. That's not really correct IMO since the point of making those columns use enums is that they are _not_ unique to the table, and are in fact equivalent to any other column with that same enum type.

core/lib/authorization/capabilities.ts
allisonking: I'm not sure I follow—when would they be different?

allisonking (approved):
great tests!! I'm not sure I follow the kanel stuff though so can't really help with that...

core/lib/authorization/capabilities.ts
kalilsn: I think you're right, they wouldn't be. If we were to add a new kind of entity and associated capabilities for it but didn't want to add a new membership type, that's when they'd be different. But I don't think that's very likely!

core/prisma/migrations/20241107193617_add_capabilities_to_db/migration.sql
gabestein: Thanks! I only see a few things:

- Pub Contributors should be able to create related pubs, per the "old" tab in #594 .
- Pub Editors should be able to move and take actions, per the "old" tab in #594 .
- Pub Editors should be able to see all related pubs, per the "old" tab in #594 (not sure if that'll be handled here?)
- I'm not positive Stage Contributors should be able to `editPubWithForm` if we're going to implement form permissions separately?
- Stage Editors should be able to see all pubs related to pubs in the stage, per the "old" tab in #638 (not sure if that'll be handled here)

kalilsn: great, thanks for checking. i'll make these updates alongside my next PR

PR #774, opened 2024-11-12
Add on conflict clause to form_memberships copy

kalilsn:
## Issue(s) Resolved
Another error in the migration from #755 
There's no error in the logs because [prisma hides migration errors if you use a transaction](https://github.com/prisma/prisma/issues/15295). But running the sql against prod manually with `psql` showed an error:
```
/usr/src/app/core # psql -f prisma/migrations/20241105184158_copy_membership_data/migration.sql
BEGIN
INSERT 0 126
psql:prisma/migrations/20241105184158_copy_membership_data/migration.sql:12: ERROR:  duplicate key value violates unique constraint "form_memberships_formId_userId_key"
DETAIL:  Key ("formId", "userId")=(d6aa0d8a-25ce-4ff5-b2ac-0bc6d3b91311, 9fb919ac-c317-414e-a8b9-f4d3d233b8c6) already exists.
ROLLBACK
```

Much like the copied memberships, there will already be some form_memberships from users who were invited to forms since #728 was merged, so we need to skip copying those ones or else we get a unique constraint error.

### DIFF
--- a/core/prisma/migrations/20241105184158_copy_membership_data/migration.sql
+++ b/core/prisma/migrations/20241105184158_copy_membership_data/migration.sql
@@ -9,7 +9,7 @@ INSERT INTO "form_memberships" ("formId", "userId")
     JOIN "permissions"
         ON "permissions"."id" = "form_to_permissions"."permissionId"
     JOIN "members"
-        ON "members"."id" = "permissions"."memberId";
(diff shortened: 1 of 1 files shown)


3mcd (approved):

allisonking (approved):
🤞

Issue #740, opened 2024-10-24
Allow API to use membership permission system


gabestein:
## Motivation

- So developers can use the API internally for, e.g., creating new related pubs in the editor

## Requirements

- Ability to resolve user capabilities from their login session instead of a token, and have the API work appropriately
- Ability to use the `/site` API from within `core`, rather than only using it externally.

## Acceptance Criteria

- API should read login information using `getLoginData` if no bearer token is detected.
- The permission check should work the same as our internal permission checks work (e.g. inside of a server action). This is conditional on how #671 is implemented.
  - Roughly, you should be able to specify some priviliges on a route, and if the user that hits the route does not meet them, throw some error
  - Possibly we could specify these authorization requirements in a layer that is shared between e.g. server actions and the API, but for now duplicating it is fine.

3mcd:
@kalilsn I just updated that requirement to say

> Ability to resolve user capabilities from their login session instead of a token, and have the API work appropriately

I'm not sure if it's essential that this ticket covers the pub access control/filtering within the API. Thoughts @gabestein ?

kalilsn:
Thanks, that makes sense to me now!

After looking at the existing API endpoints, I think we have most of the capabilities we need. But I think we need one for `viewCommunityPubTypes` or maybe a general `viewCommunityConfig` that includes reading but not writing pub types, pub fields, and forms. What memberships should grant that capability?

I also think we'll need a `viewRelatedPubs` capability, although I'm a little fuzzy on precisely how that should work. It might be complicated to implement that, in the same way that checking authorization for the getMany endpoints (efficiently) might be complicated.

kalilsn:
In the old requirements of https://github.com/pubpub/platform/issues/595 it says that stage editors (and stage admins)
>Can see and edit children/related pubs of Pubs in the stage, even if child/related Pubs are not in the stage

That's why I was thinking we'd need the `viewRelatedPubs` permission

gabestein:
> I'm not sure if it's essential that this ticket covers the pub access control/filtering within the API

That's right. We're doing this so we can create related pubs from the editor, mainly, which will have the context of a user with existing permissions. If we need to add the ability to do that, we can later.

> I'm not totally sure if we need a specific capability for viewing related pubs. IMO related pubs should be filtered in the exact same way as the requested pub. That is, if you have the ability to see a pub, it should appear in any of its subject pubs' relationships.

[...]

PR #804, opened 2024-11-21
Add pub and stage membership management

### LINKED ISSUE #594: Add basic Pub membership UI

### LINKED ISSUE #595: Add basic Stage membership UI

kalilsn:
## Issue(s) Resolved
Resolves #594 
Resolves #595 

- Refactors the add member dialog to work with pub and stage memberships
- Refactors the add member dialog to no longer be controlled through url path and query parameters
	- 	Instead of fetching suggested users in a server component and rerendering, it now uses [React Query](https://tanstack.com/query/v5/docs/framework/react/quick-start) through the [ts-rest generated client](https://ts-rest.com/docs/react-query/v5-setup) to request suggestions via fetch + an api endpoint
	- This should make it more portable and more performant
	- It's no longer possible to link directly to the member add dialog, and it won't keep state after a refresh, which I think are unnecessary features. LMK if not and I can bring them back for the members page only.
- Adds the ability to change the role and remove members from the existing pub and stage member lists
- Modifies the user search endpoint to only return a user when the email matches exactly. This was always the intention from a privacy perspective, but also fixed a bug which could in some cases prevent inviting a new user if their email partially matched an existing one.

## Test Plan
- As jill admin:
   - Visit the members page and ensure you can still invite and remove members as usual
   - Visit the stage editor, open a stage, and select the stages panel.
   - Click add member and try adding a member that exists in the community
   - Click add member and try adding that member again. You shouldn't be able to
   - Click add member and try adding a member that doesn't exist in the community. Check inbucket to make sure the invite email looks ok. The signup process should work, although they won't be a member of a community so don't expect to be able to navigate beyond the settings page after logging in.
   - Make sure you can delete a member and change their role
   - Visit an individual pub and run the same tests!
- as some@pubpub.org:
    -  Visit the same individual pub that you added some users to above. You should see the members, but not be able to edit them
   
## Screenshots

Stage member management
image

Pub member management
image

Pub members in readonly mode
image

Side by side comparison of width change on pub page:
image


packages/contracts/src/resources/site.ts
```diff
@@ -158,11 +168,32 @@ export const siteApi = contract.router(
...
+		},
 	},
 	{
 		pathPrefix: "/api/v0/c/:communitySlug/site",
 		baseHeaders: z.object({
-			authorization: z.string().regex(/^Bearer /),
+			authorization: z
+				.string()
+				.regex(/^Bearer /)
+				.optional(),
```
kalilsn, line 196: I'm not sure if making this optional was the right call, LMK if that breaks something. I assume #740 will fix it either way!

core/app/components/Memberships/MemberInviteForm.tsx
```diff
@@ -27,48 +27,69 @@ import { Input } from "ui/input";
...
-		if (state.state === "initial") {
-			return;
+	useEffect(() => {
+		if (userIsAlreadyMember) {
+			form.setError("email", {
+				type: "alreadyMember",
+				message: "This user is already a member",
+			});
 		}
+	}, [userIsAlreadyMember, form.setError, user, email]);
```
kalilsn, line 89: For some reason the form successfully detects if a user is already a member when you enter their email the first time, but if you then edit the email and then edit it back to the user's email, the form lets you add them. I handled that on the backend, so it's not a big deal, but I'm not sure why it's happening. At first I figured I missed a dependency which is why there are these extras (I will clean that up before merging since it didn't help).

kalilsn: I also didn't want to manage this error manually but I didn't find any other approach that worked. I tried validating it through the `rules` prop on the `<FormField>` component but apparently those don't run if you use a resolver. I also tried adding it as a `.refine` rule to the resolver dynamically in the useform call above (like we do in the `InputComponentConfigurationForm`) but I ended up with infinite rerenders that I couldn't figure out how to fix.

core/app/c/[communitySlug]/pubs/[pubId]/page.tsx
```diff
@@ -128,7 +148,7 @@ export default async function Page({
 							);
 						})}
 				</div>
-				<div className="flex w-64 flex-col gap-4 rounded-lg bg-gray-50 p-4 font-semibold shadow-inner">
+				<div className="flex w-96 flex-col gap-4 rounded-lg bg-gray-50 p-4 font-semibold shadow-inner">
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:151]: I made this wider to fit the member management interface. It's about the same width as the stage panel now

kalilsn:
I'm not sure why the tests are failing here. They pass on my computer (including against the production build with `CI=true`) but in CI it seems like the login form isn't working. I would love some help here since I don't think I changed anything that would affect the login form!

core/.env.test
```diff
@@ -1,5 +1,6 @@
 DATABASE_URL=postgresql://postgres:postgres@localhost:5433/postgres
 PUBPUB_URL=http://localhost:3000
+NEXT_PUBLIC_PUBPUB_URL=http://localhost:3000
```
tefkah, line 3: was this to debug e2e failure maybe? afaik we're not using `NEXT_PUBLIC` env vars anymore

tefkah: that also seems to be fine

core/lib/api.ts
```diff
@@ -0,0 +1,9 @@
+import { initTsrReactQuery } from "@ts-rest/react-query/v5";
+
+import { siteApi } from "contracts";
+
+import { env } from "./env/env.mjs";
+
+export const client = initTsrReactQuery(siteApi, {
+	baseUrl: env.NEXT_PUBLIC_PUBPUB_URL,
```
tefkah, line 8: I suggest we don't use a public env var, and just rely on the window url

```suggestion
	baseUrl: typeof window === "undefined" ? env.PUBPUB_URL : window.location.origin, 
```

this is slightly more jank, and i'm not sure the `env.PUBPUB_URL` will transfer properly or will even ever get used, but this does work. we could also just make it `? "/" : ...` or something instead of `env.PUBPUB_URL`

core/.env.test
kalilsn: Yeah I added it because I'm using it, but your way sounds better so I'll fix it! Maybe this is what's making the e2e tests break? Also fwiw this change was an error, since .env.test isn't actually used when we run the playwright tests (only the vitest tests).

kalilsn: Actually this causes the build to fail. Did you have a way around that?

core/app/components/Memberships/MembersList.tsx
```diff
@@ -0,0 +1,58 @@
...
+	members,
+	setRole,
+	removeMember,
+	targetId,
+	readOnly,
+}: MembersListProps<T>) => {
+	return (
+		<>
+			{members.map((user) => (
+				<div key={user.id} className="flex items-center gap-4">
```
tefkah, line 19: these member lists look really good! really nice you can just edit the membership inline
this is super nitpicky so feel free to ignore, but i think they'd look slightly nicer if the badge was right aligned and slightly smaller! see this and the next comment

```suggestion
				<div key={user.id} className="flex items-center justify-between gap-4">
```

core/app/components/Memberships/MembersList.tsx
```diff
@@ -0,0 +1,58 @@
...
+							/>
+
+							<RemoveMemberButton
+								userId={user.id}
+								targetId={targetId}
+								removeMember={removeMember}
+							/>
+						</>
+					)}
+				</div>
```
tefkah, line 54:
```suggestion
					<div className="flex items-center gap-2">
						{readOnly ? (
							<span className="rounded-full bg-blue-500 px-3 py-2 text-sm font-medium capitalize text-slate-50">
								{user.role}
							</span>
						) : (
							<>
								<RoleSelect
									role={user.role}
									userId={user.id}
									targetId={targetId}
									setRole={setRole}
								/>

								<RemoveMemberButton
									userId={user.id}
									targetId={targetId}
									removeMember={removeMember}
								/>
							</>
						)}
					</div>
				</div>
```

core/app/components/Memberships/RoleSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+export const RoleSelect = <T extends TargetId>({
+	targetId,
+	role: currentRole,
+	userId,
+	setRole,
+}: RoleSelectProps<T>) => {
+	return (
+		<DropdownMenu>
+			<DropdownMenuTrigger asChild>
+				<Button className="rounded-full bg-blue-500 capitalize text-slate-50 hover:bg-blue-600">
```
tefkah, line 33: again tiny tiny thing, but i think it'd look better if it was a small button instead

```suggestion
				<Button
					size="sm"
					className="rounded-full bg-blue-500 capitalize text-slate-50 hover:bg-blue-600"
				>
```

packages/contracts/src/resources/site.ts
tefkah: ideally i'd do something like `z.union([authorization, cookies])` but that's not allowed unfortunately

tefkah (changes requested):
looks and works great! great job, this one seemed quite finicky

I do have a few suggestion, most of which are nits and can be ignored, except for the ones about `NEXT_PUBLIC_PUBPUB_URL`. ill have a look as to why that causes builds to fail, but using the method i suggested below i can build locally it seems

core/lib/api.ts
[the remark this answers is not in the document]
tefkah: this is outdated, but this does seem to build for me

core/.env.test
tefkah: yes, 

```
	baseUrl: typeof window === "undefined" ? env.PUBPUB_URL : window.location.origin,
```

core/lib/api.ts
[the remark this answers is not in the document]
kalilsn: oh right that should work!

core/lib/server/user.ts
```diff
@@ -169,3 +191,141 @@ export const addUser = (props: NewUsers, trx = db) =>
...
+				role: MemberRole;
+		  }
+		| { type: MembershipType.stage; role: MemberRole; stageId: StagesId }
+		| { type: MembershipType.pub; role: MemberRole; pubId: PubsId };
+}) => {
+	const { firstName, lastName, email, membership, isSuperAdmin } = data;
+
+	try {
+		const { user } = await getLoginData();
+		const community = await findCommunityBySlug(getCommunitySlug());
```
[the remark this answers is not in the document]
kalilsn: good catch, thanks!

kalilsn:
Great call on the right alignment and size @tefkah !

image

I think I did all your other requested changes too and that seems to have fixed the test failures, which must have been coming from my misunderstanding about whether we could use frontend env vars (we can't)!

.github/workflows/e2e.yml
```diff
@@ -119,3 +119,6 @@ jobs:
               run: pnpm --filter core playwright:test
               env:
                   INTEGRATION_TEST_HOST: localhost
+            - name: Print container logs
+              if: failure()
+              run: docker compose -f docker-compose.test.yml --profile integration logs
```
tefkah, line 124: very useful addition

packages/contracts/src/resources/site.ts
```diff
@@ -158,11 +168,32 @@ export const siteApi = contract.router(
...
+				email: z.string(),
+				name: z.string().optional(),
+				limit: z.number().optional(),
+			}),
+			responses: {
+				200: safeUserSchema
+					.extend({ member: communityMembershipsSchema.nullable().optional() })
+					.array(),
+			},
+		},
```
tefkah, line 188: very sorry, i overlooked this in my initial review, but could we make this endpoint slightly more in line with our other endpoints?

main things i'm thinking of

- instead of `searchUsers`, make it `users.search` (?).
- similarly, change the `path` to match, eg `/users/search`. 
- don't require the `communityId` in the query object. since the api routes are scoped to a community we can derive the communityid that way (see my comment in the route)

related to the above: do you think it makes sense to have separate `/users` and `/members` endpoints? I do think well add a way to get all (community) members in a community, but should that be `GET /members` or `GET /users`? We can decide that later i think, but interested in your thoughts

core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts
```diff
@@ -213,6 +215,22 @@ const handler = createNextHandler(
...
+					firstName: req.query.name,
+					lastName: req.query.name,
+				},
+				limit: req.query.limit,
+			}).execute();
+			return {
+				status: 200,
+				body: users,
+			};
+		},
```
Comment (tefkah) [inline core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts:233]: related to my comments on the contract (again apologies for not mentioning this earlier), I think we don't need to get the `communityId` from the query, but you can just do `const community = await findCommunityBySlug()`

Ideally we'd do some other kind of permission check (probably just that the user has some kind of membership in the community?), but we can do that after this PR gets in (or if you feel really strongly about it you can use the method implemented in #805, but i don't want to block this pr for that)

tefkah (approved):
Other than some nitpicks on specific API things it looks great! I would like you to change these suggestions ideally, but its not blocking! Great job!

packages/contracts/src/resources/site.ts
kalilsn: i'm happy to do whatever you think is best with the api! i named it this way because i figured it would behave a little differently than the other get endpoints, but i think users.search vs users.get captures that well enough. and i'll make the other changes!

kalilsn: i'm not really sure about members vs users. i want to modify the member schema type to store a userid in the future since not everyone will have a community membership (though I would like to change that too). but other than that i don't really know what else we'll be using these endpoints for yet

core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts
Comment (kalilsn) [reply core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts:583]: i think just checking that they're logged in is good enough for now, which is what i've done. eventually we should add a new capability for this i think

PR #805, opened 2024-11-25
feat: add permission authentication to api

### LINKED ISSUE #740: Allow API to use membership permission system

tefkah:
## Issue(s) Resolved
Resolves #740 

This PR adds the ability to authenticate API calls using the `userCan` function.

### DIFF
--- a/core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts
+++ b/core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts
@@ -19,8 +19,7 @@ import {
 	createPubRecursiveNew,
 	deletePub,
 	doesPubExist,
-	getPubCached,
-	getPubs,
+	ForbiddenError,
 	getPubsWithRelatedValuesAndChildren,
 	getPubType,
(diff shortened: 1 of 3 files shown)


kalilsn (approved):
looks great and i'd love to get this merged before my PR that uses the site API. IMO we can keep improving this as we go and add auth for the endpoints missing it once we have those capabilities sorted out

PR #809, opened 2024-11-26
dev: update "pub.updatedAt" whenever a Pubs PubValues are created/updated/removed

tefkah:
- **feat: update pubUpdatedAt when one of its pubvalues is updated**
- **dev: restore simpler behavior of trigger and add test for updatedAt behavior**

## Issue(s) Resolved
Allows us to much more easily sort Pubs by when they were updated.
It's best conceived as a follow up PR to #808 to make the `/pubs` page actually sort by `updatedAt`.

This PR adds a trigger to the `pub_values` table to update a Pub's `updatedAt` column to the current timestamp whenever one of its `pub_values` changes.

I think it's reasonable to consider that a Pub has been updated when one of its PubValues changes, as, with the deprecation of assignees, there are no real other columns on the `pubs` table that could really change. Therefore it seems reasonable to me that when asked the question "When was this pub last updated", you would say "Yeah I changed the title this morning, so this morning".

This allows us to get rid of custom logic like this

https://github.com/pubpub/platform/blob/00ae312b10ec1a49269724e91d6521f990b7762b/core/lib/server/pub.ts#L1330-L1334

which computes a Pubs `updatedAt` based on the `updatedAt` of its values. Notably, this previous approach missed the (admittedly rare) case in which a PubValue is removed entirely.

## Test Plan
1. Run tests (there is a test for this!)
2. Go to `/pubs`. 
3. Modify a Pubs values
4. See it jump to the top of `/pubs`
5. Change the assignee of another pub
6. See that Pub jump to the top of `/pubs`

## Notes

Should moving a pub to a different stage also be considered updating it? I don't think so in principle (as it's not the Pub that is changed), but in practice I could see that in most listings of Pubs sorted by `updatedAt` you would like to see recently moved pubs higher than non-recently moved pubs.

That is, imagine being an admin looking through all Pubs, say after coming back to your job after a break. You don't have a high volume of traffic in your journal, so you may only have a few updates a month.

If suddenly a Pub that hasn't been changed for ages was moved from "In Production" to "Published", and that was the only change made in your abscense, I would expect that Pub to be at the top of a list of Pubs sorted by "Updated At", even though no actual value was changed.

Maybe the way to accomplish that UX is not by also updating `pubs.updatedAt` when a pub is moved though, but by providing Activity Logs or something.


core/lib/server/pub-updatedAt-trigger.db.test.ts
```diff
@@ -0,0 +1,147 @@
...
+			body: {
+				pubTypeId: pubTypes["Basic Pub"].id,
+				values: {
+					[pubFields.Title.slug]: "Test pub",
+				},
+			},
+		});
+
+		const compareUpdatedAt = async () => {
+			const updatedAtBefore = pub.updatedAt;
```
kalilsn, line 79: Don't these successive `await compareUpdatedAt()` calls need to compare against the most recent `updatedAt` each time rather than the initial one from when the pub is created? Otherwise it only tests that the first pub value change fired the trigger.

So instead of using `const updatedAtBefore = pub.updatedAt`, we would pass that in as a parameter each time:

```ts
const compareUpdatedAt = async (updatedAtBefore: Date) => {
	const newPub = await testDb
		.selectFrom("pubs")
		.select("updatedAt")
		.where("id", "=", pub.id)
		.executeTakeFirstOrThrow();

	expect(newPub.updatedAt.getTime()).toBeGreaterThan(updatedAtBefore.getTime());
};
```

tefkah: shoot yeah you're right!

kalilsn: nice, good thinking returning the new date from the compare function!

kalilsn (approved):
🕐

Issue #777, opened 2024-11-14
Show Related Pubs on Pub Page


gabestein:
## Motivation

So users can see and click through to related Pubs on the individual Pub page

## Requirements

- On the Pub Page, add a new table called "Related Pubs" with tabs for each relationship type, with name of the relationship field as the title of the tab, and a new column that shows the related pubs pub type.
   - Table follows same logic as child pubs, displaying "Title" (with link to related Pub), "Pub Type", "Stage", "Created At", "Updated At", "Actions", and any member fields, if present.

## Extra work

- In main section of Pub page, related pub fields rendered as list of titles of related pubs, linked to those pubs

Issue #736, opened 2024-10-24
Update pages to use designated pub name field as title


gabestein:
## Motivation

- So users see the designated title in all relevant locations in the app
- So developers can use the title field in, e.g. the editor and the api

## Requirements

- Wherever a title is currently displayed, render the designated title field specified in the pub type (see: #735). Incomplete list for estimation purposes (dev to use callsites for determining where to use function):
   - All pubs page
   - Stages page
   - Workflow settings page in side panel
   - Individual pub page
   - Action log

## Acceptance Criteria

- Helper function for retrieving and rendering the title of a pub
   - If title field is rich text, should render as plain text
   - If the field is blank, should render "<Pub Type> created on <Month Day, Year>"
- Update pub title query sites to user helper function
- Write test that asserts that rendered pub title is the field returned
- Rewrite relevant pages that render pubtitles to use the new `getPubsWith...` function, and have that function fetch title info.

PR #810, opened 2024-11-26
Use `getPubsWithRelatedValuesAndChildren` on the pub page

allisonking:
## Issue(s) Resolved
First part of https://github.com/pubpub/platform/issues/777

In order to show related pubs, I'd like to use the new `getPubsWithRelatedValuesAndChildren` function. So this PR is for that first step—swapping the prisma query for the new query.

## Test Plan
* No regressions in the individual pub page


core/app/c/[communitySlug]/pubs/[pubId]/components/jsonSchemaHelpers.tsx
```diff
@@ -60,21 +60,15 @@ export function recursivelyGetScalarFields(schema: JSONSchemaType<AnySchema>, va
...
 }
 
-export function renderField(fieldValue: PubValueWithFieldAndSchema) {
-	const JSONSchema = fieldValue.field.schema
-		? (fieldValue.field.schema.schema as JSONSchemaType<AnySchema>)
-		: null;
-	const fieldTitle = (JSONSchema && JSONSchema.title) || fieldValue.field.name;
-	const renderedField = JSONSchema
-		? recursivelyGetScalarFields(JSONSchema, fieldValue.value)
-		: fieldValue.value && fieldValue.value.toString();
```
Comment (allisonking) [inline core/app/c/[communitySlug]/pubs/[pubId]/components/jsonSchemaHelpers.tsx:70]: I simplified this function to make the arguments it takes easier to work with. this whole thing will likely have to be reworked when we figure out how individual pub pages should be rendered. in any case, this part that looks at the `JSONSchema` wasn't being used since we use `schemaName` over `schema` now, so everything was always rendered as `toString()`

core/app/c/[communitySlug]/pubs/[pubId]/page.tsx
```diff
@@ -132,35 +125,29 @@ export default async function Page({
...
-							<div>
-								{pub.stages.map(({ stage }) => {
-									return (
-										<div key={stage.id} data-testid="current-stage">
-											{stage.name}
-										</div>
-									);
-								})}
-							</div>
-							{pub.stages[0] ? (
```
Comment (allisonking) [inline core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:144]: the new getPub query limits stage selection to 1, so we no longer have to work with indexing the first item in an array

core/lib/pubs.ts
```diff
@@ -1,16 +1,27 @@
 export interface PubTitleProps {
-	values: { field: { slug: string }; value: unknown }[] | Record<string, unknown>;
+	values:
+		| { field: { slug: string }; value: unknown }[]
+		| Record<string, unknown>
+		| { fieldSlug: string; value: unknown }[];
```
allisonking, line 5: 🙈 

eventually this should use the new `isTitle` prop but for now, this func just takes a whole bunch of different argument shapes to keep things working without changing too much

core/lib/server/pub.ts
```diff
@@ -1059,14 +1068,20 @@ type MaybePubChildren<Options extends GetPubsWithRelatedValuesAndChildrenOptions
  * Only add the `stage` if the `withStage` option has not been set to `false
  */
 type MaybePubStage<Options extends GetPubsWithRelatedValuesAndChildrenOptions> =
-	Options["withStage"] extends true ? { stage: Stages } : { stage?: never };
+	Options["withStage"] extends true ? { stage?: Stages } : { stage?: never };
```
allisonking, line 1071: because a pub might not belong to a stage

core/app/c/[communitySlug]/pubs/[pubId]/components/jsonSchemaHelpers.tsx
Comment (tefkah) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/jsonSchemaHelpers.tsx:70]: i agree that we need to rework it, but im afraid this might break some of unjournals stuff maybe? I'm not sure im up to date on where we stand wrt to their migration

core/app/c/[communitySlug]/pubs/[pubId]/page.tsx
```diff
@@ -114,15 +106,16 @@ export default async function Page({
 				<div className="flex-1">
 					{pub.values
 						.filter((value) => {
-							return value.field.name !== "Title";
+							return value.fieldName !== "Title";
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:109]: I think we may have an easier way to figure out the title. this is more of a suggestion for the full pr, but if here

https://github.com/pubpub/platform/blob/9ac60fd743e4403008fc7ea5fe29f478a0db6387/core/lib/server/pubtype.ts#L34

you add `_PubFieldToPubType.isTitle`, you should get title information about the pubtype of pubs, which you should be able to use here to filter out the actual title.

core/lib/pubs.ts
tefkah: yeah i was also thinking about how this can be easier, thank you for doing the hard thing for now!

ideally the values from `getPubsWithChildrenetcetc` would also have `isTitle` on them, but i that requires a lot of awkward joins rn

core/lib/server/pub.ts
tefkah: a true! but then i think i should be `Stages | null` right?

core/lib/server/pub.db.test.ts
```diff
@@ -779,9 +779,43 @@ describe("getPubsWithRelatedValuesAndChildren", () => {
...
+		];
+		const userIds = await trx.insertInto("users").values(users).returning(["id"]).execute();
+		await trx
+			.insertInto("pub_memberships")
+			.values(userIds.map(({ id }) => ({ userId: id, pubId, role: MemberRole.admin })))
+			.execute();
+
+		const { getPubsWithRelatedValuesAndChildren } = await import("./pub");
+
+		const pub = await getPubsWithRelatedValuesAndChildren({ pubId }, { withMembers: true });
```
tefkah, line 814: very nice addition

core/lib/server/pub.ts
```diff
@@ -1059,14 +1068,20 @@ type MaybePubChildren<Options extends GetPubsWithRelatedValuesAndChildrenOptions
...
  */
 type MaybePubPubType<Options extends GetPubsWithRelatedValuesAndChildrenOptions> =
 	Options["withPubType"] extends true ? { pubType: PubTypes } : { pubType?: never };
 
+/**
+ * Only add the `pubType` if the `withPubType` option has not been set to `false
+ */
+type MaybePubMembers<Options extends GetPubsWithRelatedValuesAndChildrenOptions> =
+	Options["withMembers"] extends true ? { members: Users[] } : { members?: never[] };
+
```
tefkah, line 1084: just a headsup that all this was moved to `contracts` recently so this might be an annoying merge

tefkah (approved):
works well! other than that note about `renderField` maybe not needing to be changed it's good to go for me, but i would maybe check with @eastofesten whether unjournal is relying on nested pub data being displayed "prettily" on the pub page.

eastofesten:
> works well! other than that note about `renderField` maybe not needing to be changed it's good to go for me, but i would maybe check with @eastofesten whether unjournal is relying on nested pub data being displayed "prettily" on the pub page.

if i follow this correctly, unjournal *is* relying on using nested pub data here, though i think we can get away with it not being pretty.

allisonking:
> > works well! other than that note about `renderField` maybe not needing to be changed it's good to go for me, but i would maybe check with @eastofesten whether unjournal is relying on nested pub data being displayed "prettily" on the pub page.
> 
> if i follow this correctly, unjournal _is_ relying on using nested pub data here, though i think we can get away with it not being pretty.

Ah got it! I'll try to keep the same functionality then!

core/app/c/[communitySlug]/pubs/[pubId]/page.tsx
Comment (allisonking) [reply core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:139]: ah okay that works! although that puts it on `pubType` when ideally it'd be on the `values` so we can still pass it to `getPubTitle`. I'm thinking `getPubTitle` probably can just take a `pubType` with fields (would have to merge for the value though...) but I didn't want to do that here since the func is used in a few places

core/app/c/[communitySlug]/pubs/[pubId]/components/jsonSchemaHelpers.tsx
Comment (allisonking) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/jsonSchemaHelpers.tsx:70]: I didn't want to add a query for the deprecated `schema` field to `getPubsWithRelatedValuesAndChildren`, so @kalilsn checked the data for me ([convo here](https://knowledgefutures.slack.com/archives/CMAQM0BNV/p1732732811633689)) and it seems like we only have to worry about file upload previews. so I [updated this code](https://github.com/pubpub/platform/pull/810/commits/a4ff35dcc357ae60811247b3445d977ffcc6a7ed) to check if the value looks like a file upload, and if so then to render the preview. everything else continues to render as strings

core/app/c/[communitySlug]/pubs/[pubId]/page.tsx
Comment (tefkah) [reply core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:139]: yeah that's fine, i will tackle that in #736

packages/contracts/src/resources/site.ts
```diff
@@ -101,16 +104,33 @@ type MaybePubChildren<Options extends MaybePubOptions> = Options["withChildren"]
...
+> & {
+	isTitle: boolean;
+};
 type MaybePubPubType<Options extends MaybePubOptions> = Options["withPubType"] extends true
-	? { pubType: PubTypes }
+	? {
+			pubType: PubTypes & {
+				fields: PubTypePubField[];
+			};
+		}
```
tefkah, line 124: nice

tefkah (approved):
approved again!

PR #812, opened 2024-11-26
fix: change page size to 10

tefkah:

Very simply just changes the page size for the all pubs page to 10.

## Test Plan
Trust me


gabestein (approved):
TYSM!

Issue #799, opened 2024-11-21
Limit # of Pubs Per Stage On Workflows Page


gabestein:
## Motivation

So the workflows page loads quickly and is usable even when there are many pubs in the community

## Requirements

- Changes the workflow page to only display 3 pubs per workflow stage, ordered by last updated
- Makes stage name into a new link to a new stage page
- Adds an indication of the number of pubs in the stage next to the stage name, e.g. `Submitted (10)`
- Adds link to the stage page below the last displayed pub, with text `See all pubs in stage <Stage Name>`

## Acceptance Criteria

- No fancy stuff, this will be redesigned in the new year

## Extra work

- Always loads a minimum of 20 pubs, even if that means loading more than 3 in a stage (e.g., a community with a 2 stages, each with 20 pubs in them, should load the first 10 in both stages)

PR #813, opened 2024-11-26
Add individual stage page and limit number of pubs per stage

### LINKED ISSUE #800: Create Basic New Stage Page
## Motivation

So users can see all pubs in a stage once we limit the # of pubs displayed in the workflow page.

## Requirements

- Displays title of loaded stage and count of pubs in the stage
- Shows list of pub cards exactly the same as the current workflows page, sorted by last updated

## Acceptance Criteria

- Should re-use as much of the current workflows page as possible to reduce complexity, as this will be redesigned next year
- Url strategy of `/stages/id`, following app pattern
- Does not need to implement add pub button, though that's nice to have

## Extra work

- "Edit Stage Settings" button, only visible to admins, that opens the stage panel in the stage editor (e.g. `https://app.pubpub.org/c/<community>/stages/manage?editingStageId=<stage-id>`)
- Add button "View Stage" to stage page from  stage editor tab that links to new stage page
- Limits the page to 10 pubs and paginates the stage page on the server, exactly the same as the strategy for #778 
- Implements add pub button, with the current stage pre-selected

## Mockups (if available)
Image

kalilsn:
## Issue(s) Resolved
#800 
#799 

This PR adds a new page for viewing pubs in an individual stage, and limits the number of stages shown on the all stages page so that it loads faster. Refactoring the stages components to make this possible led me to make a few modifications to the stages query so that it returns the move constraint destinations and sources with exactly the data they need. I also took the opportunity to remove the integration actions and the token generation from all the components I touched.

## Test Plan
As Jill Admin, visit http://localhost:3000/c/arcadia-research/stages. The page should successfully load much faster than before (though still somewhat slow, since some of the pub_values queries still take 1-2s locally).

Make sure three pubs render per stage, the stage has the number of pubs in it, and that any stage with more than 3 pubs includes a "View all pubs" link to the individual stage page.

Click the Articles stage and verify that it paginates its pubs appropriately. Click the create pub button and verify that the current stage is selected. Create a pub, and make sure it shows up immediately in the stage.

Click Edit Stage Settings button and verify that you are taken to the stage editor with the right stage selected. Click "Visit this stage" and you should be taken back to the individual stage page.

3 pubs per stage on the stages page, "See all pubs" button, pubs count by title:
image
Individual stage page with edit stage button and create pub button:
image
Individual stage page pagination (10 pubs per page):
image
Link to individual stage page on stage editor:
image

### DIFF
--- a/core/app/c/[communitySlug]/stages/[stageId]/page.tsx
+++ b/core/app/c/[communitySlug]/stages/[stageId]/page.tsx
@@ -1,10 +1,10 @@
 import type { Metadata } from "next";
 
-import { Suspense } from "react";
+import { cache, Suspense } from "react";
 import Link from "next/link";
 import { notFound } from "next/navigation";
 
-import type { StagesId } from "db/public";
+import type { CommunitiesId, StagesId } from "db/public";
 import { Capabilities } from "db/src/public/Capabilities";
 import { MembershipType } from "db/src/public/MembershipType";
 import { Button } from "ui/button";
@@ -17,6 +17,10 @@ import { getStages } from "~/lib/server/stages";
 import { PubListSkeleton } from "../../pubs/PubList";
 import { StagePubs } from "../components/StageList";
 
+const getStageCached = cache(async (stageId: StagesId, communityId: CommunitiesId) => {
+	return getStages({ stageId, communityId }).executeTakeFirst();
+});
+
 export async function generateMetadata({
 	params: { stageId, communitySlug },
 }: {
@@ -26,7 +30,7 @@ export async function generateMetadata({
 	if (!community) {
 		notFound();
 	}
-	const stage = await getStages({ communityId: community.id, stageId }).executeTakeFirst();
+	const stage = await getStageCached(stageId, community.id);
 	if (!stage) {
 		notFound();
 	}
@@ -54,7 +58,7 @@ export default async function Page({
 
 	const page = searchParams.page ? parseInt(searchParams.page) : 1;
 
-	const stagePromise = getStages({ communityId: community.id, stageId }).executeTakeFirst();
+	const stagePromise = getStageCached(stageId, community.id);
 	const capabilityPromise = userCan(
 		Capabilities.editCommunity,
 		{ type: MembershipType.community, communityId: community.id },
--- a/core/app/c/[communitySlug]/stages/[stageId]/page.tsx
+++ b/core/app/c/[communitySlug]/stages/[stageId]/page.tsx
@@ -1,8 +1,15 @@
 import { Suspense } from "react";
+import Link from "next/link";
+import { notFound } from "next/navigation";
 
 import type { StagesId } from "db/public";
+import { Capabilities } from "db/src/public/Capabilities";
+import { MembershipType } from "db/src/public/MembershipType";
+import { Button } from "ui/button";
 
+import { CreatePubButton } from "~/app/components/pubs/CreatePubButton";
 import { getPageLoginData } from "~/lib/authentication/loginData";
+import { userCan } from "~/lib/authorization/capabilities";
 import { findCommunityBySlug } from "~/lib/server/community";
 import { getCommunityStages } from "~/lib/server/stages";
 import { PubListSkeleton } from "../../pubs/PubList";
@@ -22,10 +29,39 @@ export default async function Page({
 		findCommunityBySlug(communitySlug),
 	]);
 
+	if (!community) {
+		notFound();
+	}
+
 	const stage = await getCommunityStages({ stageId }).executeTakeFirstOrThrow();
+	const page = searchParams.page ? parseInt(searchParams.page) : 1;
+
+	const showEditButton = await userCan(
+		Capabilities.editCommunity,
+		{ type: MembershipType.community, communityId: community.id },
+		user.id
+	);
 
 	return (
 		<>
+			<div className="mb-16 flex items-center justify-between">
+				<h1 className="text-xl font-bold">{stage.name}</h1>
+				<div className="flex gap-2">
+					{showEditButton && (
+						<Button variant="outline" size="sm" asChild>
+							<Link href={`./manage?editingStageId=${stageId}`}>
+								Edit Stage Settings
+							</Link>
+						</Button>
+					)}
+					<CreatePubButton
+						text="Create Pub"
+						stageId={stageId}
+						communityId={community.id}
+						searchParams={searchParams}
+					/>
+				</div>
+			</div>
 			<Suspense
 				fallback={<PubListSkeleton amount={stage.pubsCount ?? 2} className="gap-16" />}
 			>
--- a/core/app/c/[communitySlug]/stages/[stageId]/page.tsx
+++ b/core/app/c/[communitySlug]/stages/[stageId]/page.tsx
@@ -1,0 +1,43 @@
+import { Suspense } from "react";
+
+import type { StagesId } from "db/public";
+
+import { getPageLoginData } from "~/lib/authentication/loginData";
+import { findCommunityBySlug } from "~/lib/server/community";
+import { getCommunityStages } from "~/lib/server/stages";
+import { PubListSkeleton } from "../../pubs/PubList";
+import { StagePubs } from "../components/StageList";
+
+export default async function Page({
+	params,
+	searchParams,
+}: {
+	searchParams: Record<string, string>;
+
+	params: { communitySlug: string; stageId: StagesId };
+}) {
+	const { communitySlug, stageId } = params;
+	const [{ user }, community] = await Promise.all([
+		getPageLoginData(),
+		findCommunityBySlug(communitySlug),
+	]);
+
+	const stage = await getCommunityStages({ stageId }).executeTakeFirstOrThrow();
+
+	return (
+		<>
+			<Suspense
+				fallback={<PubListSkeleton amount={stage.pubsCount ?? 2} className="gap-16" />}
+			>
+				<StagePubs
+					stage={stage}
+					pageContext={{
+						params,
+						searchParams,
+					}}
+					limit={10}
+				/>
+			</Suspense>
+		</>
+	);
+}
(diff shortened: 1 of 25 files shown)


core/app/components/PubRow.tsx
```diff
@@ -96,24 +92,14 @@ const PubRow: React.FC<Props> = async (props: Props) => {
...
-							{pubStage && (
-								<Suspense>
-									<IntegrationActions
-										pubId={pub.id}
-										token={props.token}
-										stageId={pubStage.id}
-										type="pub"
-									/>
-								</Suspense>
-							)}
```
tefkah, line 108: 🙌

core/app/components/PubRow.tsx
```diff
@@ -96,24 +92,14 @@ const PubRow: React.FC<Props> = async (props: Props) => {
...
 							<div>{props.actions}</div>
 							<PubDropDown pubId={pub.id} searchParams={props.searchParams} />
 						</div>
 					</div>
 				</RowHeader>
 				<RowContent className="flex items-start justify-between">
 					<h3 className="text-md font-medium">
-						<Link href={`pubs/${pub.id}`} className="hover:underline">
```
tefkah, line 116: if you're here already, any change of changing this to `/pubs/${communitySlug}/${pub.id}`? Always found this intermediary step kinda weird since we have no real reason to link to other communities' pubs yet.

core/app/c/[communitySlug]/stages/[stageId]/page.tsx
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	const stage = await getCommunityStages({ stageId }).executeTakeFirstOrThrow();
+	const page = searchParams.page ? parseInt(searchParams.page) : 1;
+
+	const showEditButton = await userCan(
+		Capabilities.editCommunity,
+		{ type: MembershipType.community, communityId: community.id },
+		user.id
+	);
```
Comment (tefkah) [inline core/app/c/[communitySlug]/stages/[stageId]/page.tsx:43]: since these queries don't depend on each other, could we wrap them in a `Promise.all`?

core/app/c/[communitySlug]/stages/[stageId]/page.tsx
```diff
@@ -0,0 +1,80 @@
...
+
```
Comment (tefkah) [inline core/app/c/[communitySlug]/stages/[stageId]/page.tsx:44]: Ideally, could you also add `title` metadata for this page? Together with `Promise.all`ing the two queries, it could look something like this

```suggestion
import type { Metadata } from "next";

import { cache, Suspense } from "react";
import Link from "next/link";
import { notFound } from "next/navigation";

import type { StagesId } from "db/public";
import { Capabilities } from "db/src/public/Capabilities";
import { MembershipType } from "db/src/public/MembershipType";
import { Button } from "ui/button";

import { CreatePubButton } from "~/app/components/pubs/CreatePubButton";
import { getPageLoginData } from "~/lib/authentication/loginData";
import { userCan } from "~/lib/authorization/capabilities";
import { findCommunityBySlug } from "~/lib/server/community";
import { getCommunityStages } from "~/lib/server/stages";
import { PubListSkeleton } from "../../pubs/PubList";
import { StagePubs } from "../components/StageList";

const getCommunityStagesCached = cache(async (stageId: StagesId) => {
	return getCommunityStages({ stageId }).executeTakeFirst();
});

export async function generateMetadata({
	params: { stageId },
}: {
	params: { stageId: string; communitySlug: string };
}): Promise<Metadata> {
	const stage = await getCommunityStagesCached(stageId as StagesId);
	if (!stage) {
		notFound();
	}

	return { title: `${stage.name} Stage` };
}

export default async function Page({
	params,
	searchParams,
}: {
	searchParams: Record<string, string> & { page?: string };

	params: { communitySlug: string; stageId: StagesId };
}) {
	const { communitySlug, stageId } = params;
	const [{ user }, community] = await Promise.all([
		getPageLoginData(),
		findCommunityBySlug(communitySlug),
	]);

	if (!community) {
		notFound();
	}

	const page = searchParams.page ? parseInt(searchParams.page) : 1;

	const stagePromise = getCommunityStagesCached(stageId);
	const showEditButtonPromise = userCan(
		Capabilities.editCommunity,
		{ type: MembershipType.community, communityId: community.id },
		user.id
	);

	const [stage, showEditButton] = await Promise.all([stagePromise, showEditButtonPromise]);

	if (!stage) {
		notFound();
	}

```

Comment (tefkah) [reply core/app/c/[communitySlug]/stages/[stageId]/page.tsx:44]: moving the query outside of the page and wrapping it with `cache` is in order to have the query deduped in the metadata and body of the page.

core/app/c/[communitySlug]/stages/components/StageList.tsx
```diff
@@ -54,96 +51,113 @@ export async function StageList(props: Props) {
...
+				/>
+			)}
+			{!pagination && totalPubLimit && stagePubs.length > totalPubLimit && (
+				<Button
+					variant="ghost"
+					className="text-md inline-flex text-muted-foreground"
+					size="lg"
+					asChild
+				>
+					<Link href={`stages/${stage.id}`}>See all pubs in stage {stage.name}</Link>
```
Comment (tefkah) [inline core/app/c/[communitySlug]/stages/components/StageList.tsx:158]: I don't get why this link works? it should be `/c/${communitySlug}/stages/${stage.id}`, or, as i've learned, `./${stage.id}`, but somehow next just... figures it out?

I don't get it, the docs also don't mention anything about this behavior. For my sanity, could you change it to be `/c/${communitySlug}/stages/${stage.id}`?

core/app/c/[communitySlug]/stages/manage/components/panel/StagePanelOverviewManagement.tsx
```diff
@@ -13,9 +17,14 @@ export const StagePanelOverviewManagement = (props: Props) => {
 	return (
 		<>
 			<h4 className="mb-2 font-semibold">Stage Management</h4>
-			<Button variant="secondary" onClick={onDeleteClick}>
-				Delete this Stage
-			</Button>
+			<div className="flex gap-2">
+				<Button variant="secondary" asChild>
+					<Link href={`./${props.stageId}`}>Visit this Stage</Link>
```
Comment (tefkah) [inline core/app/c/[communitySlug]/stages/manage/components/panel/StagePanelOverviewManagement.tsx:22]: similar to an above comment, i'm shocked that this works, but it does not seem to be documented behavior for next. could we instead change this to `/c/${communitySlug}/stages/${props.stageId}`?

core/lib/server/stages.ts
```diff
@@ -48,33 +49,27 @@ export const getPubIdsInStage = (stageId: StagesId) =>
 			.where("stageId", "=", stageId)
 	);
 
-export const getCommunityStages = (communityId: CommunitiesId) =>
+type CommunityStageProps = XOR<{ communityId: CommunitiesId }, { stageId: StagesId }>;
+export const getCommunityStages = ({ communityId, stageId }: CommunityStageProps) =>
```
tefkah, line 53: i like this change, but could we maybe change the name a bit? Maybe like `getStages`?

Also, maybe require passing the `communityId` anyway, that way it's less likely for us to accidentally allow fetching other communities' stage info by doing something like `/c/${your_community_slug}/stages/${someone_elses_stage}`

I did the same for the `getPubsWith...` function for the same reason

tefkah (approved):
Looks great! Other than adding the metadata and changing some `href`s it's ready to merge.

core/app/c/[communitySlug]/stages/components/StageList.tsx
Comment (kalilsn) [reply core/app/c/[communitySlug]/stages/components/StageList.tsx:158]: Yeah the way I did it here is just straight up wrong, no idea why it works. I was experiencing some very strange behavior with these links which is probably how I ended up getting this one wrong. I'll go ahead and make all these links relative to the root like you suggested!

core/lib/server/stages.ts
kalilsn: two great ideas, ty!

core/app/c/[communitySlug]/stages/[stageId]/page.tsx
Comment (kalilsn) [reply core/app/c/[communitySlug]/stages/[stageId]/page.tsx:44]: i'm not sure i follow the cache change. does autocache not work in this case for some reason?

Comment (kalilsn) [reply core/app/c/[communitySlug]/stages/[stageId]/page.tsx]: I've made the rest of the changes but didn't want to do this one without first understanding why!

Comment (tefkah) [reply core/app/c/[communitySlug]/stages/[stageId]/page.tsx]: tldr: this optimization is not per se necessary, but i think it's probably better to do it than not.

ah sorry! this specifically adds the use of `React.cache`. this dedupes two function calls made in the same render pass, making that function only being called once during the render of the page (metadata + the body of the function). this is in addition to the `autoCache` behavior, which just stores the result of the query but also does some other stuff (calculating cache keys, reading from the cache) which does not get cached. using `React.cache` is  slightly more efficient (only slightly tho), because all the other stuff in `autoCache` will not happen.

basically, without adding react cache, this happens:
- in metadata: query gets compiled, cache keys calculated, query gets executed, result gets stored in filesystem cache 
- in body: query gets compiled, cache keys calculated, query gets reads from cache.

with react cache
- in metadata: query gets compiled, cache keys calculated, query gets executed, result gets stored in filesystem + in memory
- in body: result is immediately read from memory.

More info (not necessary):

i considered building this behavior into `autoCache`, but that would require changing the api somewhat. i discussed this with @3mcd at some point and he wasn't really a fan, so we decided to just do this kind of deduping manually when needed/desired.

for reference, the api for `autoCache` would need to look like this

```ts
const getPub = autoCache((pubId: PubsId) => db.selectFrom('pubs') )
```

rather than

```ts
const getPub =(pubId: PubsId) => autoCache(db.selectFrom('pubs'))
```


i.e. you would need to wrap the entire function inside of `autoCache`, rather than just the selection part. This is bc [calling different memoized fns will read from different caches](https://react.dev/reference/react/cache#pitfall-different-memoized-functions). Since currently calling eg `getCommunityStages(stageId)` will return a new object with new functions (`.execute`, `.executeTakeFirst`, `.executeTakeFirstOrThrow`), deduping is not possible.
 Plus, you would not be able to pass objects as arguments, as [react needs stable input references for the deduplication to work](https://react.dev/reference/react/cache#memoized-function-still-runs).

Maybe this is still somewhat unclear, in which case feel free to leave it out and merge it, and we can discuss this in more detail in some point if you're interested!

PR #814, opened 2024-11-27
feat: add google drive action scaffolding

### LINKED ISSUE #780: Scaffolding for Google Drive Action
## Motivation

So users can insert content into PubPub from Google Drive

## Requirements

- Users can add the gdrive action to their workflows from the action tab of a stage
- Allows users to specify a pub field which contains a Google Drive URL
- Allows users to specify a field which will serve as the html content field for the pub, where generated HTML will be stored
- When run, given google drive URL in field, runs code that @isTravis will iterate on over time that will eventually accomplish the below.
<details>
  <summary>Eventual Gdrive Functionality</summary>
  
  - Generate HTML from Google Document in folder and update specified HTML content Pub field
  - Create related pub field for citations if needed, create or update (if already extant) related pubs for citations and adds related fields with string values for any text that should be added to the citation (e.g. page numbers)
  - Create related pub field for tags if needed, create or update tag pubs and adds related fields with null value
  - Create or updates child pubs for images, including uploading images to platform assets bucket in folder named same as pub id
  - In action settings, users can specify shortcodes for creating related pubs and transforming some text into HTML
</details>

## Acceptance Criteria

- Uses a globally stored Google account key (see: #779) to query drive API (could use a local env var for testing purposes/so as not to be blocked)

tefkah:
- **feat: add google drive import action to db**
- **feat: add scaffolding for gdrive action**
- **chore: move**
- **chore: remove simple icons from ui**
- **feat: create new outputfield component, let gdrive import use that**
- **feat: make the gdrive action work**
- **feat: make the gdrive action superadmin only**
- **fix: remove unused import**

## Issue(s) Resolved
Resolves #780

Sets up basic scaffolding for the GoogleDriveImport action.

- Creates a new googleDriveImport actionname in the db
- Does all the usual action stuff
- Creates a new OutputField custom config field, which is used to specify to which pubfield the output should be written. 
  - This is just different enough from the OutputMap component used in the `http` action that i just wrote a new one that uses a small subset of what that component uses
  - In the future we should create some more generic way to do this.

The rest main bulk of the action's `run` function will be writtten by @isTravis, this just sets up everything around it.

Currently the action just reads the `docUrl` field and writes it to the specified output field. In the future, after reading the `docUrl` field, it will do some more processing, but the rest will be the same.

## Test Plan
1. `pnpm dev:setup`
2. Go to arcadia-research
3. Add a new HTTP action to the "articles" stage
4. In the config
  a. Specify some URL as the docurl
  b. Specify `arcadia-research:doi` as the pubfield which overrides that value
  c. Specify `arcadia-research:title` as the pubfield to write to
5. Run action without setting any parameters 
6. You should see that the title of the Pub has changed to the doi.
7. Run the action again, but remove the pubfield specification
8. See that the title changes to the url you specified

Other tests:
1. Check that you can only specify `String` fields as the output.


3mcd (approved):
Nice work!

PR #816, opened 2024-11-27
Related pubs table

allisonking:
## Issue(s) Resolved

Closes https://github.com/pubpub/platform/issues/777

Related pubs can now be seen on pub detail pages.

## Test Plan
Visit the journal article called "Identification of capsid-like proteins in venomous and parasitic animals" in the Arcadia seed community. You should be able to see its related pubs, grouped by relationship type.

image

## Notes

No "actions" or "member fields" rendering yet since those would require more queries ([convo here](https://knowledgefutures.slack.com/archives/CMAQM0BNV/p1732722697668169?thread_ts=1732649416.238439&cid=CMAQM0BNV))

### DIFF
--- a/core/app/c/[communitySlug]/pubs/[pubId]/components/RelatedPubsTable.tsx
+++ b/core/app/c/[communitySlug]/pubs/[pubId]/components/RelatedPubsTable.tsx
@@ -13,7 +13,7 @@ import { Button } from "ui/button";
 import { Checkbox } from "ui/checkbox";
 import { DataTableColumnHeader } from "ui/data-table";
 
-import { DataTable } from "~/app/components/DataTable/v2/DataTable";
+import { DataTable } from "~/app/components/DataTable/DataTable";
 import { getPubTitle } from "~/lib/pubs";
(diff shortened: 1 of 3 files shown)


core/app/c/[communitySlug]/pubs/[pubId]/components/RelatedPubsTable.tsx
```diff
@@ -0,0 +1,154 @@
...
+							{field.name}
+							<span className="ml-2 font-mono text-xs">{field.count}</span>
+						</Button>
+					);
+				})}
+			</div>
+			<Table pubs={groupedBySlug[selected].map((value) => value.pub)} />
+		</div>
+	);
+};
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/[pubId]/components/RelatedPubsTable.tsx:154]: even though i theoretically like the `searchParameter` based approach used by the children table more, i have to admit that it's nice how quick navigation is in comparison haha.
I am slightly worried about pages with a ton of related pubs, but we'll just have to solve that with specific pagination for the related pubs i think, so i think this is approach is good!

3mcd (approved):

Issue #786, opened 2024-11-14
Create New Pub Editing Page Using Default Form


gabestein:
## Motivation
So users can create and edit pubs internally using the new customizable default form.
So users have more predictable ways/places to edit pubs

## Requirements
- Ability to edit existing Pub using default form
- Ability to create new Pub using default form

### Edit and Creation
- Should show side nav, like any other page (unlike invited form `/fill` and update pages)
- Should not auto-save and should not render auto-save text in header when form is rendered on internal create or edit page
- Should warn users who navigate away without saving changes.
- For both editing and creation, form should allow user to edit the stage the pub is in, the same as the existing edit modal.
- Instead of rendering any configured submission buttons on this form, when this is an edit form, “Save” button should be placed in sticky header, aligned left with title, with the text “Form will save when you click save” next to it. When clicked, text should change to “Last saved: {time}.” 
- Should use the same permissions as the current edit and create pub modal

### Edit
- If editing a pub, should use something like `/pub/{id}/edit`, and existing values should auto-fill in the page, similar to how invited forms for existing pubs currently work
- "View Pub" button takes the place of the current "copy link" button in the top right sticky header

### Creation
- If creating a new pub, should use something like `/pub/create?type=<pubTypeID>&stage=<stageId>`. Otherwise, this can effectively be the equivalent of a current `/fill` page, with the new save button.
- No "copy link" button appears in the top right sticky header
- When a user click the save button to create the new pub, they are navigated to the edit page for the newly created pub 

## Design Doc
https://docs.google.com/document/d/1cWdWG1hB0b4U3ikb30U_Nd_LjvX5bANFTU2GhRqNW3s/edit?tab=t.0#heading=h.ugg1cxkdex80

kalilsn:
Based on #787 I'm assuming this includes a pub creation page too, right? Should that go at `pubs/create`? And if so, it seems like there's probably another requirement to make this page use either a route or query param to set both the `type` and `stage` since they're not going to be selectable within the form anymore.

gabestein:
Yes. I had meant to split those into two tickets, but I think someone told me it wasn't worth it. I'll add that to the ACs.

gabestein:
I'm assuming this is 3-day for now, but it could be more and thus we may want to split out the creation page. Please let me know when you can.

allisonking:
Let's keep it at 3-day!

allisonking:
[...]

PR #820, opened 2024-12-03
Pub edit default form

allisonking:
Also see https://github.com/pubpub/platform/pull/821 for more in depth comments!

## Issue(s) Resolved

Part of https://github.com/pubpub/platform/issues/786

This mostly merges code paths between Pub editing and Pub editing via a form. 

| Before | After |
|--------|--------|
| image | image  |

There's still some repeated logic between `PubEditor` and the form fill page. So I think even more merging could be done, though I'm not quite sure the best way to go about it yet. If I get inspired in the next few days, I might try it, but also I don't know if I want to do it while @tefkah is changing all the `getPub` queries since that could get messy!

## Test Plan

Currently, because we don't have default forms in the db yet, you'll have to add a default form manually. I did this via Prisma, and set the seeded "Review" form in croccroc to `isDefault=true` 

* The new pub edit page at `/c/[communitySlug]/pubs/[pubid]/edit` should work
* The old form fill page should still work

Creating a pub is currently not working because there's no button in the modal anymore...

image

## Notes
The save button isn't as nice as it used to be because it is now outside of the form, which means we can't really query the form state to disable/make it loading. We have this same problem on the form builder page...


core/app/components/StageSelect/StageSelectServer.tsx
```diff
@@ -8,28 +8,12 @@ type Props = {
 	communityId: CommunitiesId;
 	fieldLabel: string;
 	fieldName: string;
-	value?: StagesId;
 };
 
-export async function StageSelectServer({ communityId, fieldLabel, fieldName, value }: Props) {
```
allisonking, line 14: I don't think passing the stage's value is actually needed here—the form takes care of that from what I can tell

core/app/components/StageSelect/StageSelectClient.tsx
```diff
@@ -24,16 +23,16 @@ export function StageSelectClient(props: Props) {
 					</div>
 					<Select
 						{...field}
-						value={props.stage?.id}
 						onValueChange={(value) => {
 							field.onChange(value);
 						}}
+						defaultValue={field.value}
```
allisonking, line 29: I believe we should use the value from the form rather than passing a value in

core/app/components/StageSelect/StageSelectClient.tsx
```diff
@@ -24,16 +23,16 @@ export function StageSelectClient(props: Props) {
...
-						<SelectTrigger>
-							<SelectValue placeholder="Select a stage" defaultValue={field.value}>
-								{props.stages.find((stage) => stage.id === field.value)?.name}
-							</SelectValue>
-						</SelectTrigger>
+						<FormControl>
+							<SelectTrigger>
+								<SelectValue placeholder="Select a stage" />
+							</SelectTrigger>
+						</FormControl>
```
allisonking, line 35: this seems to work better, otherwise sometimes the selector is blank after making a new selection

core/app/components/pubs/PubEditor/PubEditorClient.tsx
```diff
@@ -1,348 +1,348 @@
 "use client";
```
allisonking, line 1: this isn't a very useful diff—it would be more useful to diff this against `ExternalFormWrapper`! see diagram in PR description for more info

core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx
```diff
@@ -1,42 +1,16 @@
 "use client";
```
Comment (allisonking) [inline core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx:1]: This isn't a useful diff— `ExternalFormWrapper` has totally changed. See diagram for info.

This and `PubEditorWrapper` both serve as client-side wrappers around `PubEditorClient`. They need a wrapper so that we can pass a function `onSuccess` as a prop, since both paths want to handle success states differently.

core/app/components/StageSelect/StageSelectClient.tsx
tefkah: makes sense to me!

core/lib/server/form.ts
```diff
@@ -26,10 +26,15 @@ import { _getPubFields } from "./pubFields";
...
- * Get a form by either slug or id
+ * Get a form by either slug, id, or pubtype ID. If given a pubtype ID, retrieves the
+ * default form for that pubtype.
  */
 export const getForm = (
-	props: XOR<{ slug: string }, { id: FormsId }> & { communityId: CommunitiesId },
+	props: (
+		| { id: FormsId; slug?: never; pubTypeId?: never }
+		| { id?: never; slug: string; pubTypeId?: never }
+		| { id?: never; slug?: never; pubTypeId: PubTypesId }
```
3mcd, line 36: Do you know the difference between this approach the following?

```ts
| { id: FormsId }
| { slug: string }
| { pubTypeId: PubTypesId }
```

Is it better intellisense? The former seems much more concise.

3mcd (approved):
Looks great! Took me a minute to figure out why the new default form wasn't showing up, but it ended up being a Next cache issue (obvious in retrospect). Nice work!

core/lib/server/form.ts
allisonking: yup better intellisense, otherwise if you do a `props.slug` you get errors like

```
Property 'slug' does not exist on type '({ id: FormsId; } | { slug: string; } | { pubTypeId: PubTypesId; }) & { communityId: CommunitiesId; }'.
  Property 'slug' does not exist on type '{ id: FormsId; } & { communityId: CommunitiesId; }'
```

core/app/c/[communitySlug]/pubs/[pubId]/edit/page.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+import { ContentLayout } from "~/app/c/[communitySlug]/ContentLayout";
+import { PageTitleWithStatus } from "~/app/components/pubs/PubEditor/PageTitleWithStatus";
+import { PubEditor } from "~/app/components/pubs/PubEditor/PubEditor";
+import { getPageLoginData } from "~/lib/authentication/loginData";
+import { userCan } from "~/lib/authorization/capabilities";
+import { getCommunityBySlug } from "~/lib/db/queries";
+import { getPubsWithRelatedValuesAndChildren } from "~/lib/server";
+
+export async function generateMetadata({
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/[pubId]/edit/page.tsx:19]: I mentioned something similar in a PR from kalil, but could you add this `React.cache`d version of the function like this, and use it in both the body and the metadata?

This way, the result of the entire function call is stored in memory and deduped during the rendering of the page.

Eg we won't have to read from the filesystem cache and compute the cache keys twice. This is somewhat of a limitation to `autoCache`s current api. See here for a more thorough explanation: https://github.com/pubpub/platform/pull/813#discussion_r1869676750 

Not a blocker, you can ignore it for now if you just want to get it in, but I thought i'd mention it so we're all aware of this!

```suggestion
const getPubsWithRelatedValuesAndChildrenCached = cache(
	async (pubId: PubsId, communityId: CommunitiesId) => {
		return getPubsWithRelatedValuesAndChildren({
			pubId,
			communityId,
		});
	}
);

```

core/app/c/[communitySlug]/pubs/[pubId]/edit/page.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+	if (!pubId || !communitySlug) {
+		return null;
+	}
+
+	if (!canUpdatePub) {
+		redirect(`/c/${communitySlug}/unauthorized`);
+	}
+
+	const community = await getCommunityBySlug(communitySlug);
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/[pubId]/edit/page.tsx:76]: We should really get rid of this and rename the correct one, but could you use `findCommunityBySlug` instead? `getCommunityBySlug` still uses prisma and isn't cached!

core/app/c/[communitySlug]/pubs/create/page.tsx
```diff
@@ -0,0 +1,95 @@
...
+	searchParams,
+}: {
+	params: { communitySlug: string };
+	searchParams: Record<string, string>;
+}) {
+	const { communitySlug } = params;
+
+	const { user } = await getPageLoginData();
+
+	const community = await getCommunityBySlug(communitySlug);
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/create/page.tsx:41]: Similar to on the edit page, could you use `findCommunityBySlug` instead?

core/app/c/[communitySlug]/pubs/[pubId]/edit/page.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { userCan } from "~/lib/authorization/capabilities";
+import { getCommunityBySlug } from "~/lib/db/queries";
+import { getPubsWithRelatedValuesAndChildren } from "~/lib/server";
+
+export async function generateMetadata({
+	params: { pubId, communitySlug },
+}: {
+	params: { pubId: string; communitySlug: string };
+}): Promise<Metadata> {
+	const community = await getCommunityBySlug(communitySlug);
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/[pubId]/edit/page.tsx:24]:
```suggestion
	const community = await findCommunityBySlug(communitySlug);
```

core/app/c/[communitySlug]/pubs/create/page.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { getPageLoginData } from "~/lib/authentication/loginData";
+import { userCan } from "~/lib/authorization/capabilities";
+import { getCommunityBySlug } from "~/lib/db/queries";
+
+export async function generateMetadata({
+	params: { communitySlug },
+}: {
+	params: { pubId: string; communitySlug: string };
+}): Promise<Metadata> {
+	const community = await getCommunityBySlug(communitySlug);
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/create/page.tsx:22]:
```suggestion
	const community = await findCommunityBySlug(communitySlug);
```

tefkah (approved):
Very nice work! This was quite a complex task, but you did it so quickly and so well, really great work!

None of the changes I'm requesting are real blockers, except for changing the uses of `getCommunityBySlug`. I'll approve the PR regardless so you could get it in today if you want, but that one would be nice to get in (and we should probably delete `getCommunityBySlug` to prevent these issues in the future!)


Also, in case you're interested, this PR provides a nice little speedboost on the `/pubs` and `/stages` page! (left old, right new)


https://github.com/user-attachments/assets/f4cf3666-3fca-4eb2-b8e3-55392934c4d1


https://github.com/user-attachments/assets/d73aeb09-cd39-4a05-97e8-6d86015bcfc9

With this, i will close #677 !

allisonking:
> Also, in case you're interested, this PR provides a nice little speedboost on the `/pubs` and `/stages` page! (left old, right new)

oh amazing, thanks for recording those!! ⚡ I think I've addressed all your comments, thanks for the thorough review!

Issue #787, opened 2024-11-14
Update Pub Creation and Editing Links To Use New Edit/Create Pages


gabestein:
## Motivation
So users can create Pubs using the new create Page.
So users can more consistently access the edit page of a Pub.

## Requirements
- When user clicks add/create button (stages, workflows, workflow editor card on “Pubs tab”), they see a dropdown of the community's pub types to choose from.
- Clicking the pub type in the dropdown navigates to the fill page, which is pre-filled out with the selected type
- In the pub card hamburger menu clicking "update" loads new Pub editing page from #786.
   - Workflows
   - All pubs
   - Workflow stage settings "Pubs" panel)
- Adds "Update" button to individual pub page at the top of the sidebar, which links to new Pub editing page

## Extra Work
- Update `/fill` page to show selected stage and pub type.
- Ability to change pub type on `/fill` page

## Design Doc
https://docs.google.com/document/d/1cWdWG1hB0b4U3ikb30U_Nd_LjvX5bANFTU2GhRqNW3s/edit?tab=t.0#heading=h.ugg1cxkdex80

tefkah:
This is in conflict with #677. I'm down for that if we really want separate update/creation pages, but I don't think we should do it for performance reasons. See the video there for some perf comparisons/comments. Sorry for not mentioning this in the doc, I thought i did but i think that was in some slack message!

We could also have the best of both worlds by using [intercepting routes](https://nextjs.org/docs/app/building-your-application/routing/intercepting-routes). Basically, on selected pages, we would render a modal, but directly going to the url would take you to the dedicated page. I do think there is value in having modals in some places. In legacy for instance, i always found it kind of annoying how many clicks it takes to get to be able to edit a pub. This also has better performance than my approach in that PR, so i don't mind just closing that PR! For clarity: we can do the approach outlined in this issue and then add modals back with intercepted routes later, so we can just go with this.

tefkah:
By `/fill` page you mean the `pub create` page which renders the default form here, right?
And the extra work of changing the pub type would be to facilitate creating a different kind of pub than you intended, correct? Eg I say I want to create a new "Submission", but once I'm on the page I'm like "Oh no, it should be an Evaluation instead". 

I'm clarifying bc I initially read it as "you should be able to change the pubtype of an existing pub using forms", but i don't think that's what's meant!

tefkah:
I'd call this 2 days, even if it's quite simple, bc some e2e tests will need to be updated.

[...]

PR #821, opened 2024-12-04
Pub create default form

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/786
Closes https://github.com/pubpub/platform/issues/787

* Renders rich text properly
* Includes pub values on a pub but not on a form
* Replaces pub update modal with link to pub edit page
* Create pub flow

## Test Plan
Same as https://github.com/pubpub/platform/pull/820, need to manually add default forms to pub types in the community you are testing.

### Pub create
* Create a pub from the All Pubs page. You should be able to select a pub type, then be brought to the pub create form (which looks like the default form you specified)
* Create a pub from a stage. The stage selector should be pre-populated
* Create a pub from a pub page, as a child. The pub should be created as a child.

### Other
* Rich text fields in a form should render properly
* Create a pub with all values filled in. Then remove one of the fields from the default form. That field should still show up in the form, just at the bottom, after the default form fields. 

image

image

Redirects to Edit page after Save

image

Pub page has an "Update" button now which brings you to the Edit page

image


core/app/components/pubs/PubEditor/helpers.ts
```diff
@@ -0,0 +1,27 @@
+import type { FormElementsId } from "db/public";
```
allisonking, line 1: I removed this in https://github.com/pubpub/platform/pull/820 but turns out I still needed it, for pub values that don't have a corresponding form! no changes were made here

core/app/c/[communitySlug]/pubs/create/page.tsx
```diff
@@ -0,0 +1,95 @@
...
+	const community = await getCommunityBySlug(communitySlug);
+
+	if (community === null) {
+		notFound();
+	}
+
+	const communityId = community.id as CommunitiesId;
+
+	const canCreatePub = await userCan(
+		Capabilities.createPub,
```
Comment (allisonking) [inline core/app/c/[communitySlug]/pubs/create/page.tsx:50]: this was originally on `MembershipType.pub` but I moved it to `MembershipType.community` which I think makes more sense, but let me know if that's not right!

core/app/c/[communitySlug]/pubs/create/page.tsx
```diff
@@ -0,0 +1,95 @@
...
+		{
+			type: MembershipType.community,
+			communityId,
+		},
+		user.id
+	);
+
+	if (!canCreatePub) {
+		return null;
```
Comment (allisonking) [inline core/app/c/[communitySlug]/pubs/create/page.tsx:59]: not sure if we have standardized behavior for what things should look like if `!userCan`?

core/app/components/pubs/CreatePubButton.tsx
```diff
@@ -1,20 +1,31 @@
-import type { CommunitiesId, StagesId } from "db/public";
```
allisonking, line 1: Before, the `CreatePubButton` required `searchParams` to be passed in, which would be passed on to the PubEditor.

Now that the PubEditor is its own page, it can access its own `searchParams` easily, so this button just opens a modal for the user to pick a pubtype, then `PubTypeFormClient.tsx` shepherds the search parameters on to the PubEditor page.

core/lib/server/pub.db.test.ts
```diff
@@ -846,6 +846,36 @@ describe("getPubsWithRelatedValuesAndChildren", () => {
 			members: users.map((u) => ({ ...u, role: MemberRole.admin })),
 		});
 	});
+
+	it("should fetch a pub that has no pub values", async () => {
```
allisonking, line 850: a bug came up that our e2e tests kind of caught after this restructuring—creating a pub with no values was not "gettable" by `getPubsWithRelatedValuesAndChildren`. this adds a test to make sure we can still get it

core/app/c/[communitySlug]/pubs/create/page.tsx
Comment (kalilsn) [reply core/app/c/[communitySlug]/pubs/create/page.tsx:50]: That's right, definitely a mistake! I think it's both a community and stage capability.

Comment (kalilsn) [reply core/app/c/[communitySlug]/pubs/create/page.tsx:50]: I guess the stage part of that is a bit of an issue if the stage doesn't get selected until they're on the create page...

core/app/c/[communitySlug]/pubs/create/page.tsx
Comment (kalilsn) [reply core/app/c/[communitySlug]/pubs/create/page.tsx:59]: I think calling `notFound()` or [explicitly redirecting them to our unauthorized page](https://github.com/pubpub/platform/blob/main/core/app/c/%5BcommunitySlug%5D/fields/page.tsx#L57) is what we've been doing

When we upgrade to next 15 I would like to use https://nextjs.org/docs/app/api-reference/functions/forbidden or https://nextjs.org/docs/app/api-reference/functions/unauthorized

core/app/c/[communitySlug]/pubs/create/page.tsx
Comment (allisonking) [reply core/app/c/[communitySlug]/pubs/create/page.tsx:50]: hmm maybe we filter the stage dropdown by only the ones the user has capabilities for?

PR #822, opened 2024-12-05
feat: add title to pubs table on pub_value update

### LINKED ISSUE #736: Update pages to use designated pub name field as title

tefkah:
- **feat: add title field to pubs table and update updatedAt trigger**
- **feat: make the title updates work**

## Issue(s) Resolved
#736 

This PR adds a new `title` column to the `pubs` table. This column is not to be updated manually, but to be updated through newly added triggers on inserts/updates/deletes to the `pub_value` table. A migration is added to backfill these title values to existing pubs.

Additionally, it modifies the `getPubTitle` function and places where it's called to use this new field.
The fallback for searching for fields with `title` in their slug is kept in place.

### Possible improvements

I could make it so that instead of having `title: null` for no title, I could set it to be `Untitled ${PubType} - ${createdAt}` instead. This would make searching easier, but would make it harder to distinguish pubs without a title from those with one in code (ie instead of doing `title === null`, you'd have to do `!title.match(/Untitled/)`, which is much more fragile).

## Test Plan
1. Run tests
2. Create a new field, eg `arcadia-research:name`
3. Create a new pubtype, eg `Person`, with `arcadia-research:name` as its title
4. Create a new `Person` pub, filling out the `name` field.
5. See that on `/stages`, `/stages/manage`, `/pubs`, and `/pubs/<pubId>` page the title is set to be that new value.
6. Create another `Person` pub, this time without filling in the title.
7. See that the title is shown as `Untitled Person - ${createdAt}`, rather than `Untitled Pub`.

## Notes

The types for `PubTitle` props etc are a bit ugly. I address this in #823


core/lib/server/cache/specialTables.ts
tefkah, line 1: this is to make sure that when run an `autoRevalidate`d query that only uses `pub_values` (like the update function in `updatePub`), we are sure that any `autoCache`d queries that only use the `pubs` table are also invalidated.

example:

say we have something that's like

```ts
const PubTitle = async (props: { pubId: PubsId } ) => {
	const pub = autoCache(db.selectFrom("pubs").select("title").where("pubId", "=", 		props.pubId)).executeTakeFirstOrThrow();
	
	return <span>{pub.title}</span>
}
```

If we did not add this linking, when we go to update the value that is the title for a pub, say `croccroc:title`, the above function will not get revalidated, and we would see the old value.

Currently this is not a problem, because if we have an autocached query to get pubs we almost always also get the values, thus it's also tagged with `pub_values`, thus it would get revalidated if we update the pubvalues.

but this could prove a problem in the future if we want to have some more efficient queries that only fetch titles and no values.

core/lib/server/pub.ts
tefkah, line 1: i did not yet update `getPubsWithRelatedValuesAndChildren` to have an option to only fetch the `title`. I will add this in #823 as a follow up, where i will also make this query slightly more efficient (hopefully)

core/lib/server/pub.ts
```diff
@@ -513,6 +514,9 @@ export const createPubRecursiveNew = async <Body extends CreatePubRequestBodyWit
 						.returningAll()
 				).execute()
 			: [];
+
+		const pub = await getPlainPub(newPub.id, trx).executeTakeFirstOrThrow();
```
tefkah, line 518: so i added this bc i think it's nice to, in api responses etc, return an accurate `title` for the recently created pub.

we need to do it somewhat like this, bc we cannot accurately reconstruct what the title would be from the input alone (unless we fetched the pubtype X pubfields for the pub, which I tihnk is slower than just fetching the pub). this does make this function slightly slower, but i don't think that's that big of a deal.

core/playwright/fixtures/pubs-page.ts
```diff
@@ -71,23 +71,25 @@ export class PubsPage {
...
+		if (!path) {
+			throw new Error("Unable to get path from newly created pub");
+		}
+
+		const pubId = path.match(/\/([0-9a-f-]+)$/)?.[1];
+
 		if (!pubId) {
 			throw new Error("Unable to get pub id from newly created pub");
 		}
+
```
tefkah, line 92: this needed to be changed, bc it's not possible to accurately reconstruct the title from just the pub creation data anymore. this test now just picks the first pub from the `pubs` page, which should be the most recently created one.

core/prisma/migrations/20241205152006_add_title_field/migration.sql
```diff
@@ -0,0 +1,75 @@
...
+
+    WITH tf AS (
+        SELECT DISTINCT ON (inserted_deleted_rows."pubId") inserted_deleted_rows."pubId", inserted_deleted_rows."value" FROM inserted_deleted_rows 
+            JOIN "pubs" p ON inserted_deleted_rows."pubId" = p."id"
+            JOIN "_PubFieldToPubType" pft ON pft."A" = inserted_deleted_rows."fieldId" AND pft."B" = p."pubTypeId" AND pft."isTitle" = true
+    ) 
+    INSERT INTO tmp_affected_pubs ("pubId", "value")
+    -- the #>> '{}' transforms the json into text without the quotes
+    SELECT DISTINCT "pubId", ("value" #>> '{}') FROM tf
+    ON CONFLICT DO NOTHING;
```
tefkah, line 25: this fills the just created temporary table with `pubId, title`

core/prisma/migrations/20241205152006_add_title_field/migration.sql
```diff
@@ -0,0 +1,75 @@
...
+        "updatedAt" = CURRENT_TIMESTAMP,
+        "title" = CASE 
+            WHEN TG_OP = 'DELETE' THEN NULL 
+            ELSE COALESCE(tmp."value", "pubs"."title")
+        END
+    FROM (SELECT DISTINCT "pubId" FROM inserted_deleted_rows) AS affected
+    LEFT JOIN tmp_affected_pubs tmp ON tmp."pubId" = affected."pubId"
+    WHERE "pubs"."id" = affected."pubId";
+
+    RETURN NULL;
```
tefkah, line 38: roughly:

for each `pubId` in the `inserted_deleted_rows`, set the `title` to the constructed `title` or the old one, and set the `updateAt` to the current timestamp.

note: this currently does not reflect changes from `value: "Some title"` -> `value: null`, as we don't really do that afaik. this is mostly bc this way we can use one update statement rather than 3ish.

is this a problem? should not be too hard to change it, i'll then probably just separate out updating the `updatedAt` and updating the `title`, add a separate query for setting the `DELETE` value, and then write a third query where change the `LEFT JOIN` to an inner join.
there might be a better way of doing this, just let me know whether changing a value to `null` should be accounted for!

core/prisma/migrations/20241205152006_add_title_field/migration.sql
```diff
@@ -0,0 +1,75 @@
...
+-- backfill the title for all pubs
+WITH values_titles as (
+    SELECT "pubId", ("value" #>> '{}') as "title" FROM "pub_values"
+    JOIN "pubs" ON "pub_values"."pubId" = "pubs"."id"
+    JOIN "_PubFieldToPubType" pft ON pft."A" = "pub_values"."fieldId" AND pft."B" = "pubs"."pubTypeId" AND pft."isTitle" = true
+)
+UPDATE "pubs"
+SET "title" = values_titles."title"
+FROM values_titles
+WHERE "pubs"."id" = values_titles."pubId";
```
tefkah, line 75: i would test this backfill by commenting out everything here except for the adding of the title column, running migrations + reset, seeing that the titles are bad, then running the backfill separately using `psql`.

allisonking (commented):
nice job!! I noticed one case which I'm not sure if it is required, but let's say you have a Pub of type `Submission` with a field `Title` which is set as its title field.

1. Go to Fields and add a new field, i.e. `Display name`
2. In the pub type editor, add `Display name` to `Submission` and set it as `isTitle`
3. Visit the pub page. It will still have its old title (this is fine since there's no value for `Display name` so could be seen as a fallback value)
4. Add a value for `Display name` via the pub editor
5. This value will now show up as the title for the pub (good!)
6. Go back to the pub type editor and change the `isTitle` back to `Title`
7. Go back to the pub page—the title will still be the value of `Display name`, not `Title` (seems undesirable)

core/prisma/migrations/20241205152006_add_title_field/migration.sql
kalilsn: I think that's okay, but I'm not 100% sure. I think the only time we're using `value: null` is for relationship field values but since we haven't fully implemented those, it's hard to be certain that we'd never let that value change from something to `null`.

kalilsn:
I think the case Allison is describing means we need to trigger this update whenever the pub type's `isTitle` field is changed as well

tefkah:
nice, great catch @allisonking ! that's definitely an important case we need to check for, i'll add a trigger for that as well.

tefkah:
i have fixed the issue with the title not updating if the pubtype is changed @allisonking @kalilsn

core/lib/server/pub-trigger.db.test.ts
```diff
@@ -281,3 +282,193 @@ describe("title trigger", () => {
...
+				Description: "Some description",
+			},
+		},
+	],
+});
+
+// when a pubType is updated, we need to update the title of all pubs in that pubType
+describe("pubType title change trigger", () => {
+	it("should update titles when a field is marked as title", async () => {
+		const { community, pubFields, pubTypes, pubs } = await seedCommunity(pubTitleTestSeed);
```
allisonking, line 325: hmm I think this may need to be dynamically imported in order not to persist

allisonking (approved):
nice 💯

Issue #761, opened 2024-11-07
Restrict pub visibility based on membership


3mcd:
## Motivation

So users only see pubs they have permission to view.

## Requirements

Modify pub-related queries such that:
- Community contributors can only see pubs they have an explicit membership to
- Community editors and admins can always see all pubs

## Acceptance Criteria

- Out of scope: filtering pubs in API routes.
- Uses strategy implemented in #671 to determine if the current user has the `viewPub` capability for each rendered pub.
- End-to-end tests that assert that pubs are only visible when the above conditions are met.

tefkah:
> Uses strategy implemented in https://github.com/pubpub/platform/issues/671 to determine if the current user has the viewPub capability for each rendered pub.

this might be too complex, but I feel like it would be much better to somehow write something that integrates with the current `getPubsWithChildrenAndRelatedPubs` fn to do authorization checks "in line" so to say. If I interpret correctly what is written here, the acceptance criteria would be to check for each individual pub, after being fetched, whether the user can see it. I think this will be too inefficient, and will make paginating a nightmare, I think it's worth investing the time in figuring out how to do this in the SQL.

PR #823, opened 2024-12-05
refactor: use new pub query in as many places as feasible

tefkah:
- **feat: use getPubsWithRelatedValues in for stages and pubs pages, and update types to be slightly more flexible**
- **feat: replace some more**

## Issue(s) Resolved

General slowness.

This PR 
- uses the `getPubsWithRelatedValues` query instead of `getPubs`, `getPubCached`, or `getPub` in all places, except for the integrations api
- renames `getPubs`, `getPubCached`, and `getPub` to `_deprecated_X` to signal we shouldn't use them anymore.
- Reworks the `getPubsWithRelatedValuesAndChildren` query to retrieve values later, making it much much more efficient for larger queries, and slightly more efficient for smaller queries.
- Add an option to the `getPubsWithRelatedValuesAndChildren` query to not fetch the values, speeding it up even more

This leads to eg the `/pubs` page loading very very quickly now, so much that changing it to load `100` instead of `10` pubs barely changes how fast it loads.


core/app/components/pubs/PubEditor/PubEditor.tsx
```diff
@@ -62,7 +78,15 @@ export async function PubEditor(props: PubEditorProps) {
...
 
 	const [pubs, pubTypes] = await Promise.all([
-		getPubs({ communityId: community.id }),
+		getPubsWithRelatedValuesAndChildren(
+			{ communityId: community.id },
+			{
+				withLegacyAssignee: true,
+				withPubType: true,
+				withStage: true,
+				limit: 30,
```
tefkah, line 87: extremely arbitrary number, in the future this will be an api call any so it's fine

packages/contracts/src/resources/site.ts
```diff
@@ -172,6 +189,18 @@ type MaybePubOptions = {
 	 * @default false
 	 */
 	withMembers?: boolean;
+	/**
+	 * Whether to include the legacy assignee.
+	 *
+	 * @default false
+	 */
+	withLegacyAssignee?: boolean;
```
tefkah, line 197: the assignee is still a thing, so we need to fetch it

packages/contracts/src/resources/site.ts
```diff
@@ -172,6 +189,18 @@ type MaybePubOptions = {
...
+	/**
+	 * Whether to include the values.
+	 *
+	 * @default boolean
+	 */
+	withValues?: boolean;
```
tefkah, line 203: this speeds up things a lot in places where we only fetch a list of pubs, as we can now rely on the `pub.title` being there

packages/contracts/src/resources/site.ts
```diff
@@ -207,17 +236,21 @@ type ProcessedPubBase = {
 };
 
 export type ProcessedPub<Options extends MaybePubOptions = {}> = ProcessedPubBase & {
+	/**
+	 * Is an empty array if `withValues` is false
+	 */
```
tefkah, line 241: i chose not to do the annoying thing i was doing before and make the values `?: never` when `withValues: false` is set, because too many places rely on there being some sort of values

core/lib/server/pub.ts
```diff
@@ -1337,48 +1327,63 @@ export async function getPubsWithRelatedValuesAndChildren<
...
+								"pv.updatedAt as updatedAt",
+								"pub_fields.schemaName",
+								"pub_fields.slug as fieldSlug",
+								"pub_fields.name as fieldName",
+							])
+							.whereRef("pv.pubId", "=", "pt.pubId")
+							.orderBy("pv.createdAt desc")
+					).as("values")
+				)
+			)
```
tefkah, line 1368: i think there's an even quicker way of doing this than this correlated subquery, but my brain is empty.
for now, this is efficient enough

core/lib/server/pub.ts
```diff
@@ -1217,35 +1203,35 @@ export async function getPubsWithRelatedValuesAndChildren<
...
+							.leftJoin("pub_values as pv", (join) =>
+								join.on((eb) =>
+									eb.and([
+										eb("pv.pubId", "=", eb.ref("p.id")),
+										eb("pv.relatedPubId", "is not", null),
+									])
+								)
+							)
+							.select("pv.relatedPubId")
+					)
```
tefkah, line 1219: this (and similarly in the `union` part) is the big change:

instead of joining ALL values up here before filtering and during the recursive lookup, we only join on the relatedPubId in order to traverse the relations. we then fetch the values separately at the end.

core/lib/server/pub.ts
```diff
@@ -1275,26 +1261,27 @@ export async function getPubsWithRelatedValuesAndChildren<
...
+									qb
+										.leftJoin("pub_values", (join) =>
+											join.on((eb) =>
+												eb.and([
+													eb("pub_values.pubId", "=", eb.ref("pubs.id")),
+													eb("pub_values.relatedPubId", "is not", null),
+												])
+											)
+										)
+										.select("pub_values.relatedPubId")
```
tefkah, line 1275: this conditional select is a bit sneaky, but i wanted to be able to avoid joining on the relatedPubIds if it's not necessary.

it's sneaky, bc it's kind of easy to forget, and if you don't use the same condition to select it below `selectFrom(pub_tree as pt)...` you won't get a type error but postgres will yell at you for selecting a column that doesn't exist

core/lib/server/pub.db.test.ts
```diff
@@ -395,24 +395,89 @@ describe("getPubsWithRelatedValuesAndChildren", () => {
...
 
-		expect(pubValues).toMatchObject({
-			values: [
-				{ value: "Some title" },
-				{
-					value: "test relation value",
-					relatedPub: {
-						values: [{ value: "test relation title" }],
-					},
+		pubValues.values.sort((a, b) => a.fieldSlug.localeCompare(b.fieldSlug));
```
tefkah, line 398: this should make the tests slightly more consistent

core/lib/pubs.ts
```diff
@@ -36,3 +40,44 @@ export const getPubTitle = (pub: PubTitleProps): string => {
...
+			},
+			{} as GetPubsResult[number]["values"]
+		),
+		stages: pub.stage ? [pub.stage] : [],
+		assigneeId: pub.assignee?.id ?? null,
+		assignee: (pub.assignee ?? null) as GetPubsResult[number]["assignee"],
+		pubType: pub.pubType as GetPubsResult[number]["pubType"],
+		children: pub.children.length ? pub.children.map(processedPubToPubResult) : [],
+	};
+};
```
tefkah, line 79: this is a bridge function for places where we still use the `{ slug: value }` pubvalues shape, rather than an array
this is eg the case in the contexteditor, which i didn't feel like changing

kalilsn: could you put this comment into the code? i'm definitely going to forget what the purpose of this is

kalilsn:
I think ideally I'll modify this new query to optionally filter pubs by membership for #761. So jsyk I am starting to work from this branch and would love to help you get it merged when I'm on tomorrow!

core/app/c/[communitySlug]/pubs/PubList.tsx
```diff
@@ -58,9 +25,18 @@ type PaginatedPubListProps = {
...
 			{ communityId: props.communityId },
-			{ limit: PAGE_SIZE, offset: (props.page - 1) * PAGE_SIZE, orderBy: "updatedAt" }
+			{
+				limit: PAGE_SIZE,
+				offset: (props.page - 1) * PAGE_SIZE,
+				orderBy: "updatedAt",
+				withPubType: true,
+				withRelatedPubs: false,
+				withStage: true,
+				withValues: false,
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/pubs/PubList.tsx:37]: 💯

kalilsn:
I knew it would be faster but it's literally 100x faster for the queries on the stages page which is better than I would have hoped!

kalilsn (approved):
Great improvement!

PR #826, opened 2024-12-05
Add default form when creating pub type

### LINKED ISSUE #782: Create Default Form When Pub Type Is Created
## Motivation
So admins have a form to edit when they create a new type.
So devs have a form to render when they create a new type.

## Requirements
- When a new Pub Type is created, an accompanying new default Form for the type is created, following the current heuristics for the default form.
- Default form has name “<Pub Type> Editor (Default)”

## Acceptance Criteria
- Should specify the same form input types given field schema types as the current default form editor
- Should use the title of the field, not the slug, as the “Label” for each form field.
- Should add a paragraph field at the top with the title of the Pub (or fallback) rendered as text (not markdown, for now).
- Should add a migration to create default forms for existing pub types

### LINKED ISSUE #817: Create markdown directive for Pub title field
## Motivation
So users can add the pub's title to a form without needing to know which field it is

## Requirements
- Renders the pub's canonical/selected title field for the pub type as html in markdown editors
- Allows users to select the title from a parent, like any other field selection

## Acceptance Criteria
- Uses helper from #736
- Uses format `:value{field=title}` or optionally `:value{field=title rel=parent}` (??)
- Should render the field's context as text, stripping any formatting if the field is encoded in markdown or html (the ideal is probably to preserve `<em>` or `<strong>` but to strip anything else, but we can do that later)

## Optional Work
- If the field contains markdown or HTML, strips all elements except `<em>` or `<strong>`

kalilsn:
## Issue(s) Resolved
Resolves #782 
Resolves #817 
- Adds a pubtitle directive: `:value{field="title"}` which uses the old pubtitle logic for now
- Adds a unique constraint to prevent duplicate named pub types. This makes it less likely that we'll encounter unique constraint errors when creating the default form
- Updates the form insert query to add a paragraph element with the new title directive as the first form element on new forms
- Factors out and uses the form insert query to automatically create a default form whenever a new pub type is created
- Adds a migration that replicates our default form creation logic for all the existing pub types

## Test Plan
Since the migration occurs before the seed script runs, try resetting your db (or not), then manually running the migration with psql or whatever db client you use. You should see a number of forms created for the existing pub types, which you can then test!

### DIFF
--- a/core/lib/server/form.ts
+++ b/core/lib/server/form.ts
@@ -14,7 +14,7 @@ import type {
 	PubTypesId,
 	UsersId,
 } from "db/public";
-import { AuthTokenType } from "db/public";
+import { AuthTokenType, ElementType, StructuralFormElement } from "db/public";
 
 import type { XOR } from "../types";
 import { db } from "~/kysely/database";
@@ -256,7 +256,15 @@ export const insertForm = (
 				})
 				.returning(["slug", "id"])
 		)
-
+		.with("title_element", (db) =>
+			db.insertInto("form_elements").values((eb) => ({
+				formId: eb.selectFrom("form").select("form.id"),
+				type: ElementType.structural,
+				element: StructuralFormElement.p,
+				content: '# :value{field="title"}',
+				order: 0,
+			}))
+		)
 		.insertInto("form_elements")
 		.columns(["fieldId", "formId", "label", "type", "order", "component"])
 		.expression((eb) =>
@@ -268,10 +276,11 @@ export const insertForm = (
 					"form.id as formId",
 					"fields.name as label",
 					eb.val("pubfield").as("type"),
-					eb.fn
-						.agg<number>("ROW_NUMBER")
-						.over((o) => o.partitionBy("id"))
-						.as("order"),
+					eb(
+						eb.fn.agg<number>("ROW_NUMBER").over((o) => o.partitionBy("id")),
+						"+",
+						1 // Offset order by 1 for the title element
+					).as("order"),
 					eb
 						.selectFrom("components")
 						.select("component")
--- a/core/lib/server/form.ts
+++ b/core/lib/server/form.ts
@@ -209,6 +209,7 @@ export const insertForm = (
 	name: string,
 	slug: string,
 	communityId: CommunitiesId,
+	isDefault: boolean,
 	trx = db
 ) => {
 	return trx
@@ -251,6 +252,7 @@ export const insertForm = (
 					pubTypeId,
 					slug,
 					communityId,
+					isDefault,
 				})
 				.returning(["slug", "id"])
 		)
--- a/core/lib/server/form.ts
+++ b/core/lib/server/form.ts
@@ -2,8 +2,18 @@ import type { QueryCreator } from "kysely";
 
 import { sql } from "kysely";
 import { jsonArrayFrom } from "kysely/helpers/postgres";
+import { componentsBySchema } from "schemas";
 
-import type { CommunitiesId, FormsId, PublicSchema, PubsId, UsersId } from "db/public";
+import type {
+	CommunitiesId,
+	CoreSchemaType,
+	FormsId,
+	InputComponent,
+	PublicSchema,
+	PubsId,
+	PubTypesId,
+	UsersId,
+} from "db/public";
 import { AuthTokenType } from "db/public";
 
 import type { XOR } from "../types";
@@ -12,6 +22,7 @@ import { createMagicLink } from "../authentication/createMagicLink";
 import { autoCache } from "./cache/autoCache";
 import { autoRevalidate } from "./cache/autoRevalidate";
 import { getCommunitySlug } from "./cache/getCommunitySlug";
+import { _getPubFields } from "./pubFields";
 import { getUser } from "./user";
 
 /**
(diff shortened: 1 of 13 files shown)


core/prisma/migrations/20241205231134_backfill_default_forms/migration.sql
```diff
@@ -0,0 +1,85 @@
...
+                    lower(trim("pub_types"."name")),
+                    '[^a-z0-9\\-_]+',
+                    '-',
+                    'gi'
+                )
+        ) || '-default-editor',
+        "pub_types"."communityId",
+        TRUE
+    FROM "pub_types"
+    ON CONFLICT DO NOTHING
```
kalilsn, line 47: Unless one of us intentionally creates a form on prod to cause a conflict here, this shouldn't be necessary since I've gone with a kind of ugly slug for the default form: `<pub_type>-default-editor`. But it's handy for testing that this doesn't fail if there's already an existing default form!

core/prisma/migrations/20241205231134_backfill_default_forms/migration.sql
```diff
@@ -0,0 +1,85 @@
...
+    "form"."id" AS "formId",
+    "pub_fields"."name" AS "label",
+    'pubfield' AS "type",
+    ROW_NUMBER() OVER (PARTITION BY "form"."id") + 1 AS "order",
+    "components"."component"
+FROM
+    "form"
+    INNER JOIN "_PubFieldToPubType" ON "_PubFieldToPubType"."B" = "form"."pubTypeId"
+    INNER JOIN "pub_fields" ON "pub_fields"."id" = "_PubFieldToPubType"."A"
+    INNER JOIN "components" ON "components"."schema" = "pub_fields"."schemaName";
```
kalilsn, line 85: This should be very similar to the generated form insert query, with a few changes:
- `ON CONFLICT` handling commented on above 
- Uses joins in the final part of the query for
	- fetching pub fields, as opposed to the convoluted json logic we end up doing in the app to reuse a kysely function
	- selecting the component, instead of a subquery
- It inserts one form per pub type in the `"form"` CTE above (rather than just a single form for the given pub type id)

core/prisma/migrations/20241205231134_backfill_default_forms/migration.sql
tefkah: i think that's fine, we only need to run this backfill once and i don't foresee us already having existing `pub_type-default-editor` slugged forms!

core/prisma/migrations/20241205231134_backfill_default_forms/migration.sql
tefkah: i fully trust you on this 🫡

core/lib/server/render/pub/renderMarkdownWithPub.ts
```diff
@@ -37,12 +38,19 @@ const visitValueDirective = (node: NodeMdast & Directive, context: utils.RenderW
...
 		const parentPub = expect(context.parentPub, "Missing parent pub");
-		value = parentPub.values[field];
+		pub = parentPub;
 	} else {
-		value = context.pub.values[field];
+		pub = context.pub;
+	}
+
+	if (field === "title") {
+		value = getPubTitle(pub);
```
tefkah, line 51: nice, this way #822 will just take care of correctly adding the title.
depending on the merge order, we will need to add `pubType` information on this pub however.

core/lib/server/form.ts
```diff
@@ -183,3 +194,100 @@ export const createFormInviteLink = async (props: FormInviteLinkProps) => {
...
+		.with("fields", () =>
+			_getPubFields({ pubTypeId, communityId })
+				.clearSelect()
+				.select((eb) => [
+					eb.ref("f.id").as("fieldId"),
+					eb.ref("f.json", "->>").key("name").as("name"),
+					eb
+						.cast<CoreSchemaType>(
+							eb.ref("f.json", "->>").key("schemaName"),
+							sql.raw('"CoreSchemaType"')
```
tefkah, line 242: neat!

core/lib/server/form.ts
```diff
@@ -183,3 +194,100 @@ export const createFormInviteLink = async (props: FormInviteLinkProps) => {
...
+const componentsBySchemaTable = Object.entries(componentsBySchema)
+	.map(([schema, components]) => {
+		const component = components[0]
+			? `'${components[0]}'::"InputComponent"`
+			: `null::"InputComponent"`;
+		return `('${schema}'::"CoreSchemaType", ${component})`;
+	})
+	.join(", ");
+
+export const insertForm = (
```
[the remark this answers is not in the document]
kalilsn: I went with `insert` here because I was extracting the query portion of the `createForm` action and I wanted to distinguish between the server action function and the database query function. My thinking was that `insert` is the name of the database operation, while `create` is the action a user wants to take. I'm open to something else but I was also following a pattern I used recently when trying to avoid having to alias the query functions when we have to import them in a server action like [we used to do on the members page](https://github.com/pubpub/platform/blob/c118f6b9a98291001debf2ac026502943c7b61ee/core/app/c/%5BcommunitySlug%5D/members/%5B%5B...add%5D%5D/actions.ts#L16-L20) before I made a similar change.

core/lib/server/render/pub/renderMarkdownWithPub.ts
```diff
@@ -37,12 +38,19 @@ const visitValueDirective = (node: NodeMdast & Directive, context: utils.RenderW
...
+	} else {
+		value = pub.values[field];
```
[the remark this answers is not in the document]
kalilsn: I am open to that but hesitant to spend more time on it because I think the syntax for markdown directives is really just too complicated for most users. And I think having an example of this one at the top of every default form is better documentation than any of the other directives have!

core/prisma/migrations/20241205231134_backfill_default_forms/migration.sql
kalilsn: >But it's handy for testing that this doesn't fail if there's already an existing default form!

I phrased this poorly. I meant that this on conflict statement makes it easier to test the migration because you can safely rerun it even if there are already some default forms.

But yes, I agree! Shouldn't be risky!

kalilsn:
I added an additional commit to create default forms in the seed script, and I'm going to go ahead and merge this once the tests pass so that @allisonking can use them. But feel free to leave more feedback about the directive/function names, I can address it in a separate PR @tefkah .

PR #828, opened 2024-12-09
Fix dropdown menuitem styling

kalilsn:
## Issue(s) Resolved
Our dropdown components were looking like 
image
instead of 
image
because of an error I made when applying a shadcn update in https://github.com/pubpub/platform/commit/97325861ec81f4964b99d96e2f59cb27f059710f#diff-c611e07f0c0ccb0a999e23be580a30ef339e4f8341df5d11dd33300ad184967fL85

I replaced the `...` with the implied content because I am a human and not a copy-paste robot.


tefkah:
Nice, was wondering how that happened!

tefkah (approved):

PR #831, opened 2024-12-10
Warn users about the default form when they modify a pubtype

### LINKED ISSUE #784: Warn user about default form when updating Pub type
## Motivation
So admins know they need to update the default Pub form if they update the Pub type

## Requirements
- After modifying a pub type, users are warned: ```These updates will not be automatically reflected in the default editor form for <PubType>. <Click here> to manually update the default editor form.```

kalilsn:
## Issue(s) Resolved
Resolves #784 
- Adds a toast with a link to the default form when a field is added or removed from a type
- Removes the now defunct pub type deletion button
## Test Plan
Visit http://localhost:3000/c/croccroc/types and add or remove a field from a type. You should see a toast, and clicking the action should take you to the form builder page for editing that type's default form.
image

## Notes
I modified the language in the toast a bit:
- Changed "these updates" to "this update" because they're not batched
- Removed "Click here" for [accessibility reasons](https://www.w3.org/TR/WCAG20-TECHS/F84.html)
- Restructured so the link would be inside the action button on the toast because it's much easier to interact with that way


allisonking (approved):
works great!

PR #834, opened 2024-12-11
Add fields to pub type before default form in seed script

kalilsn:
## Issue(s) Resolved
Seeded forms only had the title paragraph element, but no input elements because they were created before any fields were added to the pub type


allisonking (approved):
works great 💯 thanks for the quick fix!

Issue #758, opened 2024-11-06
Add authentication and authorization checks to unsecured server actions


gabestein:
## Motivation
A bunch of server actions atm are not secured at all. While easy to forget, these things are basically API endpoints, and atm for a lot of actions just about anyone can access them.

## Requirements

- Add authentication checks to all server actions
- Move exports of non-client called functions out of files with `"use server"`
- Add authorization checks to relevant server actions
- Sends authorization error with generic "You are not authorized to perform this action." message

### Missing authentication & Authorization checks
- [ ] All the actions for the formbuilder: 
https://github.com/pubpub/platform/blob/605cb02e1bb2cdb12ce13fbd6991797dd8e9c78e/core/app/components/FormBuilder/actions.ts#L15-L16

- [x] All `fields` server actions
https://github.com/pubpub/platform/blob/bd08a370f26b47b434ac5cc3fe7adb815b9d68e0/core/app/c/%5BcommunitySlug%5D/fields/actions.ts#L1-L73

- [x] The `move` and `assign` actions
https://github.com/pubpub/platform/blob/e8baa1b18ce0777f3831e5659297701c91f67ab5/core/app/c/%5BcommunitySlug%5D/stages/components/lib/actions.ts#L1-L50

- [ ] All the `stages/manage` actions
https://github.com/pubpub/platform/blob/e8baa1b18ce0777f3831e5659297701c91f67ab5/core/app/c/%5BcommunitySlug%5D/stages/manage/actions.ts#L1-L296

- [ ] All the `types` actions
https://github.com/pubpub/platform/blob/180712d166b9dd231fe6ef5958fd30701690dfa7/core/app/c/%5BcommunitySlug%5D/types/actions.ts#L1-L78

- [x] The `_updatePub` function should be moved to `lib/server/pub.ts`. Currently it's exported and therefore a server action
https://github.com/pubpub/platform/blob/b9914b22cc7eeb55f9986cafaf6bf9f42789b4ac/core/app/components/pubs/PubEditor/actions.ts#L46-L56

### Missing authorization checks (in addition to the above)

(can't find that very quickly, should probably be rethought anyway once we have our new permissions model in place)

PR #835, opened 2024-12-11
Secure server actions

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/758

All actions should now at the minimum do an authentication check that the user is logged in. 

I added authorization checks too where I could, though please check my assumptions! There were some I was not sure about, namely:

* Actions
* Rules
* Move constraints

I also haven't added the ones to the form builder yet since `userCan` does not cover the form memberships table yet, and I wasn't sure if that had to be specced out more or if there was another ticket for it?

## Test Plan

Honestly I'm not sure the best way to test it—probably make users of different roles and try out the actions? I wrote a few unit tests, and the "Inviting a new user to fill out a form" E2E test was good for testing a new user who only had access to a form.

### DIFF
--- a/core/app/c/[communitySlug]/stages/manage/actions.ts
+++ b/core/app/c/[communitySlug]/stages/manage/actions.ts
@@ -32,6 +32,7 @@ import {
 } from "~/lib/server/actions";
 import { autoRevalidate } from "~/lib/server/cache/autoRevalidate";
 import { revalidateTagsForCommunity } from "~/lib/server/cache/revalidate";
+import { findCommunityBySlug } from "~/lib/server/community";
 import { defineServerAction } from "~/lib/server/defineServerAction";
 import { insertStageMember } from "~/lib/server/member";
 import { createRule, removeRule } from "~/lib/server/rules";
@@ -49,7 +50,22 @@ async function deleteMoveConstraints(moveConstraintIds: StagesId[]) {
 	if (!loginData || !loginData.user) {
 		return ApiError.NOT_LOGGED_IN;
 	}
-	// TODO: add authorization checks for move constraints?
+
+	const community = await findCommunityBySlug();
+
+	if (!community) {
+		return ApiError.COMMUNITY_NOT_FOUND;
+	}
+
+	const authorized = await userCan(
+		Capabilities.editCommunity,
+		{ type: MembershipType.community, communityId: community.id },
+		loginData.user.id
+	);
+
+	if (!authorized) {
+		return ApiError.UNAUTHORIZED;
+	}
 
 	await autoRevalidate(
 		db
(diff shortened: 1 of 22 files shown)


core/app/c/[communitySlug]/stages/manage/actions.ts
```diff
@@ -154,14 +217,34 @@ export const updateStageName = defineServerAction(async function updateStageName
...
+		return ApiError.NOT_LOGGED_IN;
+	}
+
+	const { user } = loginData;
+
+	const authorized = await userCan(
+		Capabilities.manageStage,
+		{ type: MembershipType.stage, stageId },
+		user.id
+	);
```
Comment (allisonking) [inline core/app/c/[communitySlug]/stages/manage/actions.ts:243]: not sure if this is the right one for `addAction`...

core/app/c/[communitySlug]/stages/manage/actions.ts
```diff
@@ -44,6 +45,12 @@ import {
 import { createUserWithMembership } from "~/lib/server/user";
 
 async function deleteMoveConstraints(moveConstraintIds: StagesId[]) {
+	const loginData = await getLoginData();
+	if (!loginData || !loginData.user) {
+		return ApiError.NOT_LOGGED_IN;
+	}
+	// TODO: add authorization checks for move constraints?
```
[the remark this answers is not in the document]
Comment (kalilsn) [reply core/app/c/[communitySlug]/stages/manage/actions.ts:52]: https://github.com/pubpub/platform/issues/595
image

actually I see now that stage admins are allowed to modify stage settings in general, so I guess `editCommunity` isn't appropriate!

core/app/c/[communitySlug]/stages/manage/actions.ts
Comment (kalilsn) [reply core/app/c/[communitySlug]/stages/manage/actions.ts:243]: i believe it is! and it's also the right one for the other action-actions below

core/app/components/FormBuilder/actions.ts
```diff
@@ -9,10 +9,18 @@ import { logger } from "logger";
...
 import { db } from "~/kysely/database";
 import { isUniqueConstraintError } from "~/kysely/errors";
+import { getLoginData } from "~/lib/authentication/loginData";
+import { ApiError } from "~/lib/server";
 import { autoRevalidate } from "~/lib/server/cache/autoRevalidate";
 import { defineServerAction } from "~/lib/server/defineServerAction";
 
 export const saveForm = defineServerAction(async function saveForm(form: FormBuilderSchema) {
+	const loginData = await getLoginData();
+
```
kalilsn, line 19: I think these form builder actions should actually just use the `editCommunity` capability. I don't think the form permissions are actually relevant for these actions - they're more for authorizing users to edit/create pubs with specific forms when they might otherwise not have any pub edit/create permissions in the community.

core/app/components/pubs/PubEditor/actions.ts
```diff
@@ -49,23 +57,38 @@ export const updatePub = defineServerAction(async function updatePub({
...
 
 	const community = await findCommunityBySlug();
 
 	if (!community) {
-		throw new Error("Community not found");
+		return ApiError.COMMUNITY_NOT_FOUND;
+	}
+
+	const canUpdateFromForm = formSlug
+		? await userHasPermissionToForm({ formSlug, userId: loginData.user.id })
```
kalilsn, line 82: I guess this isn't changing the situation but this allows anyone who has been invited to a form to edit any pub in the community with that form. I know you didn't introduce that problem but could you add a TODO comment here?

kalilsn (approved):
This looks great! Thanks for standardizing the errors and writing good tests! There are a few places where I think we do have the capability required that you could add additional authorization checks but those can also come later.

core/app/c/[communitySlug]/stages/manage/actions.ts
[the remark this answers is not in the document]
Comment (allisonking) [reply core/app/c/[communitySlug]/stages/manage/actions.ts:52]: okay! and for move constraints, would we want to check that they have permission to all stages listed in the move constraints?

core/app/components/pubs/PubEditor/actions.ts
allisonking: okay added a TODO for now!

core/app/components/FormBuilder/actions.ts
allisonking: sounds good, I've added `editCommunity` to each of these

PR #836, opened 2024-12-12
dev: add history log table for pub_values

### LINKED ISSUE #796: Pub Field Log Tables
## Motivation

So devs can build future version history features for Pubs
So devs can build future front-ends for readers so they can browse past versions of pubs and potentially see diffs
So community admins can eventually have an audit trail log of who modified what on Pubs in case they later need to investigate

## Requirements

- Logs all changes to Pub fields
- Logs the "responsible party" (PubPub user, API key, or action run) for the made changes
- In the future, should allow for someone to re-assemble a version of a Pub by querying by the Pub ID and a timestamp
- If a user, api key, or action run has been deleted, logs will be able to report that an "unknown user|api key|action" are responsible for the change.
- Dev comes to sprint planning with strategy for implementing tracking of user/api key/action run responsible for changes

## Acceptance Criteria

- Create an audit table that stores pub field values in addition to nullable columns for userId, apiKeyId, actionRunId
- Picks one strategy for implementing tracking of "responsible party":
   - a. Add a required `lastModifiedBy` column to the pub field values table that contains something like: `user:xxx-xxx | apiKey:xxx-xxx | actionRun:xxx-xxx` and create a trigger that runs before the statement and validates the text and the existence of the responsible party on insert
   - b. Create nullable columns for user, api, and action to match audit table as foreign keys
   - c. Accept that responsible party data might not successfully insert, and make sure our audit trigger handles that appropriately.
- Update app queries that modify pub field values to update the lastModifiedBy field (or, if we choose strategy b, adds the correct id to the correct column) with the source of the modification, whether a user, api key, or action run
- Tests that verify changes to pub fields and "responsible party" changing them are recorded
- Does not implement query to reassemble pub state at given time — we will implement this later
- If we use the `lastModifiedBy` column strategy, implements a migration that backfills that new column with (a non-existing user?), so that versions can later be created that say the field was modified by an "unknown user"

tefkah:
## Issue(s) Resolved
#796 

This PR adds a history table for the pub_values table, which logs changes to that table and who/what made that change.
These entries are almost exclusively added through triggers, with the exception of `delete` operations, see below.
The design of the table is kind of generic to allow us to implement a similar history table for other tables quite easily.

I created a script which auto-creates such a table for any existing table, and used it to create this initial one for `pub_values`.

### General design of the table

The table needs to meet the following requirements
- Allow us to reconstruct a pub_value at any point in time
- Act as an audit table, which allows us to see who edited what.

This second requirement is where most of the complexity of this table lies. Without it, a simple trigger system would have been sufficient, and we could eliminate most of the columns.

The table looks as follows 

```prisma
model PubValueHistory {
  histId        String        @id @default(dbgenerated("gen_random_uuid()"))
  createdAt     DateTime      @default(now())
  operationType OperationType

  // has check constraint to ensure that oldRowData and newRowData are not both null
  // see ./migrations/20241203164958_add_history_table/migration.sql
  // type is the type of the Table that is being changed, e.g `PubValues` for PubValuesHistory
  // using a kysely pre-render hook
  oldRowData Json?
  newRowData Json?

  // primary key of the row that was changed
  primaryKeyValue String?

  // identifying information
  user             User?           @relation(fields: [userId], references: [id])
  userId           String?
  apiAccessToken   ApiAccessToken? @relation(fields: [apiAccessTokenId], references: [id])
  apiAccessTokenId String?
  actionRun        ActionRun?      @relation(fields: [actionRunId], references: [id])
  actionRunId      String?
  // set to `system` if the change was made by the system, eg during seeds
  other            String?

  @@map(name: "pub_values_history")
}
```

I won't spend any time explaining the top colunms, except to ask for the following feedback

> [!NOTE]
> - Should `histId` just be called `id`? I just copied this from https://gist.github.com/slotrans/353952c4f383596e6fe8777db5d098f0
> - Any better name for `primaryKeyValue`? Again, just copied that

With that out of the way, we can look at the other columns.

Unless I missed some case, a PubValue (or almost anything for that matter), can be changed by one of the following entities
- A user through the UI
- Someone through the API
- An action, during the running of an action
- A seed command, or other direct manipulation of the db by us

The `userId`, `apiAccessTokenId`, `actionRunId`, and `other` column are intended to capture this data.
I have not made them mutually exclusive for now, as we might want to capture eg both the `actionRunId` and the `userId` when a user triggers an action.

The `other` column is the most vague, which i'm only setting to `system` for now if the PubValue-modfiying entity is a seed script or test or something. 

When all columns are null (which should only really happen when the modifying user/apiToken/actionRun is deleted from the db), the application should render the modifier as `unknown`.

### How is the modifier captured?

The modifying entity is captured by adding a `lastModifiedBy` column to the to-be-tracked table, in this case `pub_values.lastModifiedBy`.
This column is of a custom type `last_modified_by`, which is a  non-nullable `TEXT` column with a special `CHECK` constraint, which checks that the entered meets the following format:

- `user:<userId>|<timestamp>`
- `api-token:<apiAccessTokenId>|<timestamp>`
- `action-run:<actionRunId>|<timestamp>`
- `system|<timestamp>`
- `unknown|<timestamp>`

The trigger added to the to-be-tracked table then extracts the entity-type and entity-id from this column and uses it to set the correct data on the history table.

This is roughly how the table operates, which is i think fine in general. The main annoying part currently is that you must provide the `lastModifiedBy` column manually when creating and updating data, but there's no real way around that.

However, there are a couple of annoying caveats

#### Updates

You probably wondered what the `timestamp` was for in the `lastModifiedBy` table. The sole purpose of this is to ensure that you provide the `lastModifiedBy` column on `updateTable` and `doUpdateSet` calls.

Q: Why does this have to be constraint?

A: Because it's very easy to miss, and we want to capture this information.

Q: Can't we just do this at the type-level only?

A: That would be good enough imo, but AFAICT there is NO way in kysely to make a column required during updates on the type level. It might be possible as a plugin, but that would be slightly more annoying to implement, although it is a possibility if you think that's better.

Q: Why the timestamp though? 

A: Because there is (again, AFAICT) NO way in Postgres to get a simple list of the columns that were updated in a trigger. Please correct me if i'm wrong, but the only thing you can do is compare the old column with the new column. This means that there is no way of knowing whether  `OLD_ROW.lastModifiedBy === NEW_ROW.lastModifiedBy` is due to the same entity modifying the row, or whether you forgot to set the `lastModifiedBy` in the `UPDATE` query (bc again, there is no way of checking whether you included that or not.) The only way therefore to check whether you included the lastModifiedBy is to guarantee it to be different from the current one, which in this case i did by including a timestamp.
This is extremely ugly, if you have a better idea please let me know.

> [!NOTE]
> Do you have a better idea of guaranteeing that we set `lastModfiedBy` in update queries?

#### Deletes

Bc of this design of reading the `lastModifiedBy` from the `NEW` row to determine the modifying entity, there is no way of getting that information on `DELETE` queries. Since there is in general no way to read metadata from delete queries, i fear that this is just a general limitation we have to work with if we want to capture this information.

The solution: manually write this info to the `pub_values_history` table on delete. The number of deletes we do are small so there's not that much risk in forgetting, but it is still there.

There are three places where we do deletes for `pub_values`:

- When we delete a pub
https://github.com/pubpub/platform/blob/8d974c106b3baaaf2e25f2d66bd6229e2594d65d/core/lib/server/pub.ts#L600-L630

- When we delete related pubs
https://github.com/pubpub/platform/blob/8d974c106b3baaaf2e25f2d66bd6229e2594d65d/core/lib/server/pub.ts#L877-L919

- When we replace related pubs
https://github.com/pubpub/platform/blob/8d974c106b3baaaf2e25f2d66bd6229e2594d65d/core/lib/server/pub.ts#L996-L1020

Here i've manually added entries into the `pub_values_table`.

There are a number of ways we could do this differently, all with a different tradeoff

##### 1. Never delete things

If we never delete eg `pub_values` and `pubs`, instead just archive things, we would not need to worry about logging `DELETE` operation to `pub_values_history`.

##### 2. Strongly warn users about deletes and don't log them

A middle ground with the above would be to more strongly warn users that certain operations are not undoable.
For instance, if we archive pubs by default but still allow you to delete them if they're archived, we could warn users that this will not be restorable. In that case, we would maybe not need to log the `pub_values`, in fact we should probably delete all the existing entries for that pub.

##### 3. Not care about the entity that did the delete
We could simply always log `DELETE`s as `system`. That way we can do it with a trigger only.

##### 4. Never log deletes
We could also just not log them. Not preferred i think.

## Test Plan

Run `pnpm --filter core migrate-dev`, and check that no resets are necessary.

Run tests.

Test the following scenarios by seeing that an appropriate entry is created in `pub_values_history` using `pnpm db:studio`. The values in parentheses are what's expected in the `operationType` and `userId`|`apiAccessTokenId`|`actionRunId`|`other` respectively.

1. Adding a pub (`insert`, `userId: <id>`)
2. Updating a pub (`update` for existing values, `insert` for new ones, `userId: <id>`)
3. Ideally check that updating the same value again will create a new entry.
4. Deleting a pub (`delete`, `userId: <id>`)
5. Running the Google Drive/`http` action (`insert`|`update`, `actionRunId: <id>`)
6. (optional) create/update a pub through the api (`insert`|`update`, `apiAccessToken: <id>`)

## Notes

In general would like your thoughts on whether you think this (very) hybrid db trigger / application code approach is good enough and on the design of the table!


core/actions/_lib/runActionInstance.ts
```diff
@@ -218,8 +235,9 @@ export async function runActionInstance(args: RunActionInstanceArgs) {
 				actionInstanceId: args.actionInstanceId,
 				pubId: args.pubId,
 				userId: isActionUserInitiated ? args.userId : null,
-				status: "error" in result ? ActionRunStatus.failure : ActionRunStatus.success,
-				result,
+				result: { scheduled: `Action to be run immediately` },
+				// we are setting it to `scheduled` very briefly
+				status: ActionRunStatus.scheduled,
```
tefkah, line 240: Let me know if you have a better idea for this, but the idea is to first set the actionRun to `pending`, then set it to `success` or `error` after it resolves.

Feels kind of silly though.

core/kysely/errors.ts
```diff
@@ -4,8 +4,8 @@ import { databaseTableNames } from "db/table-names";
 
 const PostgresError = z.object({
 	code: z.string(),
-	detail: z.string(),
-	table: z.enum(databaseTableNames),
+	detail: z.string().optional(),
+	table: z.enum(databaseTableNames).optional(),
```
tefkah, line 8: these aren't there for `ForeignKeyConstraint` errors

core/lib/lastModifiedBy.ts
```diff
@@ -0,0 +1,84 @@
...
+	if (props.actionRunId) {
+		return `action-run:${props.actionRunId}|${timestamp}`;
+	}
+
+	if (props.apiAccessTokenId) {
+		return `api-access-token:${props.apiAccessTokenId}|${timestamp}`;
+	}
+
+	throw new Error("Invalid lastModifiedBy");
+};
```
tefkah, line 84: the main intent of this fn is to not have to make you think about adding the timestamp

core/lib/server/pub-trigger.db.test.ts
tefkah, line 1: are there any obvious test cases i've missed?

core/lib/server/pub.db.test.ts
```diff
@@ -85,6 +86,7 @@ describe("createPubRecursive", () => {
 					[`${community.slug}:title`]: "test title",
 				},
 			},
+			lastModifiedBy: createLastModifiedBy("system"),
```
tefkah, line 89: having to add this everywhere in tests is annoying, but it's better than forgetting to put it there when you should! this is why is chose to not include a default for `lastModifiedBy`: we should always think about who's modifying the content.

core/lib/server/pub.ts
```diff
@@ -587,8 +597,37 @@ export const createPubRecursiveNew = async <Body extends CreatePubRequestBodyWit
...
+	).executeTakeFirstOrThrow();
+
+	// this might not be necessary if we rarely delete pubs and
+	// give users ample warning that deletion is irreversible
+	// in that case we should probably also delete the pub values
+	await addDeletePubValueHistoryEntries({
+		lastModifiedBy,
+		pubValues,
+		trx,
+	});
```
tefkah, line 627: i don't like doing this, see pr description for alternatives.

core/package.json
```diff
@@ -7,6 +7,14 @@
...
 		"playwright:test:dev": "TEST_DEV=true playwright test",
 		"playwright:ui:dev": "TEST_DEV=true playwright test --ui",
+		"db:migrate-dev": "pnpm migrate-dev",
+		"db:migrate-deploy": "pnpm migrate-deploy",
+		"db:migrate-diff": "pnpm migrate-diff",
+		"db:migrate-docker": "pnpm migrate-docker",
+		"db:migrate-test": "pnpm migrate-test",
+		"db:prisma": "pnpm prisma",
+		"db:studio": "pnpm prisma studio",
+		"db:generate-history-table": "pnpm exec tsx prisma/scripts/history-tables/generate-history-table.mts",
```
tefkah, line 17: i thought it be neater this way, instead of just having a bunch of random command names everywhere.

core/playwright/fixtures/fields-page.ts
```diff
@@ -31,11 +31,23 @@ export class FieldsPage {
...
+		waitForText = true
+	) {
 		await this.openNewFieldModal();
 		await this.nameBox.fill(name);
 		await this.selectFormat(format);
 		await this.page.getByRole("button", { name: "Create" }).click();
+		if (!waitForText) {
+			return;
+		}
+		await this.page.getByText(`Created field ${name}`, { exact: true }).waitFor();
```
tefkah, line 50: for some reason i ran into the problem a lot that the navigation from the field page was too quick, so i added a waitFor

core/prisma/migrations/20241203164958_add_base_history_fn/migration.sql
```diff
@@ -0,0 +1,97 @@
...
+CREATE TYPE "OperationType" AS ENUM(
+    'insert',
+    'update',
+    'delete'
+);
+
+CREATE DOMAIN modified_by_type AS TEXT
+CHECK (
+    VALUE ~ '^(user|action-run|api-access-token):[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}\|\d+$|^(unknown|system)\|\d+$'
+);
```
tefkah, line 11: sad we can't specify these in prisma!
this basically checks whether the value is of the correct format. added as a `DOMAIN` type st we can always set the `lastModfiedBy` column to be this type, rather than adding a constraint for every table that uses this colum.

core/prisma/migrations/20241203164958_add_base_history_fn/migration.sql
```diff
@@ -0,0 +1,97 @@
...
+
+    IF TG_OP = 'INSERT' THEN
+        -- if it's an insert on conflict, we'll treat it as an update,
+        -- and we include the old row data as the new row data
+        -- v_oldRowData := (select * from pub_values where "pubId" = NEW."pubId" and "fieldId" = NEW."fieldId");
+        -- if v_oldRowData is not null then
+        --     v_operationType := 'update'::"OperationType";
+        -- else
+        --     v_operationType := 'insert'::"OperationType";
+        -- end if;
```
tefkah, line 67: i need to remove this

core/prisma/migrations/20241212164758_fix_cascading_delete_issues/migration.sql
tefkah, line 1: this is so you can actually delet api access tokens

core/prisma/migrations/20241214173138_add_pub_values_history_history_table/migration.sql
tefkah, line 1: see the template for explanatory colunms

core/prisma/schema/schema.prisma
```diff
@@ -451,20 +458,21 @@ enum Action {
...
   status           ActionRunStatus
   result           Json
-  user             User?           @relation(fields: [userId], references: [id])
+  user             User?             @relation(fields: [userId], references: [id], onDelete: SetNull)
   userId           String?
-  createdAt        DateTime        @default(now())
-  updatedAt        DateTime        @default(now()) @updatedAt
+  createdAt        DateTime          @default(now())
+  updatedAt        DateTime          @default(now()) @updatedAt
+  PubValueHistory  PubValueHistory[]
```
tefkah, line 475: it's kind of annoying all of these tables need a references to `XHistory`, but so be it

core/prisma/scripts/history-tables/generate-history-table.mts
tefkah, line 1: in order, this script roughly, for an arbitrary table X

- adds a new `XHistory.prisma` file to `prisma/schemas/history-tables` containing an `XHistory` model which looks identical to `PubValueHistory`, minus  the name of the table
- Adds a `lastModifiedby` column to  `X`
- Runs `prisma format`, which will add all the correct references to all the other models
- Creates a migration from `history-table-migration-template.sql`

core/prisma/scripts/history-tables/history-table-migration-template.sql
```diff
@@ -0,0 +1,67 @@
...
+ - Added the required column `lastModifiedBy` to the `{{tableName}}` table without a default value. This is not possible if the table is not empty.
+ */
+-- AlterTable
+-- first we add the column with a default value
+ALTER TABLE "{{tableName}}"
+  ADD COLUMN "lastModifiedBy" modified_by_type NOT NULL DEFAULT CONCAT('unknown', '|', FLOOR(EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) * 1000));
+
+-- then we remove the default value
+ALTER TABLE "{{tableName}}"
+  ALTER COLUMN "lastModifiedBy" DROP DEFAULT;
```
tefkah, line 13: we first set `lastModifiedBy` to `unknown|${Date.now()}` for all rows
then we drop the default

core/prisma/scripts/history-tables/history-table-migration-template.sql
```diff
@@ -0,0 +1,67 @@
...
+  "newRowData",
+  "primaryKeyValue"
+)
+SELECT
+  'insert'::"OperationType",
+  NULL,
+  row_to_json(t),
+  t.id
+FROM
+  "{{tableName}}" t;
```
tefkah, line 61: we create a single row with no perpetrator for each row currently in the table, which we label as `insert`.

core/prisma/scripts/history-tables/history-table-migration-template.sql
```diff
@@ -0,0 +1,67 @@
...
+
+CREATE TRIGGER trigger_{{historyTableName}}
+  AFTER INSERT OR UPDATE ON {{tableName}}
+  FOR EACH ROW
+  EXECUTE FUNCTION f_generic_history();
```
tefkah, line 66: finally, we add the trigger on `update` and `insert`

I think this technically could be more efficient if we did `FOR EACH STATEMENT` rather than `FOR EACH ROW`, but i'm not convinced it matters much

package.json
```diff
@@ -56,7 +56,7 @@
 	"prettier": "@pubpub/prettier-config",
 	"packageManager": "pnpm@9.10.0",
 	"prisma": {
-		"schema": "core/prisma/schema.prisma"
+		"schema": "core/prisma/schema/"
```
tefkah, line 59: we now look for any `*.prisma` file in this folder

core/actions/_lib/runActionInstance.ts
kalilsn: makes sense to me! "scheduled" seems accurate enough

kalilsn (approved):
Very nice work. I feel totally fine about the hybrid approach, since I was in favor of an all application approach in the beginning. I left a few comments but pretty much all superficial so I'm approving.

> Should histId just be called id?
I think so!

>Do you have a better idea of guaranteeing that we set lastModfiedBy in update queries?

I don't. I'm assuming from your comment that we'd need to modify kysely itself, not just the types generated by kanel to enforce that?

core/lib/server/pub.ts
tefkah: yes, you're right. what i mean to say is that we should also delete the pub_values_history table!

core/prisma/migrations/20241203164958_add_base_history_fn/migration.sql
```diff
@@ -0,0 +1,97 @@
...
+CREATE DOMAIN modified_by_type AS TEXT
+CHECK (
+    VALUE ~ '^(user|action-run|api-access-token):[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}\|\d+$|^(unknown|system)\|\d+$'
+);
+
+CREATE OR REPLACE FUNCTION f_generic_history()
+    RETURNS TRIGGER
+    AS $$
+DECLARE
+    vc_insert_sql constant text := 'insert into ' || TG_TABLE_NAME || '_history ( "operationType", "oldRowData", "newRowData", "primaryKeyValue", "userId", "apiAccessTokenId", "actionRunId", "other" ) values ( $1, $2, $3, $4, $5, $6, $7, $8 )';
```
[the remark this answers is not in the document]
tefkah: i changed it to be `pubValuesId`, eg. this name needs to be passed in through the trigger function, which the script takes care of automatically, bc figuring out how to change eg `pub_values` to `pubValueId` would be too annoying in postgres for me. maybe we should at some point just use python for these things haha

PR #839, opened 2024-12-16
Make login redirect work again and remove login notice

kalilsn:
## Issue(s) Resolved
Non-logged in users were always being sent to `https://app.pubpub.org/settings` which returns the next default error page - a significantly worse experience than before. Now they should get redirected to /login with the appropriate redirect param.

I also deleted the lucia migration notice because it's been a while!
I made sure to use `getPageLoginData` anywhere that we'd like a login redirect to happen for unauthenticated users!

## Test Plan
Visit localhost:3000/c/unjournal/stages locally without being logged in. You should be redirected to the login page

### DIFF
--- a/core/app/c/[communitySlug]/layout.tsx
+++ b/core/app/c/[communitySlug]/layout.tsx
@@ -3,7 +3,7 @@ import type { Metadata } from "next";
 import { redirect } from "next/navigation";
 
 import { CommunityProvider } from "~/app/components/providers/CommunityProvider";
-import { getLoginData } from "~/lib/authentication/loginData";
+import { getPageLoginData } from "~/lib/authentication/loginData";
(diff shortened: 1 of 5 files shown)


core/app/c/[communitySlug]/layout.tsx
```diff
@@ -26,7 +26,7 @@ export async function generateMetadata({
 }
 
 export default async function MainLayout({ children, params }: Props) {
-	const { user } = await getLoginData();
+	const { user } = await getPageLoginData();
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/layout.tsx:29]: This was the important change, the others were just rote cleanup. But this makes sure logged in users get the proper login redirect rather than a `/settings` redirect.

tefkah (approved):
Nice, good catch, that was a really weird flow

PR #844, opened 2024-12-17
fix: make updatedAt plugin add updatedAt on insert.onConflict

tefkah:
- **fix: add oncflict updatedAt plugin**
- **dev: add tests for updatedAt plugin**

## Issue(s) Resolved

`updatedAt` was not set for `pubvalues` bc we started setting it using `onConflict`, which was not caught by the plugin.

## Test Plan
1. Run tests
2. Create a new pub
3. Update one of its values
4. Using `pnpm db:studio`, convince yourself that for that value the `updatedAt` is newer than the `createdAt`


kalilsn:
I think this is ready for review, but curious how you feel about this approach @tefkah. Before my changes the tests failed because the plugin broke the form builder updates (which also use on conflict). They failed because form_elements didn't have an `updatedAt` column, and there doesn't seem to be any way to check if the table has an `updatedAt` column from the data on an `OnConflictNode` alone (the table for all the updates is just "excluded").

My simple solution is just to add `updatedAt` to everything, since it's useful and I totally meant to put it on form_elements initially. If that seems insufficient, lmk and I will see if I can figure out a way to determine whether there's an `updatedAt` column available in the `transformOnConflict` method!

tefkah:
ah fantastic, thanks for figuring out what the issue was!

i'm not super sure on adding `updatedAt` to everything. I do think it makes sense to do it for the form elements, but i'm slightly more hesitant about adding it to "PubsInStages" and other places for instance.

I will push up a change that ideally will detect the `updatedAt` column properly by looking at the `insert` node, rather than the `onConflict` one (kysely really should allow you to inspect the parent node tbh)

What do you think of that? I think in plenty of places we can leave the `updatedAt`, but I think for eg all the authtoken stuff and `PubsInStages` (maybe) i'd rather the plugin work correctly than us having to always remember to put in an `updatedAt` column to relation tables.

kalilsn:
Your solution is definitely better and safer! I just wasn't sure how to do it immediately, although it makes sense to me now. 

But I would still like to add `updatedAt` to everything in addition, since there's no real cost (AFAIK) and I suspect we will later want it on both of your examples. Thinking about this also reminded me that we might want to update a pub's `updatedAt` when it changes stages too.

kalilsn (approved):
Great tests and good job noticing this was happening! Feel free to merge it if you're ok with leaving those migrations in. LMK if you would rather discuss that more, or if you think we need someone else to review this!

tefkah:
Im fine with leaving the `updatedAt`'s!

ill merge it in, thanks for figuring out what was breaking the tests!
and yeah, we should update a pubs updatedAt when we move it, probably as a trigger.

PR #845, opened 2024-12-17
feat: add back api docs for site building api

tefkah:
- **dev: add back openapi spec**
- **chore: reorder things**
- **feat: make docs static**
- **chore: add prettier ignore for the stoplight styles**

## Issue(s) Resolved
No api docs!

Re add API documentation, but only for the `/site` api.

I removed the `@stoplight/elements` NPM package and instead just used it from a `script` tag, as `@stoplight/elements` still uses React 16 which is causes some conflicts for us. Easier to just load it through a script than have `next` resolve react version conflicts.

## Test Plan

1. Go to `/c/arcadia-research/settings`
2. Click on `API Docs`
3. Wow, see docs!
4. Click on the Get Pubs thing
5. Add `Bearer <default token>` to the `Authorization` on the right, click run
6. Wow, you can actually run things! (we may want to disable this on prod, as you will interact will live data)


gabestein (approved):
🥳 holy wow this is amazing!

Yes, I think we should probably disable running stuff directly on prod.

PR #846, opened 2024-12-17
fix: allow updates to DateTime CoreSchemaType values through the API

### LINKED ISSUE #837: Allow API to interpret date-strings sent over the wire
## Motivation
So devs can add and update pub fields with datetime schemas via the API (currently these fields are rejected with validation errors)

## Requirements
- Allow devs to send API requests that include dateTime fields

## Acceptance Criteria
- Ideally, API would interpret a datetime string in the same way javascript's `new Date(string)` does.

tefkah:
- **fix: throw clearer errors on validation error**
- **fix: allow you to use dates in api**

## Issue(s) Resolved

Resolves #837

When passing pubfields through `validatePubFields`, i add a manual check for date fields and hydrate them.

I also made validation errors slightly more clear.

## Test Plan

1. Send an api request like
```
POST /api/v0/c/arcadia-research/site/pubs
PREFER return=representation
{ 
   "pubTypeId": <find a pubtype>,
  "values": { "arcadia-research:created-at": "2012-07-09" }
}
```

You should be able to do it!

Ideally also test for `PATCH /pubs` and `PUT/PATCH pubs/relations`, but they all use the same function so should work.

### DIFF
--- a/core/lib/server/pub.ts
+++ b/core/lib/server/pub.ts
@@ -42,7 +42,7 @@ import { parseRichTextForPubFieldsAndRelatedPubs } from "../fields/richText";
 import { mergeSlugsWithFields } from "../fields/utils";
 import { autoCache } from "./cache/autoCache";
 import { autoRevalidate } from "./cache/autoRevalidate";
-import { NotFoundError } from "./errors";
+import { BadRequestError, NotFoundError } from "./errors";
 import { getPubFields } from "./pubFields";
 import { getPubTypeBase } from "./pubtype";
 import { SAFE_USER_SELECT } from "./user";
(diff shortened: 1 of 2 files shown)


allisonking (approved):
works great! it looks like I still only get an internal server error response if a field does not exist, but that's probably not related to this PR, I just happened on it while testing this

```
[13:19:47.919] ERROR: Pub values contain fields that do not exist in the community: arcadia-research:created-at
    package: "core"
    err: {
      "type": "Error",
      "message": "Pub values contain fields that do not exist in the community: arcadia-research:created-at",
```

tefkah:
Ah yeah that's still and issue, but that's something to fix later!

PR #849, opened 2024-12-17
fix: use fieldname as default label for form elements

tefkah:
- **fix: add better typesafety for form elements and properly display the name of the field as the default label for form elments**
- **fix: more useful logic for labels and slugs**
- **fix: make fieldcreation wait for success**

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,5 +2,6 @@
 	"files.associations": {
 		"*.css": "tailwindcss"
 	},
-	"typescript.tsdk": "./core/node_modules/typescript/lib"
+	"typescript.tsdk": "./core/node_modules/typescript/lib",
+	"javascript.preferences.autoImportSpecifierExcludeRegexes": ["@prisma\\/client"]
 }
--- a/playwright/.auth/user.json
+++ b/playwright/.auth/user.json
@@ -1,0 +1,25 @@
+{
+  "cookies": [
+    {
+      "name": "auth_session",
+      "value": "gypreicrdhiribgeoojxyfjwtvkafmaq556yhakj",
+      "domain": "localhost",
+      "path": "/",
+      "expires": 1769018339.839587,
+      "httpOnly": true,
+      "secure": false,
+      "sameSite": "Lax"
+    },
+    {
+      "name": "pubpub_community_slug",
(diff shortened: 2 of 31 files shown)


kalilsn:
You may have noticed this already and I've only skimmed this PR but I wanted to mention that the reason these labels aren't set is that I messed up the `insertForm` query and it's putting labels in `form_elements.label`, instead of the config because I forgot about [this comment](https://github.com/pubpub/platform/blob/62445145445ccbb0f3c18b2d64c8db2a37346519/core/prisma/schema/schema.prisma#L566).

Feel free to resolve that however you like. I kind of regret moving the label to the config just for checkboxes and I wonder if there's a better way to handle that and leave the labels in the label column like you've started doing here.

tefkah:
Huh, I didn't know it was bc of that, I though it was bc the `label` was set to the fieldName basically, and `config.label` was the "override" label. One reason i thought that is bc checkboxes indeed also have `label` set, but it's not used in their configuration. I added `config.label` as a config option to checkboxes (mostly to make the types nicer), but maybe i should remove that?

I'm not sure what the best way to resolve this is. I'm fine with leaving it like this for now due to the time pressure, but maybe there's a better way?

playwright/.auth/user.json
```diff
@@ -0,0 +1,25 @@
+{
```
allisonking, line 1: hm I thought this was usually gitignore'd?

kalilsn (approved):
Works well and I think this is at least fine for now! I'm sure we'll revisit this at some point next year!

.vscode/settings.json
```diff
@@ -2,5 +2,6 @@
 	"files.associations": {
 		"*.css": "tailwindcss"
 	},
-	"typescript.tsdk": "./core/node_modules/typescript/lib"
+	"typescript.tsdk": "./core/node_modules/typescript/lib",
+	"javascript.preferences.autoImportSpecifierExcludeRegexes": ["@prisma\\/client"]
```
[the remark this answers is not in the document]
tefkah: (i'm not completely sure this works)

playwright/.auth/user.json
tefkah: huh yeah, i will remove it

Issue #842, opened 2024-12-16
Prevent users invited to edit a pub from editing other pubs


kalilsn:
## Motivation
So users cannot edit pubs they don't have explicit permission to edit

## Requirements
- Users invited to fill out a form for a specific pub cannot edit other pubs by changing the `pubId` param in the URL
- Users can still create new pubs via the form

## Acceptance Criteria
- Add a nullable pubId column to the form_memberships table
- Update the function `addMemberToForm` to accept a pubId, and use this when inserting a form membership
- Update `userHasPermissionToForm` to filter by a pubId as well, when passed
- Make sure to pass the pubId to `userHasPermissionToForm` on pages where it's present

PR #850, opened 2024-12-17
Add pubId to form memberships

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/842

Prevent users invited to edit a pub from editing other pubs (previously could do this by swapping out the `pubId` parameter. This adds a `pubId` to the form memberships table so that invited users are scoped to a pubId.

## Test Plan
* Go to a pub in CrocCroc in the Submitted stage and run the action "Send Review email"
* Send the email to a member who is only a contributor (i.e. `none@pubpub.org`)
* Visit inbucket http://localhost:54324/monitor and find the email that was just sent
* View the email in 'plain text' form and copy the link 
* Open the link in an incognito tab (or any browser that isn't signed in)
* You should see the form
* Now copy the link, but swap out the `pubId` value. i.e. swap out the `[pubId]` portion here `http://localhost:3000/magic-link?token=[token]&redirectTo=%2Fc%2Fcroccroc%2Fpublic%2Fforms%2Freview%2Ffill%3FpubId%3D[pubId]`
* Paste that link into another place you aren't signed in and you should see a 404 page
* Now instead of replacing with another `pubId`, just remove the pubId, i.e. `http://localhost:3000/magic-link?token=[token]&redirectTo=%2Fc%2Fcroccroc%2Fpublic%2Fforms%2Freview%2Ffill%3F` 
* You should be able to access the create page and be able to create a pub still


core/app/components/pubs/PubEditor/actions.ts
```diff
@@ -16,30 +16,45 @@ import { ApiError, createPubRecursiveNew } from "~/lib/server";
...
+		const createdPub = await createPubRecursiveNew(createPubProps);
+
+		if (addUserToForm && formSlug) {
+			await addMemberToForm({
+				communityId: props.communityId,
+				userId: user.id,
+				slug: formSlug,
+				pubId: createdPub.id,
+			});
+		}
```
allisonking, line 57: I'm not 100% sure this is what we want but I put it in for the case where:

* A user is invited to create a pub via `/fill` (no pub id)
* They fill out the form
* Without this, after submission, the page 404s because the user doesn't actually have access to that pub id 

I figured this might be useful so that the user could also feasibly go back and edit a pub they created, but looking for feedback here! A drawback might be that this might also trigger if the user is an admin/editor, they'd also get added to the form though they don't necessarily need to be since they already have create/edit privileges to all pubs. Could get around this by checking if they already have a greater permission, though I wasn't sure if that was worth it?

core/prisma/schema.dbml
```diff
@@ -281,12 +282,14 @@ Table form_memberships {
...
+  pub pubs
+  pubId String
   createdAt DateTime [default: `now()`, not null]
   updatedAt DateTime [default: `now()`, not null]
 
   indexes {
-    (formId, userId) [unique]
-    (formId, memberGroupId) [unique]
+    (formId, userId, pubId) [unique]
+    (formId, memberGroupId, pubId) [unique]
```
allisonking, line 292: I _think_ we need to expand the unique constraint here, but not sure if there is a case where this might mess things up

kalilsn: That seems like a pretty low priority problem to me though, since there's not really any problem with having extra form memberships. We can always write a migration later that cleans up the duplicates and adds more restrictive constraints.

core/app/components/pubs/PubEditor/actions.ts
3mcd: I think this is a good solution to the case you outlined. I'm not too worried about it creating redundant permissions. It may even be desireable in some edge cases (i.e. a user has their editor permission revoked and suddenly can't use a form they were invited to).

core/prisma/schema.dbml
allisonking: ah okay! good catch! since it's low priority, I'm in favor of punting and maybe by then we'll have upgraded our postgres to make distinct nulls easier :)

allisonking:
To test the backfill migration:

1. Hop over to main and invite contributor users to review a pub 
2. Switch to this branch
3. Before migrating, with only code changes, if you click on the invite link one of the contributors received in inbucket, and you open that link in an incognito tab, you'll get a 404
4. Now run the migration `pnpm run migrate-dev`
5. Open the link in incognito again (you may need to hard refresh too). the page should now appear!

3mcd (approved):
Nice job!

PR #851, opened 2024-12-18
Limit pub visibility

kalilsn:
…update types to be slightly more flexible## Issue(s) Resolved


core/lib/server/pub.ts
```diff
@@ -1398,6 +1480,113 @@ export async function getPubsWithRelatedValuesAndChildren<
...
+							eb.val(props.communityId).as("membId"),
+							sql<MembershipType>`'community'::"MembershipType"`.as("type"),
+						])
+					)
+
+					.$castTo<{
+						role: MembershipCapabilitiesRole;
+						membId: string;
+						type: MembershipType;
+					}>()
```
kalilsn, line 1545: These three CTEs all use this cast so they can return rows with the same schema which allows us to combine the rows with a `union` in the next CTE.

core/lib/server/pub.ts
```diff
@@ -1398,6 +1480,113 @@ export async function getPubsWithRelatedValuesAndChildren<
...
+						qb
+							.where("community_memberships.userId", "=", props.userId!)
+							.where("community_memberships.communityId", "=", props.communityId)
+							.select([
+								"role",
+								"communityId as membId",
+								sql<MembershipType>`'community'::"MembershipType"`.as("type"),
+							])
+					)
+					// Add a fake community admin role when selecting without a userId
```
kalilsn, line 1532: It's pretty hard to conditionally add a CTE so I ended up with all these $if statements to handle when a userId isn't passed. And looking at this now, I realize it would probably improve performance to use a $if on the `where` condition that actually enforces the authorization check instead of inserting this dummy row.

core/lib/server/pub.ts
```diff
@@ -1354,6 +1374,68 @@ export async function getPubsWithRelatedValuesAndChildren<
 									)
 								)
 								.leftJoin("PubsInStages", "pubs.id", "PubsInStages.pubId")
+								.where((eb) =>
+									eb.exists(
+										eb.selectFrom("capabilities" as any).where((ebb) => {
```
kalilsn, line 1379: kysely/typescript don't like us selecting from the later CTE so "we" (thomas) cast it to any and then use some magic to create an ExpressionBuilder with the right type parameter (all the tables/aliases in scope) so that kysely doesn't complain

core/lib/server/pub.ts
```diff
@@ -1354,6 +1374,68 @@ export async function getPubsWithRelatedValuesAndChildren<
...
+													capabilities: {
+														membId: string;
+														type: MembershipType;
+														role: MembershipCapabilitiesRole;
+													};
+												},
+												any
+											>;
+
+											return eb.or([
```
kalilsn, line 1398: this big condition is repeated here in the recursive query, and initially in the root pubs selection. and what it does is just check if the user has the capability viewPub on each pub or viewStage on that pub's stage. we need this big set of checks to make sure that only the capabilities that are relevant for any given pub are used, since the "capabilities" cte has all of the user's pub and stage capabilities, not filtered by any pub or stage.

core/lib/server/pub.ts
```diff
@@ -1398,6 +1480,113 @@ export async function getPubsWithRelatedValuesAndChildren<
 						)
 					)
 			)
+			.with("stage_ms", (db) =>
```
kalilsn, line 1483: in these CTEs, we select all the user's stage memberships, all their pub memberships, and any community memberships for the particular community that this query is running in. unlike the query in `lib/capabilities` we don't filter the memberships by anything except the user in these first two CTEs because we don't yet know which pubs or stages are relevant.

core/lib/server/pub.ts
```diff
@@ -1398,6 +1480,113 @@ export async function getPubsWithRelatedValuesAndChildren<
...
+						join.on((eb) =>
+							eb.and([
+								eb("membership_capabilities.role", "=", eb.ref("memberships.role")),
+								eb("membership_capabilities.type", "=", eb.ref("memberships.type")),
+							])
+						)
+					)
+					.where("membership_capabilities.capability", "in", [
+						Capabilities.viewPub,
+						Capabilities.viewStage,
```
kalilsn, line 1586: these are the two kinds of capability that can allow a user to view a pub

core/lib/server/pub.ts
kalilsn: So `membId` can be a stage id, pub id, or community id, and `type` tells us which

core/lib/server/pub.ts
```diff
@@ -1407,11 +1596,37 @@ export async function getPubsWithRelatedValuesAndChildren<
 					.selectFrom("pubs")
 					.selectAll("pubs")
 					.where("pubs.communityId", "=", props.communityId)
+					.leftJoin("PubsInStages", "pubs.id", "PubsInStages.pubId")
+					.select("PubsInStages.stageId")
+					.where((eb) =>
```
kalilsn, line 1601: this condition should be the same as the `where` above in the recursive part of the query

core/app/c/[communitySlug]/pubs/[pubId]/page.tsx
```diff
@@ -127,6 +126,19 @@ export default async function Page({
...
 	);
 
+	const canView =
+		(await userCan(Capabilities.viewPub, { type: MembershipType.pub, pubId }, user.id)) ||
+		(pub.stageId &&
+			(await userCan(
+				Capabilities.viewStage,
+				{ type: MembershipType.stage, stageId: pub.stageId },
+				user.id
+			)));
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:136]: Unfortunately we have to do this check _after_ fetching the pub, because we need its stage too

Comment (tefkah) [reply core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:136]: ah right, that's very annoying. But can't we skip the `viewPub` check, as that's done in `getPubs` anyway? I guess it then becomes hard to say whether or not we don't have a pub bc you aren't allowed to view it or bc the pub does not exist.

Comment (tefkah) [reply core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:136]: you could theoretically also move the Stage display to a separate component that's `<Suspended>` and only do the stage check there

but actually, i think we should just let you view the name of the stage on this page, since stage names are kind of like statuses.

Comment (tefkah) [reply core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:136]: if you're a contributor, i think you should be able to see the pub page of your article and which stage it's in, even if you cannot see the stage completely

core/lib/server/pub-capabilities.db.test.ts
```diff
@@ -0,0 +1,150 @@
...
+						value: "test relation value",
+						pub: {
+							pubType: "Basic Pub",
+							values: {
+								Title: "A pub related to another Pub",
+							},
+						},
+					},
+				],
+			},
```
tefkah, line 84: i know you didn't write these tests per se, but i think we should add tests for pub members as well no?

core/app/c/[communitySlug]/pubs/[pubId]/page.tsx
Comment (kalilsn) [reply core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:136]: i suppose you're right about skipping the viewPub check. i'm guessing it throws an error when no pub is returned anyways.

i'm not sure i follow about the stage check though. i'm not checking that capability here for displaying the name of the stage. i'm checking it and combining here with `||` because having `viewStage` is sufficient for viewing the pub even if you don't have `viewPub` directly on the pub

Comment (kalilsn) [reply core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:136]: and now that i've said that, i'm remembering that i specifically implemented that logic inside of usercan 🤦

core/lib/server/pub.ts
```diff
@@ -1398,6 +1480,113 @@ export async function getPubsWithRelatedValuesAndChildren<
...
+										),
+										eb.val(props.communityId).as("membId"),
+										sql<MembershipType>`'community'::"MembershipType"`.as(
+											"type"
+										),
+									])
+							)
+					)
+			)
+			.with("capabilities", (cte) =>
```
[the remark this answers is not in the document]
allisonking: wow this is very involved, thank you for the PR comment walkthrough!

I think I need to do someting similar for stages, here's what I had so far... https://github.com/pubpub/platform/pull/856 I was getting kind of stuck though, so I'll be referring to this! but also if you have any tips having just gone through this, I'd love them 😅

core/app/c/[communitySlug]/pubs/[pubId]/page.tsx
Comment (kalilsn) [reply core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:136]: so i may not need either of these really, but i will put the viewPub one before the pub fetching

Comment (tefkah) [reply core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:136]: ah nevermind, that makes sense!

but yeah, i think still having the `userCan` check to be able to redirect users is better than just showing "Pub Not Found"

core/lib/server/pub-capabilities.db.test.ts
allisonking: oh right I missed that! I can add that

core/lib/server/pub.ts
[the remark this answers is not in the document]
kalilsn: it looks like you're pretty close and a similar but simpler approach should work for you! if you take the queries you've written and combine them into one or three CTEs in the `getStages` query, you should be able to `.innerJoin` your last cte (the deduplicated list of all authorized stageIds) to the main query and that should filter it. since you're writing my tests for me, i'd be happy to try that for you if you want!

core/lib/server/pub-capabilities.db.test.ts
allisonking: https://github.com/pubpub/platform/pull/858

caught a very small bug in the seed script :)

core/lib/server/pub.ts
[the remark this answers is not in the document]
allisonking: ooh okay lemme try that out first and I'll ping you if I run into problems! thanks for the guidance!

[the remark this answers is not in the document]
allisonking: whew your instructions worked perfectly! much simpler than what you had to do here 😅

allisonking (approved):
🙌

PR #852, opened 2024-12-18
fix: allow adding relation fields to pubtype

tefkah:
- **fix: fetch relations when trying to add pubfields**
- **fix: add pubtype page tests**
- **fix: add test for checking whether yuo can designate field as title**

## Issue(s) Resolved
Allow users to add relation field types to pub type

We just weren't fetching the relationships on the PubTypes page, so i made it do that.

I also added some specific tests for pubtype creation.

In addition: you are not able to set relationship fields as the title. we should probably prevent this somewhere else too, but i'm too lazy to figure that out rn.

## Test Plan

1. Create a relation field of SchemaType string
2. Create a pub type, add that relationship field to it.
3. Observe you cannot set it as the title
4. Tada

### DIFF
--- a/core/app/c/[communitySlug]/types/page.tsx
+++ b/core/app/c/[communitySlug]/types/page.tsx
@@ -57,7 +57,6 @@ export default async function Page({
 	if (!types || !fields) {
 		return null;
 	}
-	console.log(fields);
 	return (
 		<PubFieldProvider pubFields={fields}>
 			<div className="mb-16 flex items-center justify-between">
--- a/core/app/c/[communitySlug]/types/page.tsx
+++ b/core/app/c/[communitySlug]/types/page.tsx
@@ -57,6 +57,7 @@ export default async function Page({
 	if (!types || !fields) {
 		return null;
 	}
+	console.log(fields);
 	return (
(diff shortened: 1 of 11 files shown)


core/app/c/[communitySlug]/types/page.tsx
```diff
@@ -48,7 +48,10 @@ export default async function Page({
 
 	const [types, { fields }] = await Promise.all([
 		getAllPubTypesForCommunity(communitySlug).execute(),
-		getPubFields({ communityId: community.id }).executeTakeFirstOrThrow(),
+		getPubFields({
+			communityId: community.id,
+			includeRelations: true,
```
Comment (allisonking) [inline core/app/c/[communitySlug]/types/page.tsx:53]: I had added `includeRelations` while the relationship stuff was still being ironed out and we didn't want it showing everywhere. if we are always showing relations for every query now, then we might as well remove the lil bit of code that `includeRelations` triggers and always return as default!

I just looked through the code base and there are a few places we aren't passing `includeRelations` though I don't know if this is intentional. anyway, not important for this PR but for the future it's a potential thing to clean up!

allisonking (approved):
thanks for adding the pub type page tests! it looks like they're failing in CI, but I think this is good to go once that's fixed!

PR #853, opened 2024-12-18
hotfix: allow deleting empty pubvalues

tefkah:
- **fix: make arcadiaseed accesstoken correct**
- **fix: do not error if no pubvalues are deleted**

## Issue(s) Resolved
`PUT pubs/relations` route would throw if no values existed beforehand, bc it would try remove pubvalues and then manually insert them into `pub_value_history`. there were no safeguards if there were no pubvalues to insert though, so it would try to `db.insertInto('pub_values_history').values([])`, which errors.

## Test Plan
1. Run test
2. Ideally try to `PUT` a relation but this has some haste, so just wing it.

### DIFF
--- a/core/prisma/exampleCommunitySeeds/arcadia.ts
+++ b/core/prisma/exampleCommunitySeeds/arcadia.ts
@@ -693,7 +693,7 @@ export const seedArcadia = (communityId?: CommunitiesId) => {
 		},
 		{
 			randomSlug: false,
-			withApiToken: "xxxxxxxxxxxxxxxx.00000000-0000-0000-0000-000000000000",
+			withApiToken: "00000000-0000-0000-0000-000000000000.xxxxxxxxxxxxxxxx.",
 			parallelPubs: true,
 		}
 	);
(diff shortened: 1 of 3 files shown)


core/prisma/exampleCommunitySeeds/arcadia.ts
```diff
@@ -693,7 +693,7 @@ export const seedArcadia = (communityId?: CommunitiesId) => {
 		},
 		{
 			randomSlug: false,
-			withApiToken: "xxxxxxxxxxxxxxxx.00000000-0000-0000-0000-000000000000",
+			withApiToken: "00000000-0000-0000-0000-000000000000.xxxxxxxxxxxxxxxx.",
```
allisonking, line 696: just want to confirm we do want the `.` at the end?

allisonking (approved):
tested locally and got no errors!

core/prisma/exampleCommunitySeeds/arcadia.ts
tefkah: whoops, we don't, but i think it's fine

PR #855, opened 2024-12-19
fix: do not set title to null if a non-title value is updated

tefkah:
- **fix: add working test that will show the unwanted behavior**
- **fix: correctly update the title if a non-pubtitle is updated**
- **fix: create a new migration instead of updating old one**
- **fix: update existing test instead of adding new test**

## Issue(s) Resolved
When updating a non-title value on a pub, it's `title` would get set to null!
@3mcd  identified this as a problem while working on #833 

I missed this because a test was not behaving properly for some reason.

Basically, because i wanted to be efficient and update the `updatedAt` and `title` on query, I used a `LEFT JOIN` here in the update query (bc i wanted to update the `updatedAt` even if no title update was necessary)

https://github.com/pubpub/platform/blob/a57197798f56fc5af66b6181af8fe6f9a2f803ca/core/prisma/migrations/20241205152006_add_title_field/migration.sql#L56-L74

This caused a problem though, as that meant that this `ELSE` part of the update query

https://github.com/pubpub/platform/blob/a57197798f56fc5af66b6181af8fe6f9a2f803ca/core/prisma/migrations/20241205152006_add_title_field/migration.sql#L56-L65

would always be `null` if no title value needed to be updated!

This PR fixes that by adding this additional case that checks whether the `LEFT JOIN` even did anything (it will be null if no columns have been joined)

https://github.com/pubpub/platform/blob/a57197798f56fc5af66b6181af8fe6f9a2f803ca/core/prisma/migrations/20241219131700_fix_title_update_trigger/migration.sql#L51-L53

This way I can still update things in one query.

Maybe it's a better idea to split this query up into a bunch of CTEs anyway, that way it'll be much more legible, rather than this weird "create a temp table and then use on update query" strat i went with, but this was the most minimal change to the existing function i could think of.

## Test Plan

1. On `main`, open prisma studio with `pnpm --filter core db:studio`
2. Find some pub in the db with a non-null `title` field, and update a non-title value of that pub in the UI
3. Click the lil refresh thing in the topleft.
4. Observe that the `title` field is not `null`

----

1. First, on `main`, add this test somewhere to `pub-trigger.db.test.ts`

```ts
	it("should not update a title on a pub when a non-title pubvalue is updated", async () => {
		// we explictly don't want to use the rolled back transaction here
		// for some, strange strange reason, this doesn't work then
		// with the transaction rolled back
		const trx = testDb;
		const { seedCommunity } = await import("~/prisma/seed/seedCommunity");
		const { pubFields, pubs } = await seedCommunity(pubTriggerTestSeed, undefined, trx);

		expect(await getPubTitle(pubs[0].id, trx)).toBe("Some title");

		await trx
			.insertInto("pub_values")
			.values({
				pubId: pubs[0].id,
				fieldId: pubFields.Description.id,
				value: JSON.stringify("new description"),
				lastModifiedBy: createLastModifiedBy("system"),
			})
			.onConflict((oc) =>
				oc
					.columns(["pubId", "fieldId"])
					.where("relatedPubId", "is", null)
					.doUpdateSet((eb) => ({
						value: eb.ref("excluded.value"),
						lastModifiedBy: createLastModifiedBy("system"),
					}))
			)
			.executeTakeFirstOrThrow();

		expect(await getPubTitle(pubs[0].id, trx)).toBe("Some title");
	});
```
2. Convince yourself that this test accurately depicts a scenario where we update a pubvalue.
3. Run test, observe that it properly fails.

----

1. Switch to this branch
2. Run `pnpm --filter core migrate-dev`. Observe that this does not trigger a reset (important, otherwise it won't apply cleanly on main)
3. Run `pnpm --filter core vitest lib/server/pub-trigger.db.test.ts` (the test above has been added)
4. Observe that the test called `should not update a title on a pub when a non-title pubvalue is updated` passes

----

1. Open prisma studio with `pnpm --filter core db:studio`
2. Find some pub in the db with a non-null `title` field, and update a non-title value of that pub in the UI
3. Click the lil refresh thing in the topleft.
4. Observe that the `title` field is not `null`

### DIFF
--- a/core/lib/server/pub-trigger.db.test.ts
+++ b/core/lib/server/pub-trigger.db.test.ts
@@ -305,44 +305,9 @@ describe("pub_values title trigger", () => {
 	});
 
 	it("should not update a title on a pub when a non-title pubvalue is updated", async () => {
-		const trx = getTrx();
-		const { seedCommunity } = await import("~/prisma/seed/seedCommunity");
-		const { pubFields, pubs } = await seedCommunity(pubTriggerTestSeed);
-
-		expect(await getPubTitle(pubs[0].id, trx)).toBe("Some title");
-
-		const updateResult = await trx
-			.updateTable("pub_values")
-			.set({
-				value: JSON.stringify("new description"),
-				lastModifiedBy: createLastModifiedBy("system"),
-			})
-			.where("pubId", "=", pubs[0].id)
-			.where("fieldId", "=", pubFields.Description.id)
-			.executeTakeFirst();
-
-		expect(await getPubTitle(pubs[0].id, trx)).toBe("Some title");
-	});
-
-	it("should delete a title on a pub when a pubvalue is deleted", async () => {
-		const trx = getTrx();
-		const { seedCommunity } = await import("~/prisma/seed/seedCommunity");
-		const { pubFields, pubs } = await seedCommunity(pubTriggerTestSeed);
-
-		expect(await getPubTitle(pubs[0].id, trx)).toBe("Some title");
-
-		const deletedPubValue = await trx
-			.deleteFrom("pub_values")
-			.where("pubId", "=", pubs[0].id)
-			.where("fieldId", "=", pubFields.Title.id)
-			.returningAll()
-			.executeTakeFirstOrThrow();
-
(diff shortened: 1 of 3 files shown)


3mcd (approved):
That was fast! Tested it and worked great

Issue #762, opened 2024-11-07
Restrict stage visibility based on membership


3mcd:
## Motivation

So users only see stages they have permission to view.

## Requirements

Modify stage-related queries so that:
- Community contributors can only see stages they have been added to directly or via groups.
- Community editors and admins can always see all stages.

If a user does not have the ability to see any stages, the link to "Workflows" in the side nav should not appear.

## Acceptance Criteria

- Uses functions implemented in #671 to determine if the user has the `viewStage` capability for each rendered stage.
- End-to-end tests that assert that stages on the workflows page are only visible when the above requirements are met.

3mcd:
@kalilsn I agree that the requirement to show/hide the Workflows link adds a bit of overhead (one query per page load) for little gain. Instead I think we could always show the link and just put an informative message on the Workflows page if no stages are accessible to the user. Thoughts @gabestein ?

allisonking:
this would involve the following, though I'd love confirmation that this sounds on the right track:

* modifying this query to filter by only stages the user has the `viewStage` capability for https://github.com/pubpub/platform/blob/76f44f265abffce1430511a35cec493dec17cc79/core/lib/server/stages.ts#L52-L103
* if the query returns an empty list in the Workflows page here, then render a message that there are no accessible stages https://github.com/pubpub/platform/blob/76f44f265abffce1430511a35cec493dec17cc79/core/app/c/%5BcommunitySlug%5D/stages/manage/page.tsx#L69
* will returning only some stages mess with the graph layout? maybe it'll just work?

gabestein:
> will returning only some stages mess with the graph layout? maybe it'll just work

I think this is an editor- and admin-only view anyway, so we can probably just not render it if you don't have access to view all stages?

PR #856, opened 2024-12-19
Restrict stage visibility

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/762

Adds a subquery for filtering stage queries to just the ones the user has permissions for. This means the stage queries now require a `userId`, so this also makes sure that we pass `userId`s throughout where we are querying for stages.

## Test Plan

(some comments surrounded by ❓ I am looking for feedback on!)

* Add a contributor member
* Add an entry in the StageMemberships table to give this member admin privileges to a stage
* Sign in as this new contributor
* Comment out these lines so that the contributor member can see the stage page (these lines will eventually be removed when everything is properly restricted). ❓ should this be removed now, or are there other things we are waiting for ❓ 
https://github.com/pubpub/platform/blob/5b8586d5a2a377593cbbb27963399f29aea7cb30/core/app/c/%5BcommunitySlug%5D/layout.tsx#L38-L42
* Using the stageId the contributor got assigned to , visit `http://localhost:3000/c/{community}/stages/{stageId}`. You should see the page and any pubs in that stage. Other stageIds should still 404

Note: the stage won't show up in `http://localhost:3000/c/{community}/stages` because of
https://github.com/pubpub/platform/blob/5b8586d5a2a377593cbbb27963399f29aea7cb30/core/lib/stages.ts#L54-L63

This returns an empty array since the stage still has a `moveConstraintSource`. ❓ Should I rework this logic to still show the stage even though its source doesn't "exist"? Or should the `moveConstraintSource` not be returned since the user doesn't have access to that stage? ❓


core/lib/server/stages.ts
```diff
@@ -48,9 +51,54 @@ export const getPubIdsInStage = (stageId: StagesId) =>
...
+				.onRef("membership_capabilities.role", "=", "community_memberships.role")
+				.on("membership_capabilities.type", "=", MembershipType.community)
+		)
+		.innerJoin("stages", "stages.communityId", "community_memberships.communityId")
+		.where("community_memberships.userId", "=", userId)
+		.where("community_memberships.communityId", "=", communityId)
+		.where("membership_capabilities.capability", "=", Capabilities.viewStage)
+		.select(["stages.id as stageId"]);
+
+	const viewableStages = db
```
allisonking, line 85: need to somehow combine this query with `getStages` below...

allisonking:
I think this is mostly there but I haven't tested thoroughly yet. wouldn't be surprised if something sneaky came up! Importantly, going to wait to update the API until https://github.com/pubpub/platform/pull/851 is in as it has some logic I also need to get the user from a token.

tefkah:
Just tested it, works well! I do think we should rework that stage source logic, and the stage should show up on `/stages`. 
Ideally, we should only show the "move targets" on the move dropdown that the user has access to, but i'm not sure if our capability checks are specific enough for that yet
There's only a "move pub" capablity, not a "canMovePubFrom/ToStage" capability, which seem somewhat different?
I'll think about this a bit, maybe @gabestein you have some more concrete ideas?

3mcd:
@tefkah I believe it would just be a combination of two capabilities. We have a couple examples of these aggregate capabilities that we should probably better define somewhere.

allisonking:
I tried two ways—here's the first, where we keep the move constraints but I tweaked the code so that the stage shows up on the page
image

This allows the user to move pubs to other stages still—this is possible because, since the pub is in the stage the user has perms for, they also get the 'movePub' capability. Once they move the pub out of the stage, they can't move it back in though. This feels like a use case a stage admin might want, to move their stage to the next stage (though it might be a stage they aren't responsible for). This is implemented in https://github.com/pubpub/platform/pull/856/commits/6648d6e77f82f34edee4526b4a5ddbeb609f372b

The other option is to tweak the `moveConstraints` and `moveConstraintSources` of the `stage` value when we are building out the workflow so that these only show stage Ids the user has access to. This makes it so that there isn't a Move button at all. This is implemented in https://github.com/pubpub/platform/pull/856/commits/d9863812a90ac887085c08edad48f4fa9fb8d478
image

I think the way I implemented it is a little hacky too, and if we want to do this we should filter the query for move constraints at the db query level, which is more complex

core/lib/stages.ts
```diff
@@ -22,6 +22,12 @@ function createStageList<T extends CommunityStage>(
 	stages: StagesById,
 	visited: Array<T> = []
 ): Array<T> {
+	// If stage is undefined, return the current visited list
+	// This would happen if the move constraint isn't in the stage list (because of user permissions)
+	if (!stage) {
+		return visited;
+	}
+
```
allisonking, line 30: I'm not 100% sure about this logic—it works, though I was having a hard time wrapping my head around the recursiveness, so could use a second opinion! basically without this, `stage` can be `undefined` since the move constraint stage might not exist in `stagesById`. We still want to show the stage though, so I'm returning the `visited` list which should have the current stage

3mcd (approved):

PR #857, opened 2024-12-19
Pub authorization tests

allisonking:

### DIFF
--- a/core/lib/server/pub-capabilities.db.test.ts
+++ b/core/lib/server/pub-capabilities.db.test.ts
@@ -1,13 +1,11 @@
-import { beforeAll, describe, expect, expectTypeOf, it } from "vitest";
+import { describe, expect, it } from "vitest";
 
 import { CoreSchemaType, MemberRole } from "db/public";
 
 import type { Seed } from "~/prisma/seed/seedCommunity";
 import { mockServerCode } from "../__tests__/utils";
 
-const { createForEachMockedTransaction } = await mockServerCode();
-
-const { getTrx } = createForEachMockedTransaction();
+await mockServerCode();
 
 const seed = {
 	community: {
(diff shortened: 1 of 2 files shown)


core/lib/server/pub-capabilities.db.test.ts
```diff
@@ -0,0 +1,150 @@
+import { describe, expect, it } from "vitest";
```
allisonking, line 1: I pulled this into its own file because `pub.db.test.ts` has an issue of its transactions not working properly, I think because we aren't dynamically importing `seedCommunity`. this meant that I'd query for pubs only to get pubs created by other tests in the file. so this separates it out, but we should fix `pub.db.test.ts` so we can run its tests without the tests bumping into each other

core/prisma/seed/seedCommunity.ts
```diff
@@ -1196,3 +1196,5 @@ export const createSeed = <
 	pubs?: PI;
 	forms?: F;
 }) => props;
+
+export type Seed = Parameters<typeof createSeed>[0];
```
allisonking, line 1200: we can import types without problems and our transactions will work. importing `createSeed` the function, without doing a dynamic import in the test itself, seems to cause the data to persist though

tefkah: smart!

kalilsn (approved):
looks like exactly what we needed and i'm relieved it seems to just work!

core/lib/server/pub-capabilities.db.test.ts
kalilsn: separating it seems good for organization anyways!

tefkah:
This looks great, but, if it's not too much trouble, could you add a number of checks for pub memberships as well? unless i missed some that kalil already wrote.

PR #858, opened 2024-12-19
Pub membership test

allisonking:

### DIFF
--- a/core/prisma/seed/seedCommunity.ts
+++ b/core/prisma/seed/seedCommunity.ts
@@ -346,7 +346,7 @@ const makePubInitializerMatchCreatePubRecursiveInput = <
 
 		const members = Object.fromEntries(
 			Object.entries(pub.members ?? {}).map(
-				([slug, role]) => [users.find((user) => user.slug === slug)?.id!, role!] as const
+				([slug, role]) => [findBySlug(users, slug)?.id!, role!] as const
 			)
 		) as Record<UsersId, MemberRole>;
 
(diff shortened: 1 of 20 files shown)


core/prisma/seed/seedCommunity.ts
```diff
@@ -346,7 +346,7 @@ const makePubInitializerMatchCreatePubRecursiveInput = <
 
 		const members = Object.fromEntries(
 			Object.entries(pub.members ?? {}).map(
-				([slug, role]) => [users.find((user) => user.slug === slug)?.id!, role!] as const
+				([slug, role]) => [findBySlug(users, slug)?.id!, role!] as const
```
kalilsn, line 349: not important but i'm curious: why does this work differently? do the slugs not exactly `===` each other (i saw the regex in findBySlug)?

kalilsn (approved):
Perfect!

core/prisma/seed/seedCommunity.ts
allisonking: yes, by default the seed script appends a date to each user slug so that we get unique entries (I didn't know this until just now either!) looks like we can also pass `randomSlug = false` to not get that behavior though

https://github.com/pubpub/platform/blob/9b6abee86945f756457acfbac40d6fb93fedf145/core/prisma/seed/seedCommunity.ts#L826

kalilsn: ah that makes sense!

PR #859, opened 2024-12-19
Initial Test of Google Action

isTravis:
This is just an initial pass for the Google Drive action. It has most scaffolding and importing complete. Pushing it up as a test for merging and deploying.


tefkah:
you should probably disable this test for now, otherwise CI will probably never pass!

https://github.com/pubpub/platform/blob/76f44f265abffce1430511a35cec493dec17cc79/core/app/components/pubs/PubEditor/actions.db.test.ts#L95-L175

3mcd (approved):
🥳

PR #863, opened 2024-12-20
use string or date as publication date in datacite action

3mcd:
There is still a discrepancy between dev and prod. Pub values with a `DateTime` schema are `Date`s in prod and strings in dev. This is causing the DataCite action to fail with the error `"The pub is missing a value corresponding to the configured publication date field override."`. This is a temporary fix to get the action working.


tefkah (approved):
I did not test anything but this seems like it would be introduced by hydrating the values first, and this fix seems to address that

PR #867, opened 2024-12-21
Recursively render related pub values

### LINKED ISSUE #841: Render related pub values on the pubs page
## Motivation

So that users authorized to view an individual pub through a stage or pub membership can also see its values. This is especially useful when important information like the author of a pub is stored in a relationship.

## Requirements
- On the pub page (/c/commnity_slug/pubs/pubId), render values from pubs referenced in relationship fields
- These rendered values from related pubs should also render that pub's relationship values, but only to depth 2 (once)
- If the related pub is related back to the initial pub, it should not render the initial pub's values again

kalilsn:
## Issue(s) Resolved
Resolves #841 

I've added related pub values up to two pubs deep to the pubs page, and made a few style changes to it as well since there's way more content now.

- Related pubs should now render their values inline, with a link to the pub
	- 	If the depth is more than 2, we won't render the values, but there's always a link to the deepest pub we do render, so a user can click on it to see its values
	- If the related pub is the current pub, we don't render a link or a value, but add a badge saying "Current pub"
- Related pub values are collapsed by default, because the pages can be very long otherwise
- Relationship field values are now grouped under a single heading per field, rather than one heading per value
- When relationship values have a value in addition to the pub, we render it before the pub link with a `:` at the end 

## Test Plan

I'll write some automated tests for this before merging it. But for manual testing, mostly just view some pubs with various kinds of data and relations. I think our arcadia seed has a pretty good set of test data, now that I've added a few cases where there are potentially infinite cycles (that we handle gracefully).

https://github.com/user-attachments/assets/64f819e8-83ab-4d04-b96b-83bfc7f8c73d

Site pub:
image
image

Journal pub:
image

Issue pub:
image

Article pub with circular reference:
localhost_3000_c_arcadia-research_pubs_511ddd91-bc27-4c50-af01-f0d5d9a6b5dd (1)

Author pub with circular reference:
localhost_3000_c_arcadia-research_pubs_511ddd91-bc27-4c50-af01-f0d5d9a6b5dd (2)


core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx
```diff
@@ -0,0 +1,145 @@
...
+};
+
+const PubValue = ({ value }: { value: FullProcessedPub["values"][number] }) => {
+	const [isOpen, setIsOpen] = useState(false);
+	if (value.relatedPub) {
+		const { relatedPub, ...justValue } = value;
+		const justValueElement = justValue.value ? (
+			<span className="mr-2 italic">{<PubValue value={justValue} />}:</span>
+		) : null;
+		if (relatedPub.isCycle) {
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:79]: confusingly the root pub is being returned with `isCycle: true`, which we should investigate. it doesn't cause problems here since we don't check `isCycle` on the root pub, only the related pubs.

core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx
```diff
@@ -0,0 +1,145 @@
...
+	// Currently, we are only rendering string versions of fields, except for file uploads
+	// For file uploads, because Unjournal doesn't have schemaNames yet, we check the value structure
+	const fileUploadSchema = getJsonSchemaByCoreSchemaType(CoreSchemaType.FileUpload);
+	if (Value.Check(fileUploadSchema, value.value)) {
+		return <FileUploadPreview files={value.value as FileUpload} />;
+	}
+
+	if (value.schemaName === CoreSchemaType.DateTime) {
+		return new Date(value.value as string).toISOString().split("T")[0];
+	}
```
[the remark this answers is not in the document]
Comment (kalilsn) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:131]: but otherwise this logic for rendering a single field is unchanged from the old function

core/lib/pubs.ts
```diff
@@ -85,3 +85,15 @@ export const processedPubToPubResult = <T extends InputPub>(pub: T): GetPubsResu
...
+	pub: T
+): T["pubType"]["fields"][number] | undefined => pub.pubType.fields.find((field) => field.isTitle);
+
+export const valuesWithoutTitle = <T extends InputPub>(pub: T): T["values"] => {
+	const titleField = getTitleField(pub);
+	if (!titleField) {
+		return pub.values;
+	}
+	return pub.values.filter((value) => value.fieldId !== titleField?.id);
+};
```
kalilsn, line 99: I modified how we filter the title out of the values rendered on the pub page because it was relying on the title field being named `"Title"` which led to unintuitive behavior if there was a field named `"Title"` that wasn't the type's title, or if another field was the title.

core/lib/server/pub.ts
```diff
@@ -1537,9 +1554,9 @@ function nestRelatedPubsAndChildren<Options extends GetPubsWithRelatedValuesAndC
 	const processedPubsById = new Map<PubsId, ProcessedPub<Options>>();
 
 	function processPub(pubId: PubsId, depth: number): ProcessedPub<Options> | undefined {
-		// if (depth < 0) {
-		// 	return processedPubsById.get(pubId);
-		// }
+		if (depth < 0) {
+			return processedPubsById.get(pubId);
+		}
```
kalilsn, line 1559: Uncommenting this fixed an infinite recursion issue when selecting pubs with a cycle

core/lib/server/pub.ts
```diff
@@ -576,29 +581,35 @@ export const createPubRecursiveNew = async <Body extends CreatePubRequestBodyWit
 				stageId: createdStageId ?? null,
 				values: hydratedValues,
 				children: children.length ? children : [],
+				depth,
```
kalilsn, line 584: I'm passing the depth around everywhere (probably in some places it doesn't totally make sense) to satisfy the type, which I made include the depth. There is probably a more correct way to handle this!

PR #869, opened 2024-12-21
Update google drive action

gabestein:
- Casts some values so they validate
- Allows use of set communitySlug rather than hardcoding
- Adds discussions and versions only when they don't already exist
- Adds a new version if old and new content don't match

## Test Plan
1. Test the google drive action on the gdrive folder with a URL field configured to hold a gdrive folder
2. Ensure that:
- On a first run, new discussions and versions are created
- On second run, no new versions and discussions are created
3. Update the main doc in the google drive folder
4. Re-run the action
5. Ensure that a new version is created, with new content


core/actions/googleDriveImport/formatDriveData.ts
```diff
@@ -116,12 +116,12 @@ export const formatDriveData = async (dataFromDrive: DriveData): Promise<Formatt
...
 							"arcadia:anchor":
 								index === 0 && discussion.anchors.length
-									? discussion.anchors[0]
+									? JSON.stringify(discussion.anchors[0])
 									: undefined,
 							"arcadia:content": comment.text,
 							"arcadia:publication-date": comment.createdAt,
 							"arcadia:full-name": comment.author.fullName,
-							"arcadia:orcid": comment.author.orcid,
+							"arcadia:orcid": `https://orcid.org/${comment.author.orcid}`,
```
isTravis, line 124: Current use pages on legacy just render the orcid value - (and construct the URL for the href). Is there a reason to store the whole URL, as opposed to just the ORCID? Somehow feels cleaner to me to construct the URL when we need, than to try and parse out just the ID value when we need it.

isTravis (approved):
This does fix the anchor validation issue. The problem I'm bumping into now is that it doesn't remove or check for duplicate IDs on versions or discussions when re-run, and errors on that. I knew that was one of the things still to fix. 

My preference would be for ORCID to be a string, not URL type, and to just store the ID - not the full `https...`. If there's a good reason not to do it that way though, this should work as is.

core/actions/googleDriveImport/formatDriveData.ts
gabestein: Yes. Storing them as just the ID part causes a bunch of problems on Legacy because ORCIDs are technically fully qualified URIs — the https:// is part of the ID — and people are used to pasting in the full URL on every other system. All the validators are built assuming this. Etc. So on Legacy we're parsing in and parsing out all the time anyway.

gabestein:
See comment above for why we should store ORCID as a URL.

Re: duplicates...what's the error? My assumption would be that it would successfully create duplicates via the upsert.

isTravis (approved):

PR #870, opened 2024-12-22
Fix form builder and default forms

### LINKED ISSUE #868: Default form does not load as default form

1. Go to a default form
2. Delete a field from it or update the header
3. Go to a pub for that type and click update
4. A nice, functional auto-generated form will load, but it is not the default form!

Default form should load

kalilsn:
## Issue(s) Resolved
Resolves #868 
There are 3 different bugs causing unwanted fields to show up on the pubs page. This fixes two of them:
- [Recently](https://github.com/pubpub/platform/commit/50cac43b80e3a8a916630cd3ebb0c482f8e84f48#diff-26d737b62698c65862e8c8252c2fbd819fbce2be629a5dac746e0c2c4df15414L59-R60) we removed what looked like an extraneous aliasing of `id` to `elementId`. This made it impossible to delete form elements because [we use the presence of `elementId` to check whether the element has been saved already](https://github.com/pubpub/platform/blob/main/core/app/components/FormBuilder/FormBuilder.tsx#L211-L220). If it hasn't been saved, it's immediately removed from the form. This change caused saved elements to have the same behavior, rather than marking them with `{ deleted: true }` and sending them to the backend to actually get removed.
- In the default form for pub updates we render an input for any value that exists on the pub, even if it's not on the form. But we were also inadvertently rendering an element for every field on the pub type.

The third bug is just that we often create "empty" values and we don't have any way to delete them. So even after merging this and actually removing an element from the form, it may still show up on the pub update form until we make it possible to delete values. I would just filter these out in the query, except that there are a variety of empty values (`""`, `" "`,`[{}]`) not just `null` so I don't really want to make a decision about how to handle that now. The create pub form should update correctly though!

## Test Plan
Same as the issue!


core/lib/server/pubFields.ts
```diff
@@ -35,8 +35,7 @@ type GetPubFieldsInput =
  * Get pub fields
  *
  * @param props  - When nothing is supplied, return all the pub fields
- * @param props.pubId - When supplied, return all the pub fields associated with the pub through pub values and the pub type of that pub
- * When props.valuesOnly is true, only return the pub fields associated with the pub through pub values, not through the pub type
```
kalilsn, line 39: the `valuesOnly` prop was removed from the type at some point, but the query was never updated. and now `valuesOnly` is actually the behavior we always want so i've updated the comments and removed the additional pub type fields join from the query.

i also checked and the only place we use this version of the query (passing the pubid) is on the pub editor that i'm fixing here.

core/app/components/FormBuilder/FormBuilder.tsx
```diff
@@ -147,8 +147,9 @@ export function FormBuilder({ pubForm, id, stages }: Props) {
 		values: {
 			elements: pubForm.elements.map((e) => {
 				// Do not include slug here
-				const { slug, ...rest } = e;
-				return rest;
+				const { slug, id, ...rest } = e;
+				// Rename id to avoid conflict with rhf generated id
+				return { ...rest, elementId: id };
```
kalilsn, line 152: rather than restore the alias in the query, i rename id here, which is a little closer to where we actually use the elementId prop

allisonking (approved):
tested this out and was able to delete + update the default form and have it reflect in the update page!

tefkah:
Just for my understanding, does this now still let the  pub/edit page render fields on the pub that are not on the default form?

bc i think it should! otherwise how are you ever to change values whose fields are not on the default form?

PR #873, opened 2024-12-25
Fix: pubs with no discussions

gabestein:
## Issue(s) Resolved
Fixes three errors:
1. Where we would send a bad query if a legacy pub had no discussions.
2. Where we assumed a discussion always had a `comment.author`, but in cases with anonymous comments it has a `comment.commenter`.
3. We were passing in an undefined ORCID url even if the user didn't have an ORCID

### Discussions Issue
- Import a pub with no discussions and make sure versions show up
- Import a pub with discussions and make sure they show up
- Repeat both and make sure both work as expected.

### Comments and ORCID Issue
- Import a pub with anonymous `comment.commenter` discussions.
- Make sure it loads
- Make sure any users with ORCIDs and avatars have them passed in

### DIFF
--- a/core/actions/googleDriveImport/formatDriveData.ts
+++ b/core/actions/googleDriveImport/formatDriveData.ts
@@ -119,11 +119,12 @@ export const formatDriveData = async (
 						: comment.commenter
 							? comment.commenter.avatar
 							: "";
-					const commentAuthorORCID = comment.author
-						? comment.author.orcid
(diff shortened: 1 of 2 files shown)


isTravis (approved):
💯 I bumped into that commenter issue on some of the legacy processing earlier - I should've thought of that and mentioned it when we first saw this hiccup. Nice catch!

PR #875, opened 2025-01-06
dev (next 15|1): update to next 15 & react 19

tefkah:
## Issue(s) Resolved
New shiny.

Updates Next to `v15` and react to `v19`. 

This mostly brings the following changes:

- Page "request" APIs are now async. So it's `await cookies()` or `await props.params`. [see here for more info](https://nextjs.org/docs/messages/sync-dynamic-apis). This is `next`s new way of detecting when a page needs to be dynamic, which allows for easier partial prerendering. I will experiment with it a bit and see if we can leverage it, would be nice!
- Upgrading a bunch of dependencies st they can use react 19 (not complete yet, but seemingly good enough)

What next 15 can bring us:
- much better sentry reporting for server components. I haven't added this yet, but using `onRequestError` in `instumentation.ts` will allow us to capture unhandled exceptions in server components. https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation#onrequesterror-optional
- getting closer to being able to use the react compiler, which would allow us to not worry about memoization
- more secure actions: unused actions are removed, and their routes are no longer guessable
- slightly faster builds: as i'm writing this i don't know how it is on CI, but on my machine it's 124sec vs 199sec for a complete run of `pnpm build` (measured as `time pnpm build`) on 15 vs 14 respectively: pretty big improvement!
- possibility to use partial prerender: https://nextjs.org/docs/app/building-your-application/rendering/partial-prerendering. Basically, we would be able to render the shell/loading states of our app statically and serve that immediately, and then server render the actual contents. this should improve the snappiness of our app a bunch.

Other notes:
- You might get some `pnpm i` warnings, as eg `ts-rest` has not updated their types to allow for `react 19`. I don't think this makes much difference tho
- I also updated typescript while i was here. 

## Test Plan
1. Run the integrations gauntlet (maybe not necessary, are we deprecating them?)
2. Poke around the app and see if you find something egregiously wrong
3. If possible (i don't know how) test file uploads.


core/app/(user)/magic-link/route.ts
```diff
@@ -85,7 +85,7 @@ export async function GET(req: NextRequest) {
 
 	const validatedTokenPromise = validateToken(token);
 
-	const currentSessionCookie = cookies().get(lucia.sessionCookieName)?.value;
+	const currentSessionCookie = (await cookies()).get(lucia.sessionCookieName)?.value;
```
tefkah, line 88: most of the changes are of this caliber: `await`ing `cookies`, `headers`, `params`, and `searchParams`. I will comment on more nuanced changes

core/app/c/(public)/[communitySlug]/public/Header.tsx
```diff
@@ -1,22 +1,11 @@
...
-const CopyCurrentUrlButton = dynamic(
-	() =>
-		import("~/app/components/CopyCurrentUrlButton").then(
-			(module) => module.CopyCurrentUrlButton
-		),
-	{
-		ssr: false,
-	}
-);
-
```
Comment (tefkah) [inline core/app/c/(public)/[communitySlug]/public/Header.tsx:19]: In next 15 you can no longer do `ssr: false` dynamic imports in Server Components, they have to be done in client components.
In cases where this happened, I have simply moved the dynamic imports to their component file, as the reason for doing dynamic imports in these cases was not for code-splitting, but to simply forego SSR (because we are accessing `window` directly)

core/app/c/[communitySlug]/types/TypeEditor.tsx
```diff
@@ -172,7 +172,7 @@ export const TypeEditor = ({ onTypeCreation }: Props) => {
 														</td>
 														<td>
 															<FormItem
-																data-testId={`${pubField.slug}-titleField`}
+																data-testid={`${pubField.slug}-titleField`}
```
Comment (tefkah) [inline core/app/c/[communitySlug]/types/TypeEditor.tsx:175]: i added this incorrect attribute in a pr a few months ago which was messing up tests

core/app/components/CopyCurrentUrlButton.tsx
```diff
@@ -1,10 +1,17 @@
...
 
-export const CopyCurrentUrlButton = (props: Props) => {
+const CopyCurrentUrlButtonBase = (props: Props) => {
 	return <CopyButton value={window.location.href} {...props} />;
 };
+
+// necessary in order to disable SSR, as window is not available on the server
+export const CopyCurrentUrlButton = dynamic(() => Promise.resolve(CopyCurrentUrlButtonBase), {
+	ssr: false,
+});
```
tefkah, line 17: did this bc we can no longer define dynamic imports with `ssr: false` in server components.

core/app/components/pubs/RemovePubForm.tsx
```diff
@@ -15,7 +15,7 @@ const PubRemoveForm = dynamic(
 			default: mod.PubRemoveForm,
 		}));
 	},
-	{ ssr: false, loading: () => <SkeletonCard /> }
+	{ loading: () => <SkeletonCard /> }
```
tefkah, line 18: i'm not so sure about my decision to just remove `ssr: false` here, i wonder if this still accomplishes the correct codesplitting behavior. if so, then this is fine

core/instrumentation.ts
```diff
@@ -26,3 +28,10 @@ export async function register() {
 		logger.info("NEXT_RUNTIME is not `nodejs`; skipping OTEL registration.");
 	}
 }
+
+export const onRequestError: InstrumentationOnRequestError = async (...args) => {
+	// necessary in order to avoid OOM errors on build due to Sentry making it
+	// hard for webpack to optimize
+	const onRequestError = await import("./instrumentation.onRequestError.ts");
+	return onRequestError.onRequestError(...args);
+};
```
tefkah, line 37: i did this dynamic import here as well because i thought it might help prevent some OOM errors during build. I don't think it does, but it's fine to keep it like so (although maybe i should remove the comment)

core/lib/server/token.ts
```diff
@@ -81,7 +81,7 @@ export const validateToken = async (token: string, type?: AuthTokenType, trx = d
 	// because timingSafeEqual throws an error in that case, which could leak the length of the key.
 	// We aren't worried about that because we're hashing the values first (so they're constant
 	// length) and because our tokens are all the same length anyways, unlike a password.
-	if (!crypto.timingSafeEqual(dbHash, inputHash)) {
+	if (!crypto.timingSafeEqual(new Uint8Array(dbHash), new Uint8Array(inputHash))) {
```
tefkah, line 84: this change is due to a typescript update i also added (it still works without the `new UInt8Array` cast)

This does work like this, but kind of annoying to have to do

core/package.json
```diff
@@ -60,9 +60,9 @@
 		"@dnd-kit/utilities": "^3.2.2",
 		"@faker-js/faker": "^9.0.0",
 		"@honeycombio/opentelemetry-node": "catalog:",
-		"@hookform/resolvers": "^3.3.1",
+		"@hookform/resolvers": "catalog:",
 		"@icons-pack/react-simple-icons": "^10.2.0",
-		"@nimpl/getters": "^1.4.1",
+		"@nimpl/getters": "^2.0.0",
```
tefkah, line 65: version that works with next 15

core/package.json
```diff
@@ -126,6 +128,7 @@
 		"remark-rehype": "^11.1.0",
 		"remark-stringify": "^11.0.0",
 		"remove": "^0.1.5",
+		"require-in-the-middle": "^7.4.0",
```
tefkah, line 131: i remember this and `import-in-the-middle` being necessary due to how `serverExternalPackages` setting in `next.config.js` now works. it just threw some warnings, but this got rid of it.

ideally experiment with removing this

core/playwright/externalFormCreatePub.spec.ts
```diff
@@ -20,6 +20,12 @@ let page: Page;
 
 test.beforeAll(async ({ browser }) => {
 	page = await browser.newPage();
+	page.on("console", async (msg) => {
+		if (msg.type() === "error") {
+			// eslint-disable-next-line no-console
+			console.error("Error:", msg, msg.location());
+		}
+	});
```
tefkah, line 28: i was having big troubles with the tests, and this shows the browser error logs in the playwright logs.
let me know if i should remove it

core/playwright/externalFormCreatePub.spec.ts
```diff
@@ -73,7 +79,7 @@ test("Can create a pub from an external form", async () => {
 
 test.describe("Multivalue inputs", () => {
 	test("Can add multivalue inputs", async () => {
-		test.setTimeout(60_000);
+		// test.setTimeout(60_000);
```
tefkah, line 82: don't recall the reason for the big timeout, should probably remove comment

core/playwright/externalFormCreatePub.spec.ts
```diff
@@ -141,9 +151,6 @@ test.describe("Multivalue inputs", () => {
 
 		// Save the form builder and go to external form
 		await formEditPage.saveForm();
-		await expect(
-			page.getByRole("status").filter({ hasText: "Form Successfully Saved" })
-		).toHaveCount(1);
```
tefkah, line 146: i moved this chec to the form fixture

tefkah: although for some reason this message does not show anymore, so i just wait for the page to resolve instead. might be bc of some `redirect` logic

integrations/evaluations/package.json
tefkah, line 1: can we remove integrations yet 🥺

3mcd: yes

3mcd (approved):

PR #879, opened 2025-01-09
RFC: add slugs to pubs

tefkah:
## Issue(s) Resolved

- hard to keep track of uuids for a human, eg when exploring the API
- uuids give ugly urls

This is not per se meant to be merged in, I'm more interested in hearing your thoughts about this (which i think is easier with a clear implementation) than it being merged in, so I am keeping it a draft.

Slugs!

Implemented very similary to forms. I have separate branch ([tfk/api-pub-slugs](https://github.com/pubpub/platform/tree/tfk/api-pub-slugs) that allows you to query pubs by slugs in the api.

Slugs are scoped to a community. We could theoretically also do `<community-slug>:<pub-slug>` like with fields, but that does not sound great to me.
I always found it extremely confusing in Legacy that you could not create a slug that was already used by a different community, so instead i made slugs unique based on slug + community.

Additional work will include rewriting more functions to take `slug` in addition to `pubId`


core/lib/paths.ts
```diff
@@ -0,0 +1,10 @@
+type PubPath = `/c/${string}/pubs/${string}`;
+type PubEditPath = `/c/${string}/pubs/${string}/edit`;
+type PubCreatePath = `/c/${string}/pubs/create`;
+
+export const pubPath = (communitySlug: string, pubSlug: string): PubPath =>
+	`/c/${communitySlug}/pubs/${pubSlug}` as const;
+export const pubEditPath = (communitySlug: string, pubSlug: string): PubEditPath =>
+	`/c/${communitySlug}/pubs/${pubSlug}/edit` as const;
+export const pubCreatePath = (communitySlug: string): PubCreatePath =>
+	`/c/${communitySlug}/pubs/create` as const;
```
tefkah, line 10: ideally we'd use something more sophisticated for typed routes, but at least this makes linking to routes slightly less error prone

PR #881, opened 2025-01-09
dev: change everything to ESM/`"type":"module"`

tefkah:
- **feat: type module everything**
- **feat: add preconstruct config everwhere**
- **fix: remove all those weird exports**
- **fix: correct imports**
- **fix: explicitly label cjs**
- **fix: do some crazy stuff to make the seed script still work**
- **fix: i win**
- **chore: esm issue in kanel**


core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts
```diff
@@ -9,9 +9,7 @@ import type {
 	LastModifiedBy,
 } from "db/types";
 import { siteApi } from "contracts";
-import { ApiAccessScope, ApiAccessType } from "db/public";
-import { Capabilities } from "db/src/public/Capabilities";
-import { MembershipType } from "db/src/public/MembershipType";
+import { ApiAccessScope, ApiAccessType, Capabilities, MembershipType } from "db/public";
```
Comment (tefkah) [inline core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts:12]: most of the changes in this PR are changing these incorrect imports.

core/tailwind.config.cjs
tefkah, line 1: instead of renaming all these files to `.cjs` it is probably possible to just rewrite them in `ESM`, but I did not want to create too many changes!

jobs/tsconfig.json
```diff
@@ -3,7 +3,8 @@
 	"compilerOptions": {
 		"tsBuildInfoFile": "node_modules/.cache/tsbuildinfo.json",
 		"lib": ["ESNext"],
-		"types": ["node"]
+		"types": ["node"],
+		"moduleResolution": "bundler"
```
tefkah, line 7: This is needed for type-checking to pass, which I _think_ is correct as we are just straight running this with `tsx`, not building this package as an output. However, this is mostly figured out by trial and error, not by some very deep understanding of tsconfig best practices

core/.env.test
```diff
@@ -15,3 +15,5 @@ OTEL_SERVICE_NAME="pubpub-v7-dev" # should be shared across components but not e
 HONEYCOMB_API_KEY="xxx"
 
 KYSELY_DEBUG="true"
+
+GCLOUD_KEY_FILE="xxx"
```
tefkah, line 19: in one of these prs this will get added

config/tsconfig/react-library.json
```diff
@@ -6,6 +6,7 @@
 		"lib": ["DOM", "ESNext"],
 		"module": "ESNext",
 		"target": "ES6",
-		"jsx": "react-jsx"
+		"jsx": "react-jsx",
+		"moduleResolution": "bundler"
```
tefkah, line 10: i'm less certain about this change then doing the same to `jobs`, as we are compiling `ui` as a separate package. 

I did this for now, as the default `moduleResolution` setting proved incorrect, and the probably correct setting of `NodeNext` would require us to add `.js` to every import statement, eg turn

```ts
import { Button } from './Button';
```

into

```
import { Button } from './Button.js';
```

I think this would be good ((although we could also specify `.ts` extensions now)[https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#path-rewriting-for-relative-paths]), but again would be too many changes at once.

core/prisma/scripts/comment-generator.mts
```diff
@@ -23,8 +23,8 @@ import { promises as fs } from "fs";
 
 import type { DMMF, EnvValue, GeneratorOptions } from "@prisma/generator-helper";
 
-import { generatorHandler } from "@prisma/generator-helper";
-import { getDMMF, parseEnvValue } from "@prisma/internals";
+import Helper from "@prisma/generator-helper";
+import Internals from "@prisma/internals";
```
tefkah, line 27: similar to the `pg` imports in `core/kysely/db`, prisma has some weird export defined or something where this will fail if you do 

```ts
import { parseEnvValue } from '@prisma/generate-helper';
```

core/prisma/seed/seed-wrapper.cts
tefkah, line 1: this is kind of evil but works, see pr description

packages/schemas/package.json
```diff
@@ -22,5 +28,17 @@
 		"typescript": "catalog:",
 		"vitest": "^1.3.1"
 	},
+	"preconstruct": {
+		"entrypoints": [
+			"index.ts",
+			"schemas.ts"
+		],
```
tefkah, line 35: easy to miss, but I added an extra entrypoint `schemas.ts` and thus export `schemas/schemas`, because in some places we were importing things from `schema/src/schemas`, which is not what we are supposed to do.

core/.env.test
tefkah: for ref: this is needed to do `pnpm --filter core vitest`

3mcd (approved):

PR #894, opened 2025-01-13
RFC: remove trailing semicolons

3mcd:
## Issue(s) Resolved

N/A

I like to omit trailing semicolons in JS projects because I feel they increase the visual noise of code. I also find code without trailing semicolons more aesthetically pleasing.

Perhaps more convincingly, there appears to be an increasing trend to omit trailing semicolons in open source projects. I took a sample of JS/TS projects from my GitHub feed to get a (very) rough estimate of the number of projects that use trailing semicolons vs. those that do not. I navigated to the six most recently updated projects in my feed and opened a random TS or JS source file:

- https://github.com/es-data/utils/blob/main/src/series/series.ts
- https://github.com/chakra-ui/zag/blob/main/packages/core/src/machine.ts
- https://github.com/yjs/yjs/blob/bb5410b6ddc74a23e7bdb75aa09d252aaed81b8d/src/types/YText.js
- https://github.com/panuchka/bitECS/blob/master/src/Serialize.js
- https://github.com/rmp135/sql-ts/blob/master/src/ColumnTasks.ts

Each of the six projects I visited eschewed use of trailing semicolons.

I feel we should adopt this standard to catch up to the current code style conventions of JS projects. And selfishly for the subjective reasons I listed above.

## Test Plan

N/A


tefkah:
I'm totally on board for this, only considerations:

- This messes up `git blame` a bit, will make it harder to figure out when things were changed
- Ideally: could this wait until we (know whether we want to) merge some of the PRs I have open? I fear this would introduce a bunch of annoying merge conflicts, which are much easier to solve on this end (just run `pnpm format:fix` on `main` again)

otherwise, let's gooo

3mcd:
@tefkah Sounds good. This will be easy to reconcile with main when we get everything else merged.

tefkah (approved):

PR #895, opened 2025-01-14
fix(#889): reflect pub page updates after action run

### LINKED ISSUE #889: Pub values don't update on front-end after action runs

1. Updated a pub via an action on the pub page
2. When action completed, pub page did not update with new values
3. Refreshing the page showed new values

Pub page to update automatically when an action runs that changes values

E.g. the gdrive action on Arcadia's community

tefkah:
## Issue(s) Resolved
Resolves #889

Almost certainly caused bc `upsertPubRelations` was not `await`ed. 
Did not test this (bc local setup is kind of tedious), but I'm 99% sure this is the reason

## Test Plan
1. Send it
2. Run GDrive import action on prod on pub page, see if it reflects changes.


3mcd (approved):
yolo

gabestein:
Tested and can confirm this is working!

Issue #884, opened 2025-01-12
Add link to pub page in action log


gabestein:
## Motivation

So admins can quickly get from the action log to the Pub on which the action was run

## Requirements

- Adds a link to the pub on which the action ran (or failed to run on) on the action log page

PR #896, opened 2025-01-14
feat: add links to pubs in stagespanel

tefkah:
## Issue(s) Resolved
Annoying not to be able to click on Pubs to go to them in the stages panel.

Does what it says on the tin.

## Test Plan
1. Go to `/c/arcadia-research/stages/manage`
2. Click on a stage
3. Click on a pub
4. You should be transported to it.

https://github.com/user-attachments/assets/cf131772-0a10-43a7-911c-3e44ca775074

## Notes
It would be nice if we had a slightly more generic Title+link component, ideally you should be able to choose whether it has a link or not. Maybe this could be part of #884 @allisonking ?


allisonking (approved):
gengar

allisonking:
> It would be nice if we had a slightly more generic Title+link component, ideally you should be able to choose whether it has a link or not. Maybe this could be part of https://github.com/pubpub/platform/issues/884 @allisonking ?

Sure, good idea!

PR #897, opened 2025-01-14
fix: update http action deps and make it work for string properties again

tefkah:
- **chore: upgrade jsonpath-plus to remove security warning**
- **fix: do not wrap jsonpath results in array by default if no array**

## Issue(s) Resolved
Dependabot complaints about `json-path-plus`. I don't think they're very serious for us as no one is really using the app, but not allowing RCE is probably good haha.

Secondly, the `http` action was no longer working for the most simple use case, namely selecting a property of a response object to set as eg the title.

To see the issue, try the following configuration on a test community 

image

This should set the title to `gengar`, but it does not.

By default, `jsonpath-plus` returns for `{ name: "gengar" }` and `$.name` => `[ 'gengar' ]`.
Maybe we used to handle this case, but it no longer works.

By setting `wrap: false` in `JSONPath`, it now returns `'gengar` like expected.

## Test Plan

1. Go to croccroc
2. Add new action to a stage with pubs using this config
image

4. Run action
5. Name should change to `"gengar"`


tefkah:
i could also add a test that does the test plan described, but i'm slightly hesitant to use external apis in our tests (what if they go down?)

3mcd (approved):

PR #898, opened 2025-01-14
chore: remove integration deployments

3mcd:
This PR removes the ECS task definitions for our two legacy integrations. I also removed the integrations from our dockerfile and updated the monorepo lockfile.

Terraform plan (it's big, and only destructive changes): [plan.txt](https://github.com/user-attachments/files/18413837/plan.txt)


tefkah (approved):
💯

only suggestion i'd have would be to also delete these two lines in `blake`

https://github.com/pubpub/platform/blob/bc6c85d1d0ca8754d506eb5d68c12a4b962000f7/infrastructure/terraform/environments/blake/main.tf#L46-L47

PR #899, opened 2025-01-14
remove integration code and tables

3mcd:


tefkah (approved):
Fantastic!

On another branch I removed a bunch of extra references/files to do with integrations! Not per se necessary to merge in, but would be nice :)
(https://github.com/pubpub/platform/tree/tfk/remove-integrations-code)

As a follow up, we could also finally delete `prisma`...

Issue #790, opened 2024-11-19
Ability to add relation fields to forms


gabestein:
## Motivation
So admins can add reference fields to their forms/editors

## Requirements
- Ability to add new “relation” field to form
- Ability to define field label
- Ability to define helper text
- Ability to define an input for the “value” of the reference field

## Acceptance Criteria
- Should remove existing filter that restricts user from choosing a relation field when creating a new form field
- Adds new section for configuring label, helper text, and restricting pub type for the relation chooser 
- Puts existing UI for configuring field input for relation value below new section for configuring relation helper text and label under section called "Relation Value"

## Non-Goals
- Ability to specify a “child” or “relationship” reference type — we're getting rid of child relationships, so we don't need to specify this

## Mockups
https://www.figma.com/design/SqOTMElJn7fuPljfQL1gIY/PubPub-Platform?node-id=1701-25138&m=dev

## Extra Work
- Ability to restrict related pubs to a particular pub type

gabestein:
@ktyjscott — I think you were working on adding the last requirement to mocks, but I didn't see it, so lmk if it's ready yet or the link has changed

PR #900, opened 2025-01-14
Form builder relation fields

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/790

Adds the ability to configure a relationship field.

## Test Plan

1. Add a relationship field with any core schema type
2. Add this relationship field to a form via the form builder
3. You should be able to configure both the relationship and the value
4. This should save to the db FormElement table with a `config` value looking something like
```js
{
  relationshipConfig: {
    component: "relationBlock",
    label: "Authors",
    help: "Authors associated with this pub",
  },
  label: "Role",
  help: "Author roles",
  minLength: 1,
}
```

image

image

For `CoreSchemaType.Null`, I decided not to render the "relationship value" portion at all. I went back and forth about whether to just use a new `InputComponent` type of `null` but that didn't feel quite right either since we don't really want the user to select an appearance for null. So I think either way we have to special case it, but open to suggestions!


core/app/components/FormBuilder/ElementPanel/ComponentConfig/RelationBlock.tsx
```diff
@@ -0,0 +1,76 @@
...
+				control={form.control}
+				name="config.relationshipConfig.label"
+				render={({ field }) => (
+					<FormItem>
+						<FormLabel>Label</FormLabel>
+						<FormControl>
+							<Input
+								placeholder="Description of the selection"
+								{...field}
+								value={field.value ?? ""}
```
allisonking, line 23: This is so we don't get the error about a component going from controlled to uncontrolled. Not all of the fields have this though. I think we may want this for every text field, either that or figure out how to dynamically render default values for `config` with empty strings.

tefkah: yeah this is a really annoying one. i'm not sure what the best way of doing it is, but for these non-Form-fields i think this approach is fine. another way of doing it is by setting a "defaultValue" in the `useForm`, but this is nicer in this context

packages/schemas/src/schemaComponents.ts
```diff
@@ -108,6 +119,19 @@ export const richTextInputConfigSchema = Type.Object({
...
 	help: Type.Optional(Type.String()),
 });
+export const relationBlockConfigSchema = Type.Object(
+	{
+		// "relationshipConfig" is the config 'around' the related values
+		relationshipConfig: Type.Object({
+			label: Type.Optional(Type.String()),
+			help: Type.Optional(Type.String()),
+			pubType: Type.Optional(Type.String()),
+			component: Type.Enum(InputComponent),
```
tefkah, line 129: hmm, the more i think about it I'm not sure if having the `pubType` on the `config` column is very nice. I will be very annoying to update users if they either archive, change, or delete a `PubType` that they still have a Form element that only accepts pubs of that type. 
I think for now it could be fine, but ideally i think it being (yet another) column would be better, but i don't really want to make that call on my own! (@3mcd would appreciate your input)

If we do go with this config setup, could we maybe make the `pubType` -> `pubTypes`, ie an array instead? i could foresee us in the future allow multiple types of pubtypes for the same field (maybe both a figure and a plain image for instance), and it might save us a migration.

tefkah (approved):
i had one question about whether we really want the `pubType` to really be hidden in a `config` column and all the ref. integrity issues that come with that, but if that's not a concern it looks great otherwise!

would appreciate some thoughts from @kalilsn  and @3mcd on that topic, but otherwise it LGTM!

PR #902, opened 2025-01-15
chore: remove changesets

3mcd:
## Issue(s) Resolved

N/A

Removes changesets. We've never used it.

## Test Plan

N/A


pnpm-lock.yaml
```diff
@@ -14934,6 +14611,60 @@ snapshots:
 
   '@opentelemetry/api@1.9.0': {}
 
+  '@opentelemetry/auto-instrumentations-node@0.53.0(@opentelemetry/api@1.7.0)':
```
3mcd, line 14614: I'm not totally sure why a bunch of opentelemetry packages were added to the lockfile. All I did was remove the `@changesets/cli` dev dependency and run `pnpm i`.

tefkah (approved):
i do think it would be good to do changelogs at some point, but we'll get there when we get there

PR #905, opened 2025-01-15
fix: imply view permission to parent pub in form fill page

3mcd:
## Issue(s) Resolved

N/A

Fixes a regression where invited members cannot view forms because they are unable to view the edited pub's parent pub.

## Test Plan

On the Unjournal community,
1. Invite a (non-admin, non-superadmin) member to fill a pub via the email action
2. Open the form link in the email
3. The form fill page should load properly


gabestein (approved):
Thanks!

Issue #887, opened 2025-01-12
Create functions that upsert/delete a tree of pubs


gabestein:
## Motivation

Currently, it's very hard to do imports with multiple pubs nicely.

Say we want to import a pub from legacy with a bunch of relations.

The first run is easy enough, you can use `createPubRecursiveNew` to do so.

However, say in the second run some of the relations have changed (one comment was deleted, one was added, one was updated), and some values on the pub were changed. Now you cannot use `createPubRecursiveNew`, bc it would just create another pub tree. You would need to

1. Check if parent pub exists
2. Update normal values of parent pub
3. Check if all relations exist
4. Delete unused pub relations
5. Create a custom `upsertRelations` body that creates a new pub and relation for the added comment but only updates the relation for the updated comment.

This is very annoying to do, and this is only one relation deep!

gabestein:
After experiencing this via the API, I definitely think this is worth doing — as we get into imports from Legacy, it'll be increasingly needed.

tefkah:
I think this will be nice for a lot of different import cases!

Questions:
- Should related pubs always get deleted when you do such an upsert? Eg if the second run has fewer relations than the first, what should happen with the leftover ones? Is there a nice api for specifying this conditionally?

gabestein:
@tefkah -- when you can, let us know the status here so we can decide to keep it in or move it out!

PR #907, opened 2025-01-21
dev: api integration tests

tefkah:
- **feat: api tests**
- **fix: types and tests**
- **fix: add common 400 response type**

## Issue(s) Resolved

Mostly want this before implementing #887 in the API.

Adds integration tests for the API. For now just very basic tests: create  a pub, get a pub. Should add more in depth tests later.

Roughly:
1. Creates an API token through the UI
2. Initializes a `ts-rest` client
3. Makes some requests.

Other than the initial setup, these tests are quite fast as they don't really use the UI at all. 

I chose not to use `playwright`s built in API client, as it's not typed and is more error prone when it comes to routes. The goal is more to test API responses rather than how it interacts with any client.


core/app/c/[communitySlug]/settings/tokens/CreateTokenForm.tsx
```diff
@@ -61,6 +61,10 @@ export type CreateTokenForm = ReturnType<typeof useForm<CreateTokenFormSchema>>;
 export const CreateTokenForm = ({ context }: { context: CreateTokenFormContext }) => {
 	const form = useForm<CreateTokenFormSchema>({
 		resolver: zodResolver(createTokenFormSchema),
+		defaultValues: {
+			// default to 1 day from now, mostly to make testing easier
+			expiration: new Date(Date.now() + 1000 * 60 * 60 * 24),
+		},
```
Comment (tefkah) [inline core/app/c/[communitySlug]/settings/tokens/CreateTokenForm.tsx:67]: i didn't want to write the code to navigate the date picker, so i took the easy way out and set a default time instead. let me know if the default time should be longer or if we should do this differently.

packages/contracts/src/resources/site.ts
```diff
@@ -530,12 +548,19 @@ export const siteApi = contract.router(
 		},
 	},
 	{
+		strictStatusCodes: true,
```
tefkah, line 551: this makes the return types from the api client more legible. instead of getting 
```ts
type Response = 
| { status: 200, body: {...}} 
| { status: 201 | 202 | ...., body: any} 
```

you get

```ts
type Response = { status: 200, body: {...}} 
```

packages/contracts/src/resources/site.ts
```diff
@@ -530,12 +548,19 @@ export const siteApi = contract.router(
...
 				.string()
 				.regex(/^Bearer /)
 				.optional(),
 		}),
+		commonResponses: {
+			// this makes sure that 400 is always a valid response code
+			400: zodErrorSchema,
+			403: z.string(),
+			404: z.string(),
+		},
```
tefkah, line 564: did not know this was an option, now every route has the correct error responses at least.

core/playwright/site-api.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+	const apiTokenPage = new ApiTokenPage(page, COMMUNITY_SLUG);
+	await apiTokenPage.goto();
+	const token = await apiTokenPage.createToken({
+		name: "test token",
+		description: "test description",
+		permissions: true,
+	});
+
+	client = initClient(siteApi, {
+		baseUrl: `http://localhost:3000/`,
```
tefkah, line 41: this might need to be an env var

allisonking (approved):
woot, awesome!!

PR #908, opened 2025-01-21
rfc: add single place for defining page authorization checks

tefkah:
- **feat: add helper for doing auth checks on pages**
- **feat: replace a number of auth checks with new helper**


core/app/c/[communitySlug]/unauthorized/pageAuthorizationChecks.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+export const userCanViewPage = cache(
+	async <
+		T extends keyof typeof pageAuthorizationChecks,
+		Capability extends
+			(typeof pageAuthorizationChecks)[T] = (typeof pageAuthorizationChecks)[T],
+		Target extends Targeted<Capability> = Targeted<Capability>,
+	>(
+		page: T,
+		userId: UsersId,
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/unauthorized/pageAuthorizationChecks.tsx:143]: Since this is always called in a context where we've already retrieved the full user row, it would be nice to make this function accept a `User` here and then check `user.isSuperAdmin` before making the capabilities query. I miss my superadmin privileges in communities I'm not a member of 😄

core/app/c/[communitySlug]/unauthorized/pageAuthorizationChecks.tsx
```diff
@@ -0,0 +1,194 @@
...
+	allForms: Capabilities.editCommunity,
+	actionLog: Capabilities.editCommunity,
+	types: Capabilities.editCommunity,
+	fields: Capabilities.editCommunity,
+	members: Capabilities.editCommunity,
+	settings: Capabilities.editCommunity,
+	apiTokens: Capabilities.editCommunity,
+	docs: Capabilities.editCommunity,
+	apiDocs: [Capabilities.editCommunity, Capabilities.viewPub],
+} as const satisfies PageAuthorizationCheckMap;
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/unauthorized/pageAuthorizationChecks.tsx:49]: I agree, this is much nicer than what we've been doing!

core/app/c/[communitySlug]/unauthorized/pageAuthorizationChecks.tsx
Comment (tefkah) [reply core/app/c/[communitySlug]/unauthorized/pageAuthorizationChecks.tsx:143]: good idea!

PR #909, opened 2025-01-21
feat: add new recipient email field to email action

### LINKED ISSUE #767: Modify email action to accept Email as pubfield
## Motivation
So admins can use an email field to send emails to non-members — e.g., if they want to ask pseudonymous discussers to verify their email address via a form, which would then move the pub to a "verified" stage.

## Requirements

- Updates the email action to add another field that allows email address fields to be used to send emails via the email action. 

## Acceptance Criteria

- Renames "recipient email address" field in email action config to "Recipient Member"
- Makes "recipient member" optional
- Adds a new optional field called "Recipient email address" that accepts a pub email field
- If neither is provided, the action should fail
- If both are provided, the "recipient member" should take precedence
- If only one is provided, that one should be used
- If a `Recipient*` token is used in an email body, the action run should fail, and throw an error indicating that the value does not exist

3mcd:
## Issue(s) Resolved

Resolves #767

This PR renames the existing "recipient" email action config to "recipient member", makes it optional, then adds a new, optional "recipient" config that accepts an email address.

## Test Plan

1. Add an email action to a stage.
2. Specify neither a recipient nor recipient member in the action config.
3. Attempt to run the action. You should get an error message like `"No email recipient was specified"`.
4. Specify both a recipient and recipient member.
5. Run the action. The recipient member, not the recipient, should recieve the email.
6. Specify only a recipient email address.
7. Run the action. The recipient email address should recieve the email.
8. Update the email body with a `recipient*` token (e.g. `recipientname`).
9. Attempt to run the action. The action run should fail with an error reported in the action log like `"Used a recipient token without specifying a recipient"`


core/actions/datacite/action.ts
```diff
@@ -9,9 +9,9 @@ export const action = defineAction({
 	name: Action.datacite,
 	config: {
 		schema: z.object({
-			doi: z.string(),
+			doi: z.string().optional(),
 			doiPrefix: z.string().optional(),
-			doiSuffix: z.string(),
+			doiSuffix: z.string().optional(),
```
3mcd, line 14: Snuck in a small configuration fix for the DataCite action. Because you no longer need to specify a fallback DOI (or suffix), we can now take advantage of auto-generated DataCite DOIs.

core/actions/email/action.ts
```diff
@@ -28,11 +29,12 @@ export const action = defineAction({
 	params: {
 		schema: z
 			.object({
-				recipient: z
+				recipientEmail: z.string().email().describe("Recipient email address").optional(),
```
kalilsn, line 32: It's not obvious using the form that the member takes precedence. Is there any way to give this a tooltip explaining that? Or to give this field some styling (like lower opacity or a strikethrough) when the member field has a value?

core/lib/server/render/pub/renderMarkdownWithPub.ts
```diff
@@ -290,7 +290,10 @@ const renderMarkdownWithPubPlugin: Plugin<[utils.RenderWithPubContext]> = (conte
 					if ("form" in attrs) {
 						props.href = await utils.renderFormInviteLink({
 							formSlug: expect(attrs.form),
-							userId: context.recipient.user.id,
+							userId: expect(
+								context.recipient,
+								"Used a form invitation token without specifying a recipient member"
+							).user.id,
```
kalilsn, line 296: What do you think about making this error a little more verbose? Something like `'You can't use :link{form="slug"} directive with an email address. You must set the "Recipient member" field.'`

If you're open to it I'd also like similarly explicit messages for the other two token errors in `renderWithPubUtils`, specifying which token and being very explicit about what form field needs to be set.

core/actions/email/action.ts
kalilsn: I don't really know how this feature is intended to be used so this may be wrong, but I thought I remembered someone requesting to use an pub value with `CoreSchemaType.Email` for this. So does this also need the little pub value substitution button like the member field has?

core/playwright/email.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+		// Invite a new user to fill out the form
+		await runActionDialog.getByLabel("Recipient email address").fill(email);
+		await runActionDialog.getByLabel("Email subject").fill("Hello");
+		await runActionDialog.getByLabel("Email body").fill("Greetings");
+
+		await runActionDialog.getByRole("button", { name: "Run", exact: true }).click();
+		await page.getByRole("status").filter({ hasText: "Action ran successfully!" }).waitFor();
+		await runActionDialog.getByRole("button", { name: "Close", exact: true }).click();
+		await runActionDialog.waitFor({ state: "hidden" });
```
kalilsn, line 77: Not at all urgent, but it would be nice to extract some of this logic for opening the action dialog and running the email action to the `PubDetailsPage` fixture since there's some very similar code in `externalFormInvite.spec.ts`

kalilsn (commented):
overall this is great, I just had a few ux requests and one question about the intended use! thanks for adding an e2e test!

core/actions/email/action.ts
3mcd: Woops yeah, thanks for catching!

3mcd: @kalilsn The recipient email address field will now be disabled when the recipient member field has a value, reflecting the precedence rules!

Screenshot 2025-02-04 at 3 20 36 PM

core/lib/server/render/pub/renderMarkdownWithPub.ts
3mcd: Done!

core/playwright/email.spec.ts
3mcd: I extracted the shared test behavior into a new `runAction` method on the `PubDetailsPage` fixture.

PR #910, opened 2025-01-21
Related pubs in form

allisonking:


core/lib/server/pub.ts
```diff
@@ -1095,13 +1113,10 @@ export const updatePub = async ({
 			values: pubValues,
 		});
 
-		const vals = Object.entries(processedVals).flatMap(([slug, value]) => ({
-			slug,
-			value,
-		}));
+		const normalizedValues = normalizePubValues(processedVals);
 
 		const pubValuesWithSchemaNameAndFieldId = await validatePubValues({
```
allisonking, line 1118: I thought I wanted to update `updatePub` to also be able to take an array of related pubs, i.e. `slug: [{value: 'role', relatedPubId: 'xxx'}]` which `createPub` has. But now I'm not so sure, since this function has `includeRelations: false`, and even if I made it `true` updating the values on conflict gets an error `ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time`

So I can create a pub with pub value `{slug: 'croccroc:author', value: '', relatedPubId: 'xxx'}`, but I can't update the pub values to add another pub value, i.e. `{slug: 'croccroc:author', value: '', relatedPubId: 'yyy'}`. I think this is because the insert into `pub_values` has to be more clever, maybe in the `onConflict` part??

core/app/components/forms/FormElement.tsx
```diff
@@ -175,6 +176,18 @@ export const FormElement = ({
 		);
 	}
 
+	if (element.isRelation) {
+		const valueComponent = input;
```
allisonking, line 180: I had a small hope this would Just Work, but it doesn't :) We want to render an "wrapper" component and an "inner" component
image

The 'wrapper' component is the 'Pub Relations' box, and the 'inner' component is the popover 'Role' in this screenshot. Without taking any of the changes in this PR into account, `FormElement` would return the contents of 'Role', which is what we want, just in the popover instead, so I stashed that as `valueComponent` here. Then render the 'wrapper' as `RelatedPubsElement`. 

This doesn't actually work though because the `value` of the form is going to be an array now—something like `{'croccroc:author': [ {value: 'admin', relatedPubId: 'xxx'} ]}` so currently we end up rendering 
image. 

The only way I can think of to get this to work is to render a separate component inside `RelatedPubsElement` based on the `CoreSchemaType` that is a lot like `FormElement.tsx` but doesn't make all the same assumptions, and might not use i.e. `TextInputElement.tsx` since that tries to get the field value directly when really we need to index into the array.

maybe I'm missing a simpler way to do this though??

core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
```diff
@@ -61,14 +61,17 @@ const componentInfo: Record<InputComponent, SchemaComponentData> = {
...
-				placeholder={
-					element.schemaName === CoreSchemaType.String ? "For short text" : "For numbers"
-				}
-				type={element.schemaName === CoreSchemaType.String ? "text" : "number"}
-			/>
-		),
+		demoComponent: ({ element }) => {
+			const isText =
+				element.schemaName === CoreSchemaType.String ||
+				element.schemaName === CoreSchemaType.Email;
```
allisonking, line 67: Email was showing up as a number component in the demo component part

core/app/components/FormBuilder/FormBuilder.tsx
```diff
@@ -95,49 +96,12 @@ const elementPanelTitles: Record<PanelState["state"], string> = {
 	editingButton: "Edit Submission Button",
 };
 
-const PanelHeader = ({ state }: { state: PanelState["state"] }) => {
```
allisonking, line 98: I moved all the side panel stuff to another component so I could reuse it in the form fill page. I think this would be better off using a drawer or maybe even the new side bar component from shadcn, but for now I reused what we had to keep things consistent

core/app/components/forms/FormElement.tsx
```diff
@@ -1,48 +1,24 @@
-import { defaultComponent } from "schemas";
```
allisonking, line 1: Refactored this to be a wrapper around `PubFieldFormElement` which now takes a lot of the inside of this function. This is because I needed to reuse the parts in `PubFieldFormElement` to render the inner value component:

image

core/app/components/forms/PubFieldFormElement.tsx
```diff
@@ -0,0 +1,172 @@
+import { defaultComponent } from "schemas";
```
allisonking, line 1: pretty much directly pulled from `FormElement.tsx`

core/lib/server/pub.ts
```diff
@@ -1115,31 +1129,56 @@ export const updatePub = async ({
...
-						.doUpdateSet((eb) => ({
-							value: eb.ref("excluded.value"),
-							lastModifiedBy: eb.ref("excluded.lastModifiedBy"),
+		// Separate into fields with relationships and those without
+		const [pubValuesWithRelations, pubValuesWithoutRelations] = partition(
+			pubValuesWithSchemaNameAndFieldId,
+			(pv) => pv.relatedPubId
+		);
+
+		if (pubValuesWithRelations.length) {
```
allisonking, line 1138: this might result in an inconsistent return—should I update `replacePubRelationsBySlug` to also return something like what the `.insertInto("pub_values")` below does and then combine them?

packages/ui/src/form.tsx
```diff
@@ -146,7 +146,9 @@ const FormMessage = React.forwardRef<
 			body = String(error.message);
 		}
 		if (Array.isArray(error)) {
-			body = error[0]?.message;
+			const firstErrorIndex = error.findIndex((e) => !!e);
+			const firstError = error[firstErrorIndex];
+			body = firstError?.message ?? `Error with value at index ${firstErrorIndex}`;
```
allisonking, line 151: this isn't a great error message, but I wasn't sure a better way to get something to show up when there is an error with a value in one of the array fields. This shows up as 
image
 when the 'value' field is empty

core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
```diff
@@ -61,14 +61,15 @@ const componentInfo: Record<InputComponent, SchemaComponentData> = {
...
-		),
+		demoComponent: ({ element }) => {
+			const isNumber = element.schemaName === CoreSchemaType.Number;
+			return (
+				<Input
+					placeholder={isNumber ? "For numbers" : "For short text"}
+					type={isNumber ? "number" : "text"}
+				/>
+			);
+		},
```
allisonking, line 72: email and url were showing up as number input components in the demo portion

allisonking:
This is ready for review but MemberSelect doesn't work—if chosen, there'll be an error that I haven't put a lot of time into investigating yet, but I suspect it may be thorny! So I'm thinking I'll split that out into a separate chunk of work.

core/app/components/forms/elements/RelatedPubsElement.tsx
```diff
@@ -0,0 +1,203 @@
...
+	valueComponentProps: PubFieldFormElementProps;
+	slug: string;
+}) => {
+	const { title, id } = pub;
+
+	return (
+		<div className="flex items-center justify-between rounded border border-l-[12px] border-l-emerald-100 p-3">
+			<div className="flex flex-col items-start gap-1 text-sm">
+				<span className="font-semibold">{title || id}</span>
+				<ConfigureRelatedValue {...valueComponentProps} slug={slug} />
```
tefkah, line 43: I think we should use the pubtitle here as well
```suggestion
				<span className="font-semibold">{getPubTitle(pub)}</span>
```

core/lib/server/pub.ts
tefkah: i think that if this works then just leave it, i'm currently rewriting a bunch of this anyway!

tefkah (approved):
Really really nice, great work! I left some suggestions about slightly nicer error displays for the values and using pubtitles in a bunch of places, if you incorporate those im happy to merge it!

allisonking:
thanks for the great improvements @tefkah ! going to merge this now, but flagging that not all of the value inputs work yet, especially member select. but those should be fixed by https://github.com/pubpub/platform/pull/926 and https://github.com/pubpub/platform/pull/924

PR #911, opened 2025-01-22
update: Additional drive action structuring and extraction

isTravis:
This addresses many (all?) of the remaining migration changes needed to support GDrive editing.

### DIFF
--- a/core/actions/googleDriveImport/run.ts
+++ b/core/actions/googleDriveImport/run.ts
@@ -34,7 +34,7 @@ export const run = defineRun<typeof action>(
 				throw new Error("Failed to retrieve data from Google Drive");
 			}
 			const formattedData = await formatDriveData(dataFromDrive, communitySlug);
-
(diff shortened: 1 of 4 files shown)


core/actions/googleDriveImport/run.ts
```diff
@@ -158,6 +157,15 @@ export const run = defineRun<typeof action>(
 					relations,
 				});
 			}
+			await updatePub({
```
isTravis, line 160: This is my first time using `updatePub` in an action. Am I missing anything obvious with how I setup the call?

kalilsn: nope, looks good to me. that looks like the right value format and passing on the `lastModifiedBy` like you did is correct

kalilsn (approved):
i haven't run this since i don't know how but i think it's fine to merge! and the tests look good!

PR #912, opened 2025-01-22
Remember the last visited community

### LINKED ISSUE #629: Track last visited community and automatically send user to it on visit to app.pubpub.org
## Motivation

So users in multiple communities don't spend time switching
So users in multiple communities don't get confused and make changes to a community they don't expect to be in

## Requirements

- Tracks the last community a user visited
- The next time they visit app.pubpub.org, takes them to that community rather than the first created community, which is (I think) currently the default

kalilsn:
## Issue(s) Resolved
Resolves #629 

Adds a cookie to keep track of the last visited community. We set it in the c/community layout file, and check it before our redirects on login, loading the login page when you're already logged in, and visiting `/`.

~It's a bit convoluted because you can only set cookies from a server action.~ If there's a more sensible way to do this I would love to know!


core/app/components/LastVisitedCommunity/SetLastVisited.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default function SetLastVisited({ communitySlug }: { communitySlug: string }) {
+	const runSetLastVisited = useServerAction(setLastVisited);
+
+	useEffect(() => {
+		runSetLastVisited(communitySlug);
+	}, [communitySlug, runSetLastVisited]);
+
+	return <></>;
+}
```
tefkah, line 16: I think doing this on the client is probably easier, since we don't need it during the initial render on the server (ie we dont need the cookie info after it's set). it's the same way the (new) sidebar does it to remember open state.

```suggestion
import { LAST_VISITED_COOKIE, LAST_VISITED_COOKIE_MAX_AGE } from "./constants";

export default function SetLastVisited({ communitySlug }: { communitySlug: string }) {
	useEffect(() => {
		document.cookie = `${LAST_VISITED_COOKIE}=${communitySlug}; path=/; max-age=${LAST_VISITED_COOKIE_MAX_AGE}`;
	}, [communitySlug]);

	return <></>;
}
```

core/app/components/LastVisitedCommunity/constants.ts
```diff
@@ -0,0 +1 @@
+export const LAST_VISITED_COOKIE = "lastVisitedCommunitySlug";
```
tefkah, line 1: i'd add a max age of 30 days to this, could maybe be longer


```suggestion
export const LAST_VISITED_COOKIE = "lastVisitedCommunitySlug";
export const LAST_VISITED_COOKIE_MAX_AGE = 60 * 60 * 24 * 30;
```

tefkah (changes requested):
functions great! i would change the cookie setting behavior to happen on the client though instead of through a server action, saves some interaction with our server and it's not really anything sensitive or important we'd need to compute on the server.

core/app/components/LastVisitedCommunity/SetLastVisited.tsx
tefkah: there is possibly some more elegant way of doing this, but this does seem to work.

tefkah: based on this code from the new sidebar

https://github.com/pubpub/platform/blob/8f04e3a3fec92e0c18611199874bdad02c7ed7d4/packages/ui/src/sidebar.tsx#L77-L91

tefkah (approved):
great!

PR #913, opened 2025-01-23
Use breadth first flattening order for stages

### LINKED ISSUE #888: Workflow display out of order when workflow has circular branch
## Gabe's Special Instructions
The end goal here is for the last step of any given workflow to display last — the order of branched stages are not particularly important.

Timebox this to 1 day, and if it feels impossible given that we are flattening a multi-dimensional structure, we can shelve it for now and come up with a better UI later.

1. Create a workflow with a circular branch in it (e.g., a branched stage that later connects to the main branch)
2. On the workflows page, the second branch of the tree will show at the bottom of the workflow, rather than below the first branch

The second branch of the tree to show below the first branch of the workflow, but above the place where the two workflows come back together

https://app.pubpub.org/c/gabestein/stages

Order is currently:
1. Draft
2. Publish
3. Moved
4. Published
5. Branch 1

Would expect order to be:
1. Draft
2. Publish
3. Moved
4. Branch 1
5. Published

kalilsn:
## Issue(s) Resolved
#888 

This PR fixes a few bugs with the rendering of stages on the stage page. First of all, it uses a breadth first flattening, which should produce a better feeling order in many cases (such as the one linked in the issue). It also no longer attempts to separate the stages into workflows (disconnected graphs) since we don't actually have any way to reference the workflows individually.

This PR also ensures that we only render each stage once. In cases where there were multiple root stages, we would previously end up rendering duplicate stages. So these stages would previously render "In Production" and "Published" twice.
image

Another bug that's fixed here is that biconnected graphs were _never_ rendered before since they have no root nodes. For example, the stages in the screenshot below wouldn't render at all.
image


3mcd (approved):
Great work!!

PR #919, opened 2025-02-03
feat: full text search for pubs

tefkah:
- **feat: rough fts**
- **feat: add search to api**
- **feat: search dialog**
- **fix: allow optional pubValues for getPubTitle**
- **fix: only log out explain of search query when debugging**
- **fix: add better tests**
- **fix: add title and description for command dialogue and keep previous data while fetching**
- **feat: add more sophisticated searching, slower though**
- **fix: slightly more efficient auth check for api routes**

### DIFF
--- a/core/app/components/SearchDialog.tsx
+++ b/core/app/components/SearchDialog.tsx
@@ -46,9 +46,13 @@ export function SearchDialog({ defaultOpen, onOpenChange, onPubSelect }: SearchD
 				params: { communitySlug: community.slug },
 			}),
 		enabled: debouncedQuery.length > 0,
+		placeholderData: (prev, prevQuery) => {
+			if (prevQuery?.state.error || prevQuery?.state.data?.status !== 200) {
+				return prevQuery?.state.data;
+			}
+			return prev;
+		},
 	});
-
-	console.log(results);
 
 	// CMD+K handler
 	useEffect(() => {
@@ -64,11 +68,14 @@ export function SearchDialog({ defaultOpen, onOpenChange, onPubSelect }: SearchD
 		return () => document.removeEventListener("keydown", down);
 	}, [onOpenChange]);
 
-	// console.log(results.body.map((pub) => pub.pubType.name));
-
 	const router = useRouter();
 	return (
-		<CommandDialog open={open} onOpenChange={setOpen}>
+		<CommandDialog
+			open={open}
+			onOpenChange={setOpen}
+			title="Search pubs"
+			description="Search pubs by title and content"
+		>
 			<Command shouldFilter={false}>
 				<CommandInput placeholder="Search pubs..." value={query} onValueChange={setQuery} />
 				<CommandList>
--- a/core/app/components/SearchDialog.tsx
+++ b/core/app/components/SearchDialog.tsx
@@ -1,0 +1,139 @@
+"use client";
+
+import { useCallback, useEffect, useState } from "react";
+import { useRouter } from "next/navigation";
+import { useQuery } from "@tanstack/react-query";
+import { useDebounce } from "use-debounce";
+
+import {
+	Command,
+	CommandDialog,
+	CommandEmpty,
+	CommandGroup,
+	CommandInput,
+	CommandItem,
+	CommandList,
+} from "ui/command";
+import { AlertCircle, Loader2 } from "ui/icon";
+
+import type { fullTextSearch } from "~/lib/server/pub";
+import { client } from "~/lib/api";
+import { getPubTitle } from "~/lib/pubs";
+import { useCommunity } from "./providers/CommunityProvider";
+
+interface SearchDialogProps {
+	defaultOpen?: boolean;
+	onOpenChange?: (open: boolean) => void;
+	onPubSelect?: (pub: any) => void;
+}
+
+export function SearchDialog({ defaultOpen, onOpenChange, onPubSelect }: SearchDialogProps) {
+	const [open, setOpen] = useState(defaultOpen);
+	const [query, setQuery] = useState("");
+	const [debouncedQuery] = useDebounce(query, 300);
+	const community = useCommunity();
+
+	const {
+		data: results = { status: 400, body: [] },
+		error,
+		isError,
+		isFetching,
+	} = useQuery({
+		queryKey: ["pubs", "search", community.id, debouncedQuery],
+		queryFn: () =>
+			client.pubs.search.query({
+				query: { query: debouncedQuery },
+				params: { communitySlug: community.slug },
+			}),
+		enabled: debouncedQuery.length > 0,
+	});
+
+	console.log(results);
+
+	// CMD+K handler
+	useEffect(() => {
+		const down = (e: KeyboardEvent) => {
+			if (e.key === "k" && (e.metaKey || e.ctrlKey)) {
+				e.preventDefault();
+				onOpenChange?.(true);
+				setOpen(true);
+			}
+		};
+
+		document.addEventListener("keydown", down);
+		return () => document.removeEventListener("keydown", down);
+	}, [onOpenChange]);
+
+	// console.log(results.body.map((pub) => pub.pubType.name));
+
+	const router = useRouter();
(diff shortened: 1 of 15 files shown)


core/app/components/SearchDialog.tsx
```diff
@@ -0,0 +1,146 @@
...
+			if (e.key === "k" && (e.metaKey || e.ctrlKey)) {
+				e.preventDefault();
+				onOpenChange?.(true);
+				setOpen(true);
+			}
+		};
+
+		document.addEventListener("keydown", down);
+		return () => document.removeEventListener("keydown", down);
+	}, [onOpenChange]);
```
tefkah, line 69: this may be done more elegantly, we should probably have some global thing for this, but for now this is fine i think

core/app/components/SearchDialog.tsx
```diff
@@ -0,0 +1,146 @@
...
+					{results.status === 200 && results.body.length === 0 && (
+						<CommandEmpty className="flex w-full flex-col items-center p-6">
+							{isFetching ? (
+								<Loader2 className="h-4 w-4 animate-spin" />
+							) : isError ? (
+								<AlertCircle className="h-4 w-4" />
+							) : (
+								"No results found"
+							)}
+						</CommandEmpty>
```
tefkah, line 91: not very elegant, lots of states not well represented

core/lib/pubs.ts
```diff
@@ -5,7 +5,7 @@ import type { GetPubsResult } from "./server";
 export type PubTitleProps = {
 	title?: string | null;
 	createdAt: Date;
-	values:
+	values?:
```
tefkah, line 8: we don't need the values for this technically

core/lib/server/pub.fts.db.test.ts
```diff
@@ -0,0 +1,218 @@
...
+				Tags: "Web, JavaScript, Programming",
+			},
+		},
+		{
+			pubType: "Basic Pub",
+			values: {
+				Title: "Database Design Patterns",
+				Description: "Essential patterns for database architecture",
+				Content: "Understanding SQL and NoSQL database design principles.",
+				Tags: "Database, SQL, Architecture",
```
tefkah, line 73: definitely wrote these myself

core/prisma/migrations/20250130165541_add_ts_vector_to_pub_values/migration.sql
```diff
@@ -0,0 +1,103 @@
+-- Create the search vector column
+ALTER TABLE "pubs" ADD COLUMN "searchVector" tsvector;
+
+-- CreateIndex
+CREATE INDEX "pubs_searchVector_idx" ON "pubs" USING GIN ("searchVector");
```
tefkah, line 5: this adds the column and the idx (note: we should add more indices)

core/prisma/migrations/20250130165541_add_ts_vector_to_pub_values/migration.sql
```diff
@@ -0,0 +1,103 @@
...
+                WHEN TG_OP = 'DELETE' OR tmp."value" IS NULL THEN NULL
+                ELSE tmp."value"
+            END AS new_title
+        FROM (
+            SELECT DISTINCT "pubId"
+            FROM inserted_updated_deleted_rows
+        ) AS affected
+        LEFT JOIN tmp_affected_pubs tmp ON tmp."pubId" = affected."pubId"
+        JOIN pubs ON pubs.id = affected."pubId"
+    )
```
tefkah, line 71: this is just slightly rewriting this as a CTE, st i can use that title in the search vector as well

https://github.com/pubpub/platform/blob/78aab116dd1c5c91393bcd1255f72d2e8cd2fd0b/core/prisma/migrations/20241219131700_fix_title_update_trigger/migration.sql#L46-L67

core/prisma/migrations/20250130165541_add_ts_vector_to_pub_values/migration.sql
```diff
@@ -0,0 +1,103 @@
...
+            SELECT 
+                setweight(to_tsvector('english', COALESCE(updates.new_title, '')), 'A') ||
+                setweight(to_tsvector('english', COALESCE(
+                    (SELECT string_agg(CAST(value #>> '{}' AS TEXT), ' ')
+                    FROM pub_values
+                    WHERE "pubId" = updates."pubId"), 
+                '')), 'B')
+        )
+    FROM updates
+    WHERE "pubs"."id" = updates."pubId";
```
tefkah, line 87: this, very naively, just sets the search vector as
```
(with A priority) title
(with B priority) `${pub.Values.join(' ')}` 
```
we could do this more sophisticatedly, ignore all the values we don't want to look up (maybe only add stringlike fields), join with values we do want to look up (add the user info instead of the memberId for instance), or give certain fields different priorities (rich text fields should probably have different priorities in the future)

but for now, i think this is fiiine

packages/contracts/src/resources/site.ts
```diff
@@ -352,6 +352,18 @@ export const zodErrorSchema = z.object({
...
+			search: {
+				method: "GET",
+				path: "/pubs/search",
+				summary: "Search for pubs",
+				description: "Search for pubs by title or value.",
+				query: z.object({
+					query: z.string(),
+				}),
+				responses: {
+					200: z.any().array(),
```
tefkah, line 364: todo: type this

packages/db/src/kanel/kanel-kysely-zod-compatibility-hook.cjs
```diff
@@ -9,6 +9,9 @@ const asZodObject = (schemaName) =>
 	`as z.ZodObject<{[K in keyof ${schemaName}]: z.Schema<${schemaName}[K]>}>`;
 
 const replaceSchemaCast = (/** @type {string} */ line) => {
+	// replace undefined with z.string() for tsvector columns
+	line = line.replace(/: undefined\./g, ": z.string().");
+
```
tefkah, line 14: by default, `kanel` would make the searchVector column `undefined.optional()`, which is nothing

packages/db/src/kanel/kanel-kysely-zod-compatibility-hook.cjs
```diff
@@ -106,10 +110,13 @@ function kanelKyselyZodCompatibilityPreRenderHook(outputAcc, instantiatedConfig)
 						 */
 						const imports = declaration.typeImports?.map(renameImportToUseSchemaSuffix);
 
+						if (declaration.name === "pubs") {
+							console.log(declaration);
+						}
+
```
tefkah, line 116: should remove this

packages/ui/src/command.tsx
```diff
@@ -17,20 +17,48 @@ const Command = React.forwardRef<
 	<CommandPrimitive
 		ref={ref}
 		className={cn(
-			"flex h-full w-full flex-col overflow-hidden rounded-md bg-white text-gray-950",
+			"flex h-full w-full flex-col overflow-hidden bg-white text-gray-950",
```
tefkah, line 20: interesting: for some reason this "rounded" caused the whole command list to be blurry

https://github.com/hngngn/shadcn-solid/issues/100

packages/ui/src/command.tsx
```diff
@@ -17,20 +17,48 @@ const Command = React.forwardRef<
...
+	showDescription,
+	...props
+}: CommandDialogProps) => {
 	return (
 		<Dialog {...props}>
 			<DialogContent className="overflow-hidden p-0 shadow-lg">
+				<DialogTitle className={cn(showTitle || "sr-only")}>{title}</DialogTitle>
+				<DialogDescription className={cn(showDescription || "sr-only")}>
+					{description}
+				</DialogDescription>
```
tefkah, line 61: these are necessary for accessibilty, otherwise you get a bunch of errors in the console/in dev

3mcd (approved):
Feels great to use. Thanks for all the code comments and great job on those tests too!

PR #920, opened 2025-02-03
Use breadth first flattening order for stages (take 2)

kalilsn:
## Issue(s) Resolved
Resolves https://github.com/pubpub/platform/issues/888 and fixes the bug that led to reverting in #917

I added one new condition to the body of the loop to prevent infinite loops when the graph has a cycle. I've also added a 500ms timeout to the sorting function after which it just returns the deduped but unsorted stage list, in case there are other bugs and to make sure the app behaves reasonably with communities that have an unreasonable number of stages.

### DIFF
--- a/core/lib/stages.ts
+++ b/core/lib/stages.ts
@@ -44,6 +44,13 @@ export function getOrderedStages<T extends CommunityStage>(stages: T[]): Array<T
 			// doesn't know that
 			break;
 		}
+
+		// If we've already visited this stage we're in a cycle, and shouldn't add its destinations.
+		// But we don't break, because there may be other stages already in the queue that should
+		// still be visited
+		if (orderedStages.has(stage)) {
+			continue;
+		}
 		orderedStages.add(stage);
 		stage.moveConstraints.forEach((destinationStage) =>
 			stagesQueue.push(stagesById[destinationStage.id])
--- a/core/lib/stages.ts
+++ b/core/lib/stages.ts
@@ -9,44 +9,6 @@ export type StagesById<T extends CommunityStage = CommunityStage> = {
 };
 
 /**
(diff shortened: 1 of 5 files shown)


core/lib/stages.ts
```diff
@@ -70,11 +34,47 @@ export function getStageWorkflows<T extends CommunityStage>(stages: T[]): Array<
...
+
+	const orderedStages = new Set<T>();
+	const stagesQueue: T[] = stageRoots;
+	const startTime = Date.now();
+	try {
+		// Breadth-first traversal of the graph(s)
+		while (stagesQueue.length > 0) {
+			if (Date.now() - startTime > 500) {
+				throw new Error("Stage sorting timed out");
+			}
```
kalilsn, line 46: I think this is the only reliable way to make sure this loop stops after a timeout. I'd rather trust some other kind of global timeout but AFAIK that's not really possible in node. I'd love to know if that's wrong and there's some other technique we can use to set an enforceable global timeout

allisonking (approved):
really appreciate the comments + tests!!

core/lib/stages.ts
tefkah: nah i think this is fine! i was initially thinking "couldn't you use a settimeout for this?", but that will never resolve before the while loop is done, something something microtask queue something something event loop

kalilsn: > something something microtask queue something something event loop

yeah this is what I meant to say. i am curious if using worker threads and/or some kind of process/fork based server would help with this in production? is that even feasible with next? i doubt this is the last time i'll accidentally add an infinite loop (or really any cpu intensive code) and i would rather not need to have timeout logic all over the app if possible!

PR #923, opened 2025-02-04
Sort pubs before limiting

### LINKED ISSUE #890: Pub stage list order changes unpredictably when moving pubs

1. Moved a pub from one stage to another on the Workflows page
2. When I did so, a pub not in view replaced the top pub on the 
3. When I did so, the display of the count of pubs in the stage stayed the same

(see video below)

1. The pub below the first one should move to the top of the stack
~2. The display of the count of pubs in the stage to decrease~

I also think ordering on the stage page may be a little out of whack in general — is it possible that it does not update `updatedAt` on the pub when you create a new relationship?

https://github.com/user-attachments/assets/d590a434-7bd8-41ab-ae23-76ce90e21f92

kalilsn:
## Issue(s) Resolved
#890  

This PR addresses the confusing sort order of pubs by duplicating the `orderBy` clause in our pubs query into the "root" cte where we apply the limit and offset, rather than applying the limit and offset before sorting the pubs!

It also adds a trigger that updates a pub's `updatedAt` whenever it's moved between stages, which will ensure that after a pub is moved on the workflows page it's always still in the top 3 pubs selected from its new stage. Finally I did some consolidation of the move query so I could use it in my test.


tefkah:
ahhh makes total sense! good catch!

core/lib/server/pub.sort.db.test.ts
```diff
@@ -0,0 +1,195 @@
...
+import { CoreSchemaType } from "db/public";
+
+import { mockServerCode } from "~/lib/__tests__/utils";
+import { createLastModifiedBy } from "../lastModifiedBy";
+
+const { createSeed, seedCommunity } = await import("~/prisma/seed/seedCommunity");
+
+const { createForEachMockedTransaction } = await mockServerCode();
+
+const { getTrx, rollback } = createForEachMockedTransaction();
```
kalilsn, line 15: I ended up moving this test to its own file, because it fails if it's run as part of another test. I think it has to do with the automatic rollback but I don't completely understand what's happening. Running the test with `it.only` or moving it to its own file make it pass, although it seems like that prevents the db from being cleaned up (including if I explicitly call rollback).

I think this test failing in a transaction makes sense, since the timestamps are all set using the transaction start time. But I'm not sure why moving it to its own file or using `it.only` prevents the transaction from working. Also is it ok to do this and just not clean up?

core/lib/server/stages.ts
```diff
@@ -165,3 +165,15 @@ export const getStages = ({ communityId, stageId, userId }: CommunityStageProps)
 };
 
 export type CommunityStage = AutoReturnType<typeof getStages>["executeTakeFirstOrThrow"];
+
+export const movePub = (pubId: PubsId, stageId: StagesId, trx = db) => {
+	return autoRevalidate(
```
allisonking, line 170: I can't quite remember the context around this, but just want to make sure it covers what we talked about [here](https://knowledgefutures.slack.com/archives/CMAQM0BNV/p1723148283839199) a long time ago. I think using the `.with` didn't delete properly, hence the two calls of 

```
await autoRevalidate(
  trx.deleteFrom("PubsInStages").where("PubsInStages.pubId", "=", pubId)
).execute();
await autoRevalidate(
  trx.insertInto("PubsInStages").values({ pubId, stageId })
).execute();
```
in updatePub?

kalilsn: I think the problem wasn't that the delete doesn't work, but that it doesn't necessarily happen before the insert. And I noticed that happening specifically when you update a pub, which attempts to move the pub even if the stage hasn't changed. So in that case, (`pubid`,`stageid`) is the same one being deleted and added and we get an error.  I added the `on conflict do nothing` clause below to handle that, and I haven't noticed any issues with the query in other places! It does successfully remove the pub from the stage it's in, and add a new entry for the stage it's moving to.

kalilsn: I don't think we actually did what that thread says either, since the [pub moved trigger](https://github.com/pubpub/v7/blob/cb0a574d96cb5aeaf11c6970f08d2f3511ed153c/core/prisma/migrations/20240411012752_create_trigger_on_pub_stage_join_table/migration.sql#L21-L22) wasn't updated. I don't really want to do it in this PR, but I'm happy to make that change in a followup PR, and at the same time get rid of the `PubsInStages` table and replace it with a `pubs.stage` column.

tefkah: very much pro adding a `pubs.stageId` column instead!

3mcd (approved):

PR #926, opened 2025-02-04
Member select refactor

allisonking:
## Issue(s) Resolved
Related to https://github.com/pubpub/platform/issues/791

Refactors the MemberSelect components to use our API to query for users rather than direct db calls from server components.

This is useful for https://github.com/pubpub/platform/pull/910 since the server component was having trouble rendering in the related pub value popover.

## Test Plan
Make sure all of the member selects still work as expected.


core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts
```diff
@@ -647,7 +647,8 @@ const handler = createNextHandler(
 				});
 
 				const users = await getSuggestedUsers({
-					communityId: community.id,
+					communityId: req.query.communityId,
```
Comment (allisonking) [inline core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts:650]: ❗ a subtle change that honestly I don't quite understand, but if I passed `community.id` (where `community` is from `checkAuthorization` above, this would sometimes be a different id than the community id! 

to reproduce:
1. revert this to `community.id`. add a console log for `req.query.communityId` and `community.id`
2. Go to a page that has a member select
3. Type a member's email i.e. `all@pubpub.org`
4. this route should be hit, but the two community id's are different! `req.query.communityId` is the right one, whereas `community.id` was returning an id that did not belong to any of the communities in my database, and I'm not sure why that is. I did notice it would return the write community id when just loading my browser, but not when I explicitly called the query via `client.users.search`

core/app/components/MemberSelect/MemberSelectAddUserButton.tsx
```diff
@@ -9,27 +9,33 @@ import { MemberSelectAddUserForm } from "./MemberSelectAddUserForm";
...
 	community: Communities;
 	email: string;
+	onUserAdded: () => void;
 };
 
 export const MemberSelectAddUserButton = (props: Props) => {
 	const [open, setOpen] = useState(false);
 
-	const onClick = useCallback((e: React.MouseEvent<HTMLButtonElement>) => {
-		e.preventDefault();
```
allisonking, line 18: I don't think we actually need these `e.preventDefault()`s if we specify the button type to be 'button'

core/app/components/MemberSelect/MemberSelectClientFetch.tsx
```diff
@@ -0,0 +1,114 @@
...
+		<MemberSelectClient
+			helpText={helpText}
+			community={community}
+			fieldLabel={fieldLabel}
+			fieldName={fieldName}
+			member={(user as MemberSelectUserWithMembership) ?? undefined}
+			users={users}
+			allowPubFieldSubstitution={allowPubFieldSubstitution}
+			onChange={setSearch}
+			onUserAdded={refetchUsers}
```
allisonking, line 111: I wasn't sure if there was a better way to trigger a refresh of the data here. this is for the case where a user is added through the form, and we need to tell the app to requery all the users so that the new user shows up in the list

core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx
```diff
@@ -30,6 +31,7 @@ export const ExternalFormWrapper = ({
 	const router = useRouter();
 	const pathname = usePathname();
 	const params = useSearchParams();
+	const [pubId] = useState(props.pub.id);
```
Comment (allisonking) [inline core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/ExternalFormWrapper.tsx:34]: sneaky bug where if you used the member select on a new form the server would make a new pubid but we were not caching it here (but are caching it in `PubEditorClient` so the pubids become out of sync and we make a pub with a different id than the one we navigate to

core/lib/server/user.ts
```diff
@@ -89,31 +90,41 @@ export const getUser = cache((userIdOrEmail: XOR<{ id: UsersId }, { email: strin
...
+						.selectFrom("community_memberships")
+						.selectAll("community_memberships")
+						.whereRef("community_memberships.userId", "=", "users.id")
+						.where("community_memberships.communityId", "=", communityId)
+				).as("member"),
+			])
+			.innerJoin("community_memberships", "users.id", "community_memberships.userId")
+			.where("community_memberships.id", "=", memberId)
+			.limit(limit);
+	}
```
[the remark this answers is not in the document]
3mcd: I'm in favor of removing `memberId` from the search and adding a new endpoint to get a member by ID.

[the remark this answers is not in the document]
allisonking: `/member/{memberId}`? `/user/member/{memberId}`? should it require a `communityId`? 

I think `/member/{memberId}` probably makes the most sense, though I don't know if the response payload should be the same structure as for a user (user info with nested `member`) or maybe it should be inside out (member info with nested `user`?)

allisonking:
Refactored endpoint, and merged main, confirmed related pubs with member values works now 🎉 

image
image

PR #927, opened 2025-02-05
Put stages with no sources or destinations at the end of the list

kalilsn:
## Issue(s) Resolved
The arcadia seed order is confusing right now.
For communities with a number of stages with no connections, these stages are sorted equivalent with root stages, which is a bit awkward. For example, in the test arcadia seed we have this stage graph:
image
And the sorted stage order produced has articles at the beginning, then every other stage before stage 2. This PR makes sure those other stages get sorted last.

A better improvement would probably be to first divide the stages into independent graphs, then separate those onto their own tabs. But that's hard and this was easy.

I also upgraded to node 22 lts because Object.groupBy isn't in 20.

## Test Plan

Look at the arcadia seed's all workflows page. It's better!


allisonking (approved):
looks great!

PR #932, opened 2025-02-07
Make first new versions when there's no migrated version

gabestein:
## Issue(s) Resolved
Previously, no new version would be created if there wasn't already a version from the legacy migration. This fixes that.

If there's no existing version, it creates one from the existing HTML.

## Test Plan
Tested with brand new pub and also with a legacy pub, and both seem to work.


isTravis (approved):
My brain is a bit mush at the moment, so I dont feel 100% confident I'm thinking of all edge cases, but it looks like a straightforward fix to an obvious (in retrospect) problem. 👍

PR #965, opened 2025-02-11
RFC: pubOp for creating and updating pubs more easily

tefkah:
## Issue(s) Resolved

Adds a new PubOp system for creating, updating and managing relationships between Pubs using a fluent-like API/

> [!NOTE]
> This PR is an RFC! Meaning I'm totally happy not merging it!
> I'm more interested in your general thoughts on this approach, please nitpick it to hell!

This PR introduces a new `PubOp` system that provides a fluent, builder-style API for managing Pub operations. The system handles complex scenarios like creating multiple related pubs, managing nested relationships, and handling pub value updates in a single transaction.

### API

The API follows a builder pattern, allowing for chaining operations:

```ts
const pub = await PubOp.create({
  communityId: community.id,
  pubTypeId: pubType.id,
  lastModifiedBy: createLastModifiedBy({ userId: user.id }),
})
  .set('title', 'My Pub')
  .relate('someRelation', 'relation value', relatedPubId)
  .execute();
```

#### All builders

Currently there's 3 kinds of operations you can do:

- `create`
- `update`
- `upsert`

For `create`, there's also an explicit `createWithId`.
For both `update` and `upsert`, there's an `updateByValue` and an `upsertByValue`. These allow you to, instead of selecting a pub by id, select a pub by a specific value, eg a google drive id.

#### Basic capabilities

Currently, the `PubOp` allows you to do a bunch of things, here's a small table that shows what you can do with each of the operations:

| Capability                      | Create | Update | Upsert |
| ------------------------------- | ------ | ------ | ------ |
| Set values                      | ✓      | ✓      | ✓      |
| Connect/create related pubs     | ✓      | ✓      | ✓      |
| Set stage/move pub              | ✓      | ✓      | ✓      |
| Explicitly disconnect relations | -      | ✓      | -      |
| Explicitly disconnect values    | -      | ✓      | -      |
| Override existing values        | -      | ✓      | ✓      |
| Replace existing relations      | -      | ✓      | ✓      |

##### Set values

```ts
const pub = await PubOp.createWithId(pubId, {
  /*...*/
})
  .set('community-slug:title', 'My Pub')
  .execute();
```

##### Set many values in one go

```ts
const pub = await PubOp.upsert(pubId, {
  /*...*/
})
  .set({
    'community-slug:title': 'My Pub',
    'community-slug:description': 'My Pub description',
  })
  .execute();
```

##### Relate pubs

```ts
const pub = await PubOp.create({
  /*...*/
})
  .relate('community-slug:someRelation', 'relation value 1', relatedPubId1)
  .relate('community-slug:someRelation', 'relation value 2', relatedPubId2)
  .execute();
```

##### Relate multiple pubs in one go

```ts
const pub = await PubOp.create({
  /*...*/
})
  .relate('community-slug:someRelation', [
    {
      value: 'relation value 1',
      target: relatedPubId1,
    },
    {
      value: 'relation value 2',
      target: relatedPubId2,
    },
  ])
  .execute();
```

##### Nested relate

You can nest pub operations, allowing you to create, relate, and/or update multiple pubs in a single operation:

```ts
const pub = await PubOp.upsert(pubId, {
  /*...*/
})
  .relate(
    'someRelation',
    'value',
    PubOp.upsert(relatedPubId, {
      /*...*/
    })
      .set('title', 'Related Pub')
      .relate(
        'anotherRelation',
        'value2',
        PubOp.upsert(relatedPubId2, {
          /*...*/
        }).set('title', 'Related Pub 2')
      )
  )
  .execute();
```

If you don't want to keep repeating `{ communityId: ..., lastModifiedBy: ... }`, you also define a function instead
This just returns a new `PubOp` with `communityId` etc already set.

You'll still need to provide the `pubTypeId` for `upsert` and `create` operations.

```ts
const pub = await PubOp.create({
  communityId,
  pubTypeId,
  lastModifiedBy: createLastModifiedBy({ userId }),
})
  .set('community-slug:title', 'Community Title')
  .relate('community-slug:someRelation', 'relation value', (pubOp) =>
    pubOp
      .create({
        pubTypeId,
      })
      .set('pub-type-slug:title', 'Nested Pub')
  )
  .execute();
```

## Test Plan

1. Look at the tests in
https://github.com/pubpub/platform/blob/64ebece2eab27b80e4c1b7d8c3c6a988d863f193/core/lib/server/pub-op.db.test.ts#L1-L1329

2. Write one extra test to get a feel for it (don't need to push it, but would be appreciated!)

## Notes

> [!NOTE]
> This new API isn't used anywhere yet, that comes later!

### DIFF
--- a/core/lib/__tests__/matchers.ts
+++ b/core/lib/__tests__/matchers.ts
@@ -1,0 +1,67 @@
+import { expect } from "vitest";
+
+import type { ProcessedPub } from "contracts";
+import type { PubsId } from "db/public";
+
+import type { db } from "~/kysely/database";
+
+expect.extend({
+	async toExist(received: PubsId | ProcessedPub, expected?: typeof db) {
+		if (typeof received !== "string") {
+			throw new Error("toExist() can only be called with a PubsId");
+		}
+		const { getPlainPub } = await import("../server/pub");
+
+		const pub = await getPlainPub(received, expected).executeTakeFirst();
+		const pass = Boolean(pub && pub.id === received);
+		const { isNot } = this;
+
+		return {
+			pass,
+			message: () =>
+				pass
+					? `Expected pub with ID ${received} ${isNot ? "not" : ""} to exist, and it does ${isNot ? "not" : ""}`
+					: `Expected pub with ID ${received} ${isNot ? "not to" : "to"} exist, but it does not`,
+		};
+	},
+
+	toHaveValues(
+		received: PubsId | ProcessedPub,
+		expected: Partial<ProcessedPub["values"][number]>[]
+	) {
+		if (typeof received === "string") {
+			throw new Error("toHaveValues() can only be called with a ProcessedPub");
+		}
+
+		const pub = received;
+		const sortedPubValues = [...pub.values].sort((a, b) =>
+			(a.value as string).localeCompare(b.value as string)
+		);
+
+		const expectedLength = expected.length;
+		const receivedLength = sortedPubValues.length;
+
+		const isNot = this.isNot;
+		if (!isNot && !this.equals(expectedLength, receivedLength)) {
+			return {
+				pass: false,
+				message: () =>
+					`Expected pub to have ${expectedLength} values, but it has ${receivedLength}`,
+			};
+		}
+
+		// equiv. to .toMatchObject
+		const pass = this.equals(sortedPubValues, expected, [
+			this.utils.iterableEquality,
+			this.utils.subsetEquality,
+		]);
+
+		return {
+			pass,
+			message: () =>
+				pass
+					? `Expected pub ${isNot ? "not" : ""} to have values ${JSON.stringify(expected)}, and it does ${isNot ? "not" : ""}`
+					: `Expected pub ${isNot ? "not to" : "to"} match values ${this.utils.diff(sortedPubValues, expected)}`,
+		};
+	},
+});
--- a/core/lib/__tests__/matchers.ts
+++ b/core/lib/__tests__/matchers.ts
@@ -17,10 +17,7 @@ const deepSortValues = (pub: ProcessedPub): ProcessedPub => {
 };
 
 expect.extend({
-	async toExist(received: PubsId | ProcessedPub, expected?: typeof db) {
-		if (typeof received !== "string") {
-			throw new Error("toExist() can only be called with a PubsId");
-		}
+	async toExist(received: PubsId, expected?: typeof db) {
 		const { getPlainPub } = await import("../server/pub");
 
 		const pub = await getPlainPub(received, expected).executeTakeFirst();
@@ -36,14 +33,7 @@ expect.extend({
 		};
 	},
 
-	toHaveValues(
-		received: PubsId | ProcessedPub,
-		expected: Partial<ProcessedPub["values"][number]>[]
-	) {
-		if (typeof received === "string") {
-			throw new Error("toHaveValues() can only be called with a ProcessedPub");
-		}
-
+	toHaveValues(received: ProcessedPub, expected: Partial<ProcessedPub["values"][number]>[]) {
 		const pub = received;
 		const sortedPubValues = deepSortValues(pub);
 
--- a/core/lib/__tests__/matchers.ts
+++ b/core/lib/__tests__/matchers.ts
@@ -5,6 +5,17 @@ import type { PubsId } from "db/public";
 
 import type { db } from "~/kysely/database";
 
+const deepSortValues = (pub: ProcessedPub): ProcessedPub => {
+	pub.values
+		.sort((a, b) => (a.value as string).localeCompare(b.value as string))
+		.map((item) => ({
+			...item,
+			relatedPub: item.relatedPub?.values ? deepSortValues(item.relatedPub) : item.relatedPub,
+		}));
+
+	return pub;
+};
+
 expect.extend({
 	async toExist(received: PubsId | ProcessedPub, expected?: typeof db) {
 		if (typeof received !== "string") {
@@ -34,12 +45,10 @@ expect.extend({
 		}
 
 		const pub = received;
-		const sortedPubValues = [...pub.values].sort((a, b) =>
-			(a.value as string).localeCompare(b.value as string)
-		);
+		const sortedPubValues = deepSortValues(pub);
 
 		const expectedLength = expected.length;
-		const receivedLength = sortedPubValues.length;
+		const receivedLength = sortedPubValues.values.length;
 
 		const isNot = this.isNot;
 		if (!isNot && !this.equals(expectedLength, receivedLength)) {
@@ -51,7 +60,7 @@ expect.extend({
 		}
 
 		// equiv. to .toMatchObject
-		const pass = this.equals(sortedPubValues, expected, [
+		const pass = this.equals(sortedPubValues.values, expected, [
 			this.utils.iterableEquality,
 			this.utils.subsetEquality,
 		]);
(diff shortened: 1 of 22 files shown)


core/lib/__tests__/matchers.ts
tefkah, line 1: these are helper methods to make matching pubvalues easier.

eg instead of doing

```ts
expect(pub.values).toHaveLength(3);
pub.values.sort(/* sorting bc you want the order of the values to be stable over different tests */)
expect(pub.values).toMatchObject([ 
	...
]);
```

you just do

```ts
expect(pub).toHaveValues([...])
```

Similarly for 

```ts
expect(pubId).toExist()
```

just easier than manually looking up the pub yourself

core/lib/server/pub-op.db.test.ts
```diff
@@ -0,0 +1,1331 @@
...
+			//     /        \      |
+			//   B           C --> I
+			//   |         /  \
+			//   G -->  E      D
+			//               /  \
+			//             F     H
+			//                 /  \
+			//                K --> L
+
+			// create leaf nodes first
```
tefkah, line 826: this is by far the most complex test, and very curious whether you agree with whether this should be the behavior.
i spent a bit too long trying to figure this out haha

see my comment in `pub-op.ts` for a more in depth explanation about the "algorithm" for determining orphans

core/lib/server/pub-op.ts
```diff
@@ -0,0 +1,1359 @@
...
+	protected async executeWithTrx(trx: Transaction<Database>): Promise<PubsId> {
+		const operations = this.collectOperations();
+
+		await this.createAllPubs(trx, operations);
+		await this.processStages(trx, operations);
+		await this.processRelations(trx, operations);
+		await this.processValues(trx, operations);
+
+		return this.id;
+	}
```
tefkah, line 640: this is basically entirely what a `pubOp` can do. First create all pubs, add them to stages, do relation stuff, then do value stuff (those last two might be able to be done at the same time, or changed)

notably, all of these things happen at the same time for all pubs. so no more recursive calls, we first collect all operations and then do these things one by one. much more easy to reason about that way i think

core/lib/server/pub-op.ts
```diff
@@ -0,0 +1,1322 @@
...
+					)
+					.insertInto("PubsInStages")
+					.values(
+						nonNullStages.map((stageOp) => ({
+							pubId: stageOp.pubId,
+							stageId: stageOp.stageId,
+						}))
+					)
+			).execute();
+		}
```
tefkah, line 1130: hopefully thisll be a bit simpler once #967 lands!

core/prisma/seed/createSeed.ts
tefkah, line 1: seperated this out into a separate package so you can `createSeed` more easily

core/lib/server/pub-op.db.test.ts
tefkah, line 1: please look here for usage!
im really curious what you all think. please comment any ideas/feedback you have for better naming or anything else regarding the api! really interested in anything, especially "i hate this" or something!

core/lib/server/pub-capabilities.db.test.ts
```diff
@@ -2,12 +2,12 @@ import { describe, expect, it } from "vitest";
 
 import { CoreSchemaType, MemberRole } from "db/public";
 
-import type { Seed } from "~/prisma/seed/seedCommunity";
+import { createSeed } from "~/prisma/seed/createSeed";
```
allisonking, line 5: iirc importing `createSeed` up here did make data persist whereas importing just the `type { Seed }` didn't?

tefkah: yes, but i think that was bc they were both imported from `/seedCommunity`. I separated out the createSeed to a different file, which (I did not explicitly check) i think should fix that issue.
i do think using `createSeed` is slightly better as that preserves the type info once you pass it to `seedCommunity`, eg if you have `pubTypes: { 'SomepubType': ... } ` in your seed, the output of seedcommunity will have `pubTypes.SomePubType` as well

core/lib/__tests__/matchers.ts
```diff
@@ -0,0 +1,76 @@
...
+					: `Expected pub with ID ${received} to exist, but it ${pass ? "does" : "does not"}`,
+		};
+	},
+
+	toHaveValues(
+		received: PubsId | ProcessedPub,
+		expected: Partial<ProcessedPub["values"][number]>[]
+	) {
+		if (typeof received === "string") {
+			throw new Error("toHaveValues() can only be called with a ProcessedPub");
```
allisonking, line 44: how come `received: PubsId | ProcessedPub` up above instead of `received: ProcessedPub` then?

tefkah: good catch! it was left over bc i think i wanted to be able to pass in both in either case, and was having some trouble with the types. i later decided to just use `PubsId` for toExist and `ProcessedPub` for toHaveValues, but never updated it! I'll fix it

core/lib/server/pub-op.db.test.ts
```diff
@@ -0,0 +1,1329 @@
...
+
+		await expect(pub.id).toExist();
+
+		const pub2 = await PubOp.upsert(crypto.randomUUID() as PubsId, {
+			communityId: seededCommunity.community.id,
+			pubTypeId: seededCommunity.pubTypes["Basic Pub"].id,
+			lastModifiedBy: createLastModifiedBy("system"),
+		})
+			.relate(seededCommunity.pubFields["Some relation"].slug, "test relations value", pub.id)
+			.execute();
```
allisonking, line 178: this is quite nice! just so I understand, if pub2 were already created, would we also have to call `.upsert(...).relate(...)`? i.e. there's nothing like `PubOp.relate`

tefkah: Yeah correct!
You'd more likely do `PubOp.update(pub2Id).relate()` if you did not want to override the existing relations, but yes.

core/lib/server/pub-op.ts
```diff
@@ -0,0 +1,1313 @@
...
+			});
+		}
+
+		await this.cleanupOrphanedPubs(trx, Array.from(relationsToCheckForOrphans));
+	}
+
+	/**
+	 * remove pubs that have been disconnected/their value removed,
+	 * has `deleteOrphaned` set to true for their relevant relation operation,
+	 * AND have no other relations
```
tefkah, line 790: maybe it's a better idea to have this be a configuration setting on the `PubType` (or the pubField?)
like, for some relations this could be desireable, like versions and discussions in the current arcadia community.

but for others, like `tag` or something, you would not want this.

tefkah: in the UI i think we can just force the user to manually delete all the relations first, but i think it would be good in certain places to automatically decide this (import actions)

core/lib/server/pub-op.db.test.ts
```diff
@@ -0,0 +1,1329 @@
...
+	 * when creating nested PubOps
+	 */
+	it("should be able to do PubOps inline in a relate", async () => {
+		const pub = await PubOp.create({
+			communityId: seededCommunity.community.id,
+			pubTypeId: seededCommunity.pubTypes["Basic Pub"].id,
+			lastModifiedBy: createLastModifiedBy("system"),
+		})
+			.set(seededCommunity.pubFields["Title"].slug, "Test")
+			.relate(seededCommunity.pubFields["Some relation"].slug, "relation1", (pubOp) =>
```
allisonking, line 1198: that's really nice! 🙌

core/lib/server/pub-op.db.test.ts
allisonking: got it, thanks for the context, makes sense!

3mcd (approved):
The tests look like they cover most of the PubOp behavior and the logic for pub orphans seems sound. Looking forward to tomorrow's chat about the recursive CTE.

Issue #953, opened 2025-02-10
Button to create new related pub on pub details page


gabestein:
## Motivation

So users can create related pubs the way they can create children

So Unjournal can continue to use their process

## Requirements

- Adds a "Create Related Pub" button below the "Related Pubs" header, just like the current "Add new pub" button
- In popup, adds a field to select a relationship type that allows user to select from relationship fields on the "from" pub type
- In popup, render interface for adding the value of the the relationship, same as current value UI.
- When new pub is created, it should be added as a relation to the "parent" pub

## Acceptance Criteria

- Reuses "add related" design from forms to add relationship value in popup

## Mockups (if available)
![Image](https://github.com/user-attachments/assets/832a2cd9-9010-4de6-9fa4-e62e1f9fa5bc)

![Image](https://github.com/user-attachments/assets/6d70c5e8-96c8-4d8c-bf5b-718f05d81b22)

gabestein:
Right. Woof. I think for now leaving null is fine. Then maybe we could add it to...the form that loads, right?

gabestein:
Actually, more I think about it, more I think we should render a field for it if it's not null...

PR #973, opened 2025-02-12
Add related pub button

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/953

TODO
* [x] TS error on `slug` for `ConfigureRelatedValue`
* [ ] Adding a related pub currently replaces any current related value instead of appending—check if new update/upsert/create logic might address this?
* [x] Check file upload, member select, context editor and other tricky form values work ok
* [x] Refactor `editorSpecifiers`
* [x] Tests

## Test Plan
1. Add any related field (null should also work)
2. Add the field to a pub type
3. Create a pub with that pub type
4. On the created pub page, click the "Add related pub" button
5. You should get a modal that lets you choose a pubtype, relationship field, and relationship value
6. Click "Create pub"
7. This should bring you to the create pub page. Create a pub here
8. Go back to the pub created in step 3 and verify that the pub created in step 7 is now in the "Related pubs" table. You should also see the field it was related on and the value.

image


core/app/c/[communitySlug]/fields/page.tsx
```diff
@@ -59,7 +59,6 @@ export default async function Page(props: Props) {
 
 	const pubFields = await getPubFields({
 		communityId: community?.id as CommunitiesId,
-		includeRelations: true,
```
Comment (allisonking) [inline core/app/c/[communitySlug]/fields/page.tsx:62]: `includeRelations` was originally meant to hide relationship fields while we were still building them out. now that they are being used everywhere (I think) I've removed the param in favor of an `isRelated: boolean` which can filter for only relations (or non-relations)

core/app/c/[communitySlug]/pubs/[pubId]/page.tsx
```diff
@@ -232,13 +232,6 @@ export default async function Page(props: {
 			</div>
 			<div>
 				<h2 className="text-xl font-bold">Pub Contents</h2>
-				<p className="text-muted-foreground">
-					Use the "Add New Pub" button below to create a new pub and add it to this pub's
-					contents.
-				</p>
-			</div>
-			<div className="mb-2">
-				<CreatePubButton text="Add New Pub" communityId={community.id} parentId={pub.id} />
```
Comment (allisonking) [inline core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:241]: I know this ticket wasn't to remove the add children button, but since it's getting removed anyway I did it now so that `CreatePubButton` didn't have to keep working with children

core/app/components/pubs/InitialCreatePubForm.tsx
```diff
@@ -0,0 +1,253 @@
...
+		return `${path}?${urlSearchParams.toString()}`;
+	}, [path, searchParams]);
+
+	const closeForm = useCallback(() => {
+		router.replace(pathWithoutFormParam);
+	}, [pathWithoutFormParam]);
+
+	const onSubmit = async (values: FieldValues) => {
+		const related = values.relatedPub
+			? { ...values.relatedPub, value: JSON.stringify(values.relatedPub.value) }
```
allisonking, line 213: I'm not totally sure that this is safe—basically the problem is that our "create pub" flow is split across two views. The first view selects the pub type, and now also the relation definition. Then it redirects to the full create pub page, passing along its info via query parameters.

Is there any reason we wouldn't want the `value` field of the relation to be passed via query param to the other page? I think perhaps it could get long? or in the case of file upload, the path to the s3 file would end up in the query param i.e. `http://localhost:3000/c/croccroc/pubs/create?pubTypeId=c8be5cf7-3fd0-47f1-8ddc-eb954526efa3&relatedPubId=1a5a3582-01b6-45b1-8ad5-1bd6019289ab&slug=croccroc%3Arelated-file&value=%5B%7B%22id%22%3A%22relatedPub.value-test%2Ffile%2Ftxt-1d-1e-text%2Fplain-11-1722373001847%22%2C%22fileName%22%3A%22test-file.txt%22%2C%22fileSource%22%3A%22dashboard-relatedPub.value%22%2C%22fileType%22%3A%22text%2Fplain%22%2C%22fileSize%22%3A11%2C%22fileMeta%22%3A%7B%22relativePath%22%3Anull%2C%22name%22%3A%22test-file.txt%22%2C%22type%22%3A%22text%2Fplain%22%7D%2C%22fileUploadUrl%22%3A%22https%3A%2F%2Fs3.us-east-1.amazonaws.com%2Fassets.v7.pubpub.org%2F1a5a3582-01b6-45b1-8ad5-1bd6019289ab%2Ftest-file.txt%22%7D%5D`

allisonking: the alternative would be to store the value in something like local storage, or maybe a context?

core/app/components/pubs/InitialCreatePubForm.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+	const pathWithoutFormParam = useMemo(() => {
+		const urlSearchParams = new URLSearchParams(searchParams ?? undefined);
+		urlSearchParams.delete("create-pub-form");
+		return `${path}?${urlSearchParams.toString()}`;
+	}, [path, searchParams]);
+
+	const closeForm = useCallback(() => {
+		router.replace(pathWithoutFormParam);
+	}, [pathWithoutFormParam]);
```
allisonking, line 228: all of this is from the original `PubTypeFormClient` which I renamed since this does more than just pub type now

core/app/components/pubs/InitialCreatePubForm.tsx
```diff
@@ -0,0 +1,272 @@
...
+	if (!pubField) {
+		return null;
+	}
+	const element = makeFormElementDefFromPubFields([pubField])[0] as PubFieldElement;
+	return (
+		<ConfigureRelatedValue
+			element={element}
+			pubId={sourcePubId}
+			// @ts-ignore TODO: how best to fix this?
+			slug="relatedPub.value"
```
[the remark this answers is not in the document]
tefkah: ah tricky one! 

i'd try something like this in `RelatedPubsElement.tsx`

```tsx
type FormValue = {
	[slug: string]: FieldValue[];
};
type FormValueSingle = {
	[slug: string]: FieldValue;
};

const parseRelatedPubValueSlugError = (
	slug: RelatedPubValueSlug,
	formStateErrors: FieldErrors<FormValueSingle> | FieldErrors<FormValue>
) => {
	const [baseSlug, index] = slug.split(".");
	const indexNumber = index ? parseInt(index) : undefined;

	if (!indexNumber || isNaN(indexNumber)) {
		const baseError = (formStateErrors as FieldErrors<FormValueSingle>)[baseSlug];
		return baseError?.value;
	}

	const valueError = (formStateErrors as FieldErrors<FormValue>)[baseSlug]?.[indexNumber]?.value;

	return valueError;
};

// this is existing code
export const ConfigureRelatedValue = ({
	slug,
	element,
	onBlur,
	className,
	...props
}: PubFieldFormElementProps & {
	slug: RelatedPubValueSlug;
	onBlur?: () => void;
	className?: string;
}) => {
	const configLabel = "label" in element.config ? element.config.label : undefined;
	const label = configLabel || element.label || slug;

	const { watch, formState } = useFormContext<{
		[slug: string]: FieldValue[] | FieldValue;
	}>();
	const [isPopoverOpen, setPopoverIsOpen] = useState(false);
	const value = watch(slug);
	const showValue = value != null && value !== "";

// new 
	const valueError = parseRelatedPubValueSlugError(slug, formState.errors);
```

tefkah (commented):
it works well! other than that minor suggestion about when to show the related pubs button/table, the question about how to pass the value to the create form is very tricky!

i think i'd like to hear @3mcd or @kalilsn 's take on this! i won't press request changes as that'll block the pr! sorry, bit in a hurry as i wanted to get this in before i signed off!

3mcd:
@tefkah @allisonking I definitely like Thomas's suggestion to move the configuration of the related pub field value to the create form somehow, if it wouldn't be too much work. Maybe spend an hour or two on it, and if it doesn't work out in that time, we can stick with the URL serialization and create a card for it.

core/app/c/[communitySlug]/pubs/[pubId]/page.tsx
```diff
@@ -247,8 +240,14 @@ export default async function Page(props: {
...
 				<h2 className="mb-2 text-xl font-bold">Related Pubs</h2>
+				<CreatePubButton
+					text="Add Related Pub"
+					communityId={community.id}
+					relatedPub={{ pubId: pub.id, pubTypeId: pub.pubTypeId }}
+					className="w-fit"
+				/>
 				<RelatedPubsTable pub={pub} />
 			</div>
 		</div>
```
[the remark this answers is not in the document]
Comment (allisonking) [reply core/app/c/[communitySlug]/pubs/[pubId]/page.tsx:253]: yes that's a great idea!

allisonking:
Update—related value is now on the pub create page 
image

3mcd (approved):

PR #974, opened 2025-02-12
(self-hosting|2): add minio for local file uploads

### LINKED ISSUE #954: Add minio and basic self-hosting guidance

tefkah:
## Issue(s) Resolved

Mostly resolves #954 

This PR adds `minIO`, a self-hosted file server with an S3 compatible API, to our docker stack. 
This will allow users to host their files on their own server, instead of forcing them to set up an S3 bucket.
It also makes local development and testing a bit nicer for us, as we don't need to use our actual S3 credentials.

### Reorganization

I (to my own detriment) reorganized the way the env vars are set for the dev/test compose images a bit. Instead of setting them in the `docker-compose.yml`, I load them through the corresponding `.env.compose.(test|dev)`.
This makes it slightly easier to manage and override, but also makes it slightly less explicit and more easy to lose track of due to the overriding (i got rid of that as much as possible though).

Let me know if you have any thoughts on this regard. Env var organization is ... annoying.

## Test Plan

1. `pnpm dev:setup`
2. Get rid of any S3 config in `.env.local` if you have any
3. Go to `croccroc`, create a submission
4. Upload a file
5. Watch it succeed!

___

Convince yourself that this is a good test

https://github.com/pubpub/platform/blob/c25e50a5101bf97b751ad0f58d364d3bcc633d44/core/playwright/fileUpload.spec.ts#L1-L133


.github/workflows/e2e.yml
```diff
@@ -118,7 +114,7 @@ jobs:
                   INTEGRATION_TEST_HOST: localhost
 
             - name: Print container logs
-              if: failure()
+              if: ${{failure() || cancelled()}}
```
tefkah, line 117: if the tests are taking really long to fail and you cancel them, it's nice to be able to see the container logs

core/app/components/forms/FileUpload.tsx
```diff
@@ -21,7 +21,7 @@ export function FileUploadPreview({ files }: { files: FileUpload }) {
 									The file is <strong>{file.fileSize}</strong> bytes in size. Its
 									MIME type is <strong>{file.fileType}</strong>.
 								</p>
-								<Button variant="secondary">
+								<Button variant="secondary" asChild>
```
tefkah, line 24: was for the test, makes clicking this thing more consistent

core/lib/env/env.mjs
```diff
@@ -25,6 +25,7 @@ export const env = createEnv({
 		ASSETS_REGION: z.string(),
 		ASSETS_UPLOAD_KEY: z.string(),
 		ASSETS_UPLOAD_SECRET_KEY: z.string(),
+		ASSETS_STORAGE_ENDPOINT: z.string().url().optional(),
```
tefkah, line 28: the env var that determines whether or not you are using AWS or not

core/lib/server/assets.ts
```diff
@@ -12,11 +12,13 @@ export const generateSignedAssetUploadUrl = async (pubId: PubsId, fileName: stri
...
+		endpoint: env.ASSETS_STORAGE_ENDPOINT,
 		region: region,
 		credentials: {
 			accessKeyId: key,
 			secretAccessKey: secret,
 		},
+		forcePathStyle: !!env.ASSETS_STORAGE_ENDPOINT, // Required for MinIO
 	});
 	const command = new PutObjectCommand({
 		Bucket: bucket,
```
tefkah, line 24: these are all the changes that are necessary to make `minIO` (or any S3 compatible storage solution) work with our existing setup! isn't that great?

core/playwright.config.ts
```diff
@@ -17,6 +17,8 @@ export default defineConfig({
 	expect: {
 		timeout: process.env.CI ? 5_000 : 60_000,
 	},
+	// don't continue going after 3 tests have failed, that's too many
+	maxFailures: process.env.CI ? 3 : undefined,
```
tefkah, line 21: maybe controversial. since every single test was failing and retrying 3 times, i'd rather know after 3 tests fail that i'm cooked already

core/playwright/fixtures/test-assets/test-diagram.png
tefkah, line 1: isn't it beautiful.

this is just to have something to upload

docker-compose.base.yml
```diff
@@ -1,22 +1,37 @@
...
+            - app-network
+        entrypoint: >
+            /bin/sh -c '
+            /usr/bin/mc config host add myminio http://minio:9000 "$${MINIO_ROOT_USER}" "$${MINIO_ROOT_PASSWORD}";
+            /usr/bin/mc mb --ignore-existing myminio/"$${ASSETS_BUCKET_NAME}";
+            /usr/bin/mc anonymous set download myminio/"$${ASSETS_BUCKET_NAME}";
+            /usr/bin/mc admin user add myminio "$${ASSETS_UPLOAD_KEY}" "$${ASSETS_UPLOAD_SECRET_KEY}";
+            /usr/bin/mc admin policy attach myminio readwrite --user "$${ASSETS_UPLOAD_KEY}";
+            exit 0;
+            '
```
tefkah, line 28: it might be better to just have this in a script.

basically what this does:
- create the bucket
- allow public downloads for said bucket
- add a new user
- allow said user to upload

this sadly can't be done by just configuring some env vars for minio itself, very annoying, you need to use their `mc` client. I'm not sure this is the best way to go about doing this.

docker-compose.dev.yml
```diff
@@ -1,75 +1,57 @@
...
-        build: ./infrastructure/nginx
-        container_name: core-nginx
-        environment:
-            - NGINX_LISTEN_PORT=8080
-            - NGINX_PREFIX=/
-            - NGINX_UPSTREAM_HOST=core
-            - NGINX_UPSTREAM_PORT=3000
-            - OTEL_SERVICE_NAME=core.nginx
-        depends_on:
-            - core
```
tefkah, line 37: i removed the `core` and `jobs` build services for `.dev`. I never use them, I don't think any of you use them

package.json
tefkah, line 1: apologies for the format here, `prettier` seems to go back and forth whether it cares that this file has spaces or tabs. I think it makes sense to just make it the same as everywhere else

packages/schemas/src/formats.ts
```diff
@@ -16,7 +16,7 @@ export function IsEmail(value: string): boolean {
 }
 
 const Url =
-	/^(?:https?|wss?|ftp):\/\/(?:\S+(?::\S*)?@)?(?:(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z0-9\u{00a1}-\u{ffff}]+-)*[a-z0-9\u{00a1}-\u{ffff}]+)(?:\.(?:[a-z0-9\u{00a1}-\u{ffff}]+-)*[a-z0-9\u{00a1}-\u{ffff}]+)*(?:\.(?:[a-z\u{00a1}-\u{ffff}]{2,})))(?::\d{2,5})?(?:\/[^\s]*)?$/iu;
+	/^(?:https?|wss?|ftp):\/\/(?:\S+(?::\S*)?@)?(?:(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z0-9\u{00a1}-\u{ffff}]+-)*[a-z0-9\u{00a1}-\u{ffff}]+)(?:\.(?:[a-z0-9\u{00a1}-\u{ffff}]+-)*[a-z0-9\u{00a1}-\u{ffff}]+)*(?:\.(?:[a-z\u{00a1}-\u{ffff}]{2,}))|localhost)(?::\d{2,5})?(?:\/[^\s]*)?$/iu;
```
tefkah, line 19: very smol modification: allow `localhost` as a valid host to pass the `Url` scheme. This was necessary in order to allow local uploads, and also it's just a valid URL

packages/ui/src/customRenderers/fileUpload/fileUpload.tsx
```diff
@@ -40,27 +51,40 @@ const FileUpload = forwardRef(function FileUpload(props: FileUploadProps, ref) {
...
+					method: "PUT",
+					url: signedUrl,
+					headers: {
+						"content-type": file.type,
+					},
+				};
+			},
+		});
+	}, [props.upload]);
+
```
tefkah, line 87: this does not really change the logic (adds two extra checks but that's it)

this should be wrapped in an `effect` though

packages/ui/src/hover-card.tsx
```diff
@@ -7,22 +7,26 @@ import { cn } from "utils";
...
+			ref={ref}
+			align={align}
+			sideOffset={sideOffset}
+			className={cn(
+				"z-50 w-64 rounded-md border bg-popover p-4 text-popover-foreground shadow-md outline-none data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
+				className
+			)}
+			{...props}
+		/>
+	</HoverCardPortal>
```
tefkah, line 29: slightly hard to see, but the addition is of the `<HoverCardPortal>`. Previously the little card that pops up over uploaded files would get obscured by a lot of things, making it hard to click in tests. Now with the portal this is fixed

core/playwright.config.ts
3mcd: Good change!

docker-compose.base.yml
3mcd: Could you add this explanation as a comment to the yml file?

3mcd (approved):
Great work! The environment variable and docker compose changes all look good to me.

PR #975, opened 2025-02-12
fix: default date configs to empty; coerce date config fields to date

### LINKED ISSUE #904: Date fields should show as blank unless a date is actually selected

1. Created a new pub and saved it without filling out the date picker
2. No date shows on pub page
3. When I go to edit the pub, the date picker shows in the date field shows today's date rather than "select a date"
4. When I go to save, no date is saved, even though the field showed today's date
5. Date saves only if I go and select a date — if I want to select today's date, it requires that I select another date and then today's date agin

Also investigate why, e.g., datacite action does not fill in date automatically on manual run.

- When there is no date, I expect that field to show as blank or some unselected state.
- When the field shows a date and I click save, I expect it to save that date to the field

https://app.pubpub.org/c/unjournal/pubs/a05aaa9c-0e53-4c37-9762-b8638c6f8b7b/edit?pubTypeId=81d18691-3ac4-42c1-b55b-d3b2c065b9ad&parentId=5fb6acd7-dd18-4738-bd5b-bfd429546c3f&saveStatus=1736961613334

3mcd:
## Issue(s) Resolved

Resolves #904 

This PR removes the implicit default of `new Date()` from our datepicker component, meaning it will be empty by default as expected.

In addition it updates the Datacite action's date fields to use `z.coerce.date` instead of `z.date`, which will automatically coerce action config dates (represented as strings) to Date objects.

## Test Plan

Using the Arcadia dev community,
1. Add the DataCite action to a stage with pubs in it.
2. Configure the action. The "Publication Date" field should be empty be default. Do not modify the field.
3. Attempt to save the action config. It should fail, informing you that the Publication Date is required.
4. Set a value for the Publication Date, optionally specifying a pub field override.
5. Save the action. It should save successfully.
6. Run the action on a Pub. The action run should _not show any validation errors_. If you didn't set a pub field override in step 4, you should get an error telling you to.
7. Give the publication date an override value and run the action again. You should get the same result as from step 6.


allisonking (approved):
🎉

Issue #942, opened 2025-02-10
Italics button/support


gabestein:
## Motivation

So users can make text italic

## Requirements

- Select/unselect italic (em) at cursor/highlight

## Acceptance Criteria

- Markdown shortcuts (* or _)

Issue #941, opened 2025-02-10
Bold button/support


gabestein:
## Motivation

So users can make text bold.

## Requirements

- Select/unselect bold at cursor/highlight

## Acceptance Criteria

- Markdown shortcuts (** or __)

Issue #937, opened 2025-02-10
Sticky Menu Bar


gabestein:
## Motivation

So users can modify the contents of the editor no matter where they are in the field.

## Requirements

- Sticks to the top of a pub field
- Ability to apply and unapply styles at the point of the cursor or to highlighted text
- Shows what styles/blocks are currently applied at the cursor/highlight

## Acceptance Criteria

- Can primarily be based on Prosemirror examples/Legacy editor -- picking pieces we like and rewriting pieces we don't
- Can implement small subset of simple buttons to begin with and later be merged with other features (e.g. paragraph selector)

## Mockups (if available)
Can more or less match Legacy.

<img width="901" alt="Image" src="https://github.com/user-attachments/assets/c4774346-6809-4f6e-b342-a899f75d874a" />

allisonking:
can anyone link the files that might be relevant from legacy? maybe this one? https://github.com/pubpub/pubpub/blob/master/client/components/FormattingBar/FormattingBar.tsx

PR #977, opened 2025-02-13
Sticky editor menu

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/937
Closes https://github.com/pubpub/platform/issues/941
Closes https://github.com/pubpub/platform/issues/942

Adds a menu bar to the context editor which can stick to the top of an editor. The editor needs to have a `max-height` and `overflow: scroll` for the stickiness to kick in.

## Test Plan

1. Run `pnpm run storybook` in `packages/context-editor` and play with the editor!

Or, to see it in the app:

1. Add a rich text field to a form
2. Add enough content to the rich text field so that the max height kicks in and a scroll bar appears. 
3. The menu should stay sticky

demo of bold/italic commands

https://github.com/user-attachments/assets/d94b3695-bad9-4d14-b209-5dd2e40d36d7

it's sticky!

https://github.com/user-attachments/assets/847d09d3-c607-4f22-8147-5ca0de8d2b1d

## Notes

I based the bold and italic input rules off of the inline code one in legacy:
https://github.com/pubpub/pubpub/blob/4e39976b74a7204cb5703f2b5e5b2cf7f2ce3233/client/components/Editor/plugins/inputRules.ts#L45-L57

including the space at the end to trigger the rule. I thought the space was unnecessary at first, but I think in this case where the editor is applying rules as soon as the user types, it's necessary for `*italic*` vs `**bold**`, otherwise `**bold**` can never trigger. (It works for things like this GH editor because you have to navigate _to_ the preview view which can then figure out you mean to bold instead of seeing `**bold*` and immediately making it '*_bold_'


packages/context-editor/src/ContextEditor.tsx
```diff
@@ -96,14 +103,24 @@ function UnwrappedEditor(props: ContextEditorProps) {
...
+			plugins: [
+				...basePlugins(
+					baseSchema,
+					props,
+					panelPosition,
+					setPanelPosition,
+					suggestData,
+					setSuggestData
+				),
+				new Plugin({
```
allisonking, line 115: it's easier to include this plugin here rather than in `basePlugins` since this one needs access to `pluginViewFactory` which we get from the hook

packages/context-editor/src/utils/marks.ts
```diff
@@ -0,0 +1,10 @@
+import type { MarkType } from "prosemirror-model";
+import type { EditorState } from "prosemirror-state";
+
+export const markIsActive = (markType: MarkType, editorState: EditorState) => {
```
allisonking, line 4: took this from legacy

3mcd (approved):
Clean and straightforward. Great work!

PR #980, opened 2025-02-18
fix: pub create/edit form grow to container

### LINKED ISSUE #962: Make sure forms have minimum width
## Motivation
So users have a predictable editing interface

## Requirements
Right now, forms with rich content fields can be super-narrow, and then get wider as you type. It's weird. We should set a min width for all forms, and probably a fixed one if a rich text field is there (with, of course, variance on desktop and mobile).

Consult @ktyjscott on what forms should look like at larger resolutions.

Image

3mcd:
## Issue(s) Resolved

Resolves #962 

Quick fix to make the pub create/edit forms a bit wider.

## Test Plan

1. Open the pub create and edit forms and observe that they are now wider!

Screenshot 2025-02-18 at 12 36 27 PM

### DIFF
--- a/core/app/c/[communitySlug]/pubs/[pubId]/edit/page.tsx
+++ b/core/app/c/[communitySlug]/pubs/[pubId]/edit/page.tsx
@@ -130,7 +130,7 @@ export default async function Page(props: {
 			}
 		>
 			<div className="flex justify-center py-10">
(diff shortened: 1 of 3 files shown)


tefkah (approved):
looks good, but don't know why the tests are failing!

PR #982, opened 2025-02-19
(self-host|3): self-hostable platform!

### LINKED ISSUE #954: Add minio and basic self-hosting guidance

tefkah:
## Issue(s) Resolved

Resolves #954 

Adds proper self-hosting support!

1. Adds preconfigured `docker-compose.yml` in `self-host`
2. Adds sort of extensive instructions on how to get setup
3. Adds a script for generating an admin user (kind of finicky right now, requires a special mount until this PR is merged in)
4. Adds default Caddy config so users can reverse proxy the app/files easily

### Basic
Try to follow the instruction in `self-host/README.md` and get the app running at least.

### Caddy config

Once you have the app up and running, try to get Caddy to serve the app and files over https locally.

All in `self-host`

1. Copy the local test file
```sh
cp Caddyfile Caddyfile_bak
cp Caddyfile.test Caddyfile
```

2. Uncomment all this
https://github.com/pubpub/platform/blob/503022fff124b1358e561bc47541d273b758d394/self-host/docker-compose.yml#L85-L101

3. Configure local certificates
```sh
brew install mkcert
mkcert -install
```

4. Add `127.0.0.1 example.com` to your hosts (check whether it's not already there first)
```sh
sudo echo "127.0.0.1 example.com >> /etc/hosts"
```
(or just edit it like a normal person)

5. Restart the containers
```sh
docker compose down
docker compose up -d
```

6. Go to `https://example.com` in your browser
7. Click away warnings blahblah
8. Should be greeted by PubPub!

9. Try to upload some files! They should be uploaded to `example.com/assets/<bucket>/<etc..>`

### Desired feedback

My main concerns are about how easy this is to follow for a slightly technical person. The documentation can and should be leagues better, but is it good enough that someone slightly familiar with a `docker-compose.yml` could figure it out?

Also, what do you think of the default inclusion of `Caddy`? I think it's miles nicer than `nginx`, and it would be nice to provide a built in solution for exposing your files on the same domain as your main one (annoying to have to set up multiple sometimes), but I did not really do a good job yet explaining how it works in the docs, and it does make things slightly more complex. On the other hand, you will need to set up SSL certificates anyway, so it's good we offer (albeit maybe slightly clunky) a way to do it.

### What is missing

- [ ] Backups! Extremely important, we can't really call ourselves self-host ready if that's not a thing
- [ ] Guides on how to set up hosting/domains etc
- [ ] Guides on how to set up communities/default communities you can use. Currently it's very clunky.
- [ ] Better migration strat


self-host/create-admin.sh
tefkah, line 1: i'm bad at writing bash scripts. it would be nice if just had a cli for these things

self-host/docker-compose.yml
```diff
@@ -0,0 +1,153 @@
...
+            - path: .env
+              required: true
+        environment:
+            DATABASE_URL: postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
+        # FIXME: remove this once it's merged into main
+        volumes:
+            - ../core/prisma/create-admin-user.cts:/usr/src/app/core/prisma/create-admin-user.cts
+        command: ["pnpm", "--filter", "core", "migrate-docker"]
+        networks:
+            - app-network
```
tefkah, line 59: this is used to, well, run the migrations when the app starts up. It's also used to generate the admin user. 
I think we can probably get rid of this and put the important parts in `platform` instead, but for now I'm not sure what the best way of doing this would be.

self-host/docker-compose.yml
```diff
@@ -0,0 +1,153 @@
...
+                condition: service_started
+        image: ghcr.io/pubpub/platform-migrations:latest
+        env_file:
+            - path: .env
+              required: true
+        environment:
+            DATABASE_URL: postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
+        # FIXME: remove this once it's merged into main
+        volumes:
+            - ../core/prisma/create-admin-user.cts:/usr/src/app/core/prisma/create-admin-user.cts
```
tefkah, line 56: this should be removed before merging, it's just here for testing.

reason: the `core/prisma/create-admin-user.cts` does not exist yet on `platform-migrations:latest`, because i just now added it 😅
once this pr is merged to main all will be swell

tefkah:
Some feedback from @eastofesten: apparently the `POSTGRES_USER` needs to be `POSTGRES` in order for the migration script to work. will investigate

eastofesten:
also noting that step 4 for the caddy instructions didn't work (didn't say permission denied.) used `sudo -- sh -c "echo 127.0.0.1 example.com >> /etc/hosts"` instead to make it work.

3mcd (approved):
Thanks for going running through the steps with me!

PR #985, opened 2025-02-24
feat: allow detailed filtering of pubvalues and updatedAt/createdAt through the api

### LINKED ISSUE #950: Add filter by updatedAt and createdAt to Pub API
## Motivation

So developers can restrict their Pub API queries based on createdAt and updatedAt dates

## Requirements

- Filter Pubs returned by createdAt, with ability to specify before or after the date specified
- Filter Pubs returned by updatedAt, with ability to specify before or after the date specified
- Filter within a range of updatedAt/createdAt, e.g. pubs created after 01-01-2025 but before 01-31-2025.
- Combine updatedAt and createdAt queries.

## Acceptance Criteria

- Propose friendly-ish url params in REST API
- Update docs to reflect new options

tefkah:
## Issue(s) Resolved

Resolves #950 (and does some more)

Adds generic filtering of pubs through the api (and anywhere) by their pub values, and by `updatedAt` and `createdAt`.

The general pattern lifted straight from [Strapi](https://docs.strapi.io/dev-docs/api/rest/filters-locale-publication)

you can specify filter like so

```
/pubs?filters[arcadia:title][$containsi]=hello
```

thisll get you all pubs with `arcadia:title` matching `/hello/i`.

the syntax is 

simple queries
```
filters[<slug>][<operator]=... 
```

logic

```
filters[$and][0][<slug>][<operator>]=...&filters[$and][1][<other-slug>][<operator>]=...
```
this turns into
```ts
{ 
  $and: [
         {[slug1]: { $eq: 'something' }}, 
		 {[slug2]: { $eq: 'something else' }}
   ]
}
```
which gets you the pubs which has a value where `slug1='something` and which has a value where `slug2='something else'`

operators on the same level are automatically treated as "and" for a single value.

so
```ts
{ 
  $and: [
         {[slug1]: { $contains: 'something' }}, 
		 {[slug1]: { $containsi: 'else' }}
   ]
}
```
is equiv to

```ts
{ 
  [slug1]: { 
       $contains: 'something', 
	   $containsi: 'else'
  }
}
```

which is _also_ equiv to

```ts
{ 
  [slug1]: {
	   $and: { 
            $contains: 'something', 
	        $containsi: 'else'
    }
  }
}
```

As you may have noticed, there is a slight quirk: logical operators can be written either as an array, or as an object.  
This may be somewhat confusing, but basically logical operators can exist on two levels: 
1. "above" a slug, eg 
```ts
{ 
  $and: [
         {$or: [
              { [slug1]: $contains: 'something' }},
			  { [slug2]: $contains: 'something' }},
         ]}, 
		 {[slug1]: { $containsi: 'else' }}
   ]
}
```
2. "below" a slug, eg
```ts
{ 
  [slug1]: {
	   $and: { 
            $contains: 'something', 
	        $containsi: 'else'
    }
  }
}
```

in 1, which i call the "top level logical operators", you need to supply an array (except for `$not`, that one is always an object)

for 2, you need to supply an object.

maybe this should be simpler.

### Rough transformation

The query has the following lifecycle

1. The query enters the ts-rest route, and is parsed by ts-rest if supplied in `jsonQuery` format

https://github.com/pubpub/platform/blob/4d766c238979ff83d58a19561468ea3005e1ccbc/core/app/api/v0/c/%5BcommunitySlug%5D/site/%5B...ts-rest%5D/route.ts#L349

2. we manually parse the query if it is in "human readable" format using `qs` and our complex zod schema

https://github.com/pubpub/platform/blob/4d766c238979ff83d58a19561468ea3005e1ccbc/core/app/api/v0/c/%5BcommunitySlug%5D/site/%5B...ts-rest%5D/route.ts#L358

https://github.com/pubpub/platform/blob/4d766c238979ff83d58a19561468ea3005e1ccbc/core/app/api/v0/c/%5BcommunitySlug%5D/site/%5B...ts-rest%5D/route.ts#L233-L306

since everything is a string, we are coercing them to their proper formats, eg `2025-01-01` becomes a date, `12.1` a number, and `true` a boolean.

3. we validate the filter by checking that you are using the correct operators (that you don't use `contains`, which transforms into `like`, on eg a `number`

https://github.com/pubpub/platform/blob/61dba3c497f4ec5c83bb72a9706ff68878729b1e/core/lib/server/validate-filters.ts#L176-L227

4. we pass the filter to `getPubsWithRelatedValues`
5. the filters are applied in the `root_pubs_limited` cte 

https://github.com/pubpub/platform/blob/f2eab3c4373e9fbc27d5778c2d39ebd7f21e6d86/core/lib/server/pub.ts#L1724-L1726

Possibly step 3 is not necessary. It takes some time (as we need to fetch the schemas of the provided slugs), and the code is quite ugly. But i do think it provides a better experience if we explicitly tell you what went wrong, rather than the thing silently failing because you are `like`-ing a number, which will give unexpected results.

### What isn't working

- `isNull` checks don't work. i added them bc maybe we at some point go back to setting values to "null" if they are empty on a pub, but unlikely. 
- similarly, it's not easy to eg get pubs by the absence of some value. you can do `$not $contains ""` or something, but that only works for string values (and is super inefficient). im not sure it's a very common usecase, but let me know if you think differently.
- We likely would want to be able to filter on related pubs, eg "give me all pubs which have an author field where the authors have a firstname matching 'Jimothy'". it wouldn't be super easy to add, as i think we would basically need to do a separate recursive query to find these things, but if we limit it to one (or maybe) two levels deep i think we can do it inside the same query.

## Test Plan
1. try to study the tests
2. maybe experiment with the api, but not necessary, the tests should be sufficient

## Feedback

Mostly interested in feedback on the "api" (eg the shape of the filters etc). Also very interested in comments on the specific code, but it's just a bunch of transformations. If you have suggestions on how to improve it im all ears, but i don't expect you to comb over it in great detail!


core/lib/server/validate-filters.ts
tefkah, line 1: this is all a bit jank, but basically the idea is to find all the pubfields and the filters that are used on them, and then check whether those are "allowed", eg you should be able to do `$contains` on a `Number` pubfield

core/lib/server/pub-filters.db.test.ts
tefkah, line 1: there are a really large number of tests here because i kept running into little edge cases. maybe we can remove some of them to make it simpler, but i'm quite happy with these tests

core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts
```diff
@@ -226,6 +230,81 @@ const shouldReturnRepresentation = async () => {
...
+ * eg this does not fly
+ *  ```
+ * ?filters[community-slug:fieldName][$eq]=value
+ * ```
+ * but this does
+ *  ```
+ * ?filters=%7B%22%7B%22updatedAt%22%3A%20%7B%22%24gte%22%3A%20%222025-01-01%22%7D%2C%22field-slug%22%3A%20%7B%22%24eq%22%3A%20%22some-value%22%7D%7D`
+ * ```
+ *
+ * the latter is what a ts-rest client sends if `json-query: true`. we want to support both syntaxes.
```
Comment (tefkah) [inline core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts:247]: it's not so much wanting to support _both_ syntaxes, i just want to support the human readable one. turns out it's easier to support both.

core/package.json
```diff
@@ -117,6 +117,7 @@
 		"pg": "^8.11.3",
 		"prosemirror-markdown": "^1.12.0",
 		"prosemirror-model": "^1.24.1",
+		"qs": "^6.14.0",
```
tefkah, line 120: necessary for parsing nested queries, ts-rest does not do this for us

core/lib/server/pub-filters.ts
```diff
@@ -0,0 +1,308 @@
...
+				return eb.exists(
+					eb
+						.selectFrom("pub_values")
+						.innerJoin("pub_fields", "pub_fields.id", "pub_values.fieldId")
+						.select(eb.lit(1).as("exists_check"))
+						.where("pub_values.pubId", "=", eb.ref("pubs.id"))
+						.where("pub_fields.slug", "=", field)
+						.where((innerEb) => {
+							return applyFieldOperators(innerEb, operators, "value");
+						})
```
tefkah, line 303: you might think: thomas, why do this convoluted approach, why not just join the pubvalues to the pubs table and do a simple "where"?

wow, great point! here's the rough problem (a message i ended up sending kalil)

tldr: you run into problems with `AND` queries for multiple fields, eg give me all the pubs which have `title = 'hello'` and `abstract = 'goodbye'`. Bc (unless we do cross_join), the result of `pub join pub_values` are all the pub_values, basically. if you filter for pub_values where `(slug = title and value = hello) and (slug = description and value =  goodbye)`, you will match nothing (bc no pub-value has slugs which are both title and description).

therefore, we instead do these exists queries. i think they are fast enough, especially once we add some indices.

___
i basically added this to the root_pubs_limited in the big query
```ts
					.$if(Boolean(options?.filters), (qb) =>
						// TODO: maybe dedupe this
						qb
							.leftJoin("pub_values as pv", "pv.pubId", "pubs.id")
							.innerJoin("pub_fields as pf", "pf.id", "pv.fieldId")
							.where((eb) => applyFilters(eb, options!.filters!))
				)
```
where the applyFilters takes a filter (not really important what it looks like right now) and creates a big where clause
the problem lies in filters that are as such
```json
{
  $and: [
          { 'croccroc:title': { '$contains': 'hello' }}, 
          { 'croccroc:abstract': { '$contains': 'goodbye' }},
  ]
}
```

which means: give me all pubs where the title contains hello and the abstract contains goodbye.

how ive set it up now the resulting sql from applyFilters would look like (badly formatted but you get the point)
```sql
where ( slug = title and value like hello ) and ( slug = abstract and value like goodbye )
```
now, maybe you’ve already figured out the problem. say there’s only one pub and it fulfills the criteria and has only a title and an abstract

basically, that subquery up above gives
```json
[ 
  {
    id: 1,
    slug: 'croccroc:title'
    value: 'hello world',
  },
  {
    id: 1,
    slug: 'croccroc:abstract'
    value: 'goodbye cruel world',
  }
]
```

Problem: neither of these have **both** `slug = title` and `slug = abstract`, obviously.

___

core/lib/server/pub-filters.db.test.ts
```diff
@@ -0,0 +1,961 @@
...
+	{
+		title: "simple equality",
+		filter: {
+			[slug("title")]: { $eq: "Some title" },
+		},
+		querystring: `filters[${slug("title")}][$eq]=Some title`,
+		sql: `"slug" = $1 and "value" = $2`,
+		parameters: [slug("title"), '"Some title"'],
+		foundIds: [titleId],
+	},
```
tefkah, line 275: im running all these test cases through a number of tests, namely

- checking whether the querystring gets converted to the proper filter object
- whether said filter object passes parsing
- whether said filter object generates the correct `sql`
- whether said filter ends up fetching the correct pubs

core/lib/server/pub-filters.db.test.ts
```diff
@@ -0,0 +1,961 @@
...
+	// {
+	// 	title: "null and notNull operators",
+	// 	filter: {
+	// 		[slug("relation")]: { $null: true },
+	// 	},
+	// 	querystring: `filters[${slug("relation")}][$null]`,
+	// 	sql: `"slug" = $1 and "value" is null`,
+	// 	parameters: [slug("relation")],
+	// 	foundIds: [relationId],
+	// },
```
tefkah, line 490: i couldn't really get these to work well, but they might also be superfluous: i added them in case we ever ended up setting values to `null` to indicate that the value is "empty", but we are likely not doing that right.

core/lib/server/pub-filters.db.test.ts
```diff
@@ -0,0 +1,961 @@
...
+	it("rejects string operators on number fields", async () => {
+		const trx = getTrx();
+		const filter = {
+			[slug("number")]: { $containsi: "test" },
+		};
+
+		await expect(validateFilter(community.community.id, filter, trx)).rejects.toThrow(
+			`Operators [$containsi] are not valid for schema type Number of field ${communitySlug}:number`
+		);
+	});
```
tefkah, line 852: these are all "invalid" cases so i didn't add them to the big array above

core/.env.test
```diff
@@ -16,7 +16,7 @@ MAILGUN_SMTP_USERNAME="xxx"
 OTEL_SERVICE_NAME="pubpub-v7-dev" # should be shared across components but not environments
 HONEYCOMB_API_KEY="xxx"
 
-KYSELY_DEBUG="true"
+# KYSELY_DEBUG="true"
```
allisonking, line 19: oh I've always wondered what this does...

core/lib/mapping.ts
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Type-safe version of Object.entries()
+ * Returns an array of tuples containing key-value pairs with proper typing
+ * Includes all properties that exist in the object, even if undefined
+ */
+export function entries<
+	const T extends Record<string, unknown>,
+	const KeepUndefined extends boolean = false,
+>(
```
allisonking, line 11: ooh nice!

tefkah:
to make this easier to find later: i added `exists` queries after @allisonking 's review, added some tiny benchmarking code for later, and cleaned up the logic a bit. ill merge it now!

Issue #952, opened 2025-02-10
Remove children


gabestein:
## Motivation

So we can move to a fully relationship-based world

## Requirements

- Removes child table from the pub details page
- Updates email markdown tokens to allow users to specify a relation by field slug, and grabs the first created instance of that relation from another pub (e.g. `:value{field="unjournal:title" rel="unjournal:evaluations"}`

## Acceptance Criteria

- Remove parent association in prisma schema
- Write a corresponding migration that converts children to fields named for the pub type of the child — or, if easier, manually execute query for unjournal to convert children into corresponding relationship fields. E.g. all "evaluation" "children" should instead by related to the parent pub via an "evaluations" relationship field with null values.
- Update seed script to make sure there are no children
- Removes code for children from other places in the app, e.g. the API
- Remove `rel=parent` for all tokens that support it and replace with `rel="communitySlug:field"`. If the provided slug is not a relationship field, the parser should throw an error and the action should fail.

## Launch Requirements
- Inform @eastofesten when PR is accepted so that documentation and Unjournal's templates can be updated, and Unjournal can be notified prior to merge

3mcd:
We will also need to update our email/markdown tokens to let users specify a relation by field slug, e.g.

```
:value{field="unjournal:title" rel="parent"}
```

becomes

```
:value{field="unjournal:title" rel="unjournal:evaluations"}
```

where `unjournal:evaluations` is the relation through which the parent submission pub is related to the child evaluation pub.

3mcd:
@gabestein That feels like more trouble than it's worth as there's no general way to do that replacement (the script would need to know what the new relation fields are that are meant to replace the now obsolete parent/child hierarchy)

gabestein:
Thinking through this, this might break shortcodes...what if there's more than one "parent" pub that matches?

PR #986, opened 2025-02-24
(rm-children|1/4) rel="rel_field_slug" markdown directive attribute

3mcd:
## Issue(s) Resolved

Partial #952 

This PR adds support for a new `rel="relation_field_slug"` variant in markdown templates. This attribute provides the means to reference a pub that is related through a relation field that exists _on the related pub_.

For example, if an Evaluation is related to a Submission through a `community:sumissions` pub value belonging to the Submission, one could reference the Submission in the context of the Evaluation like so:

```
You have been invited to review :value{field="community:title" rel="community:submissions"}!
```

## Test Plan

1. Log in to the [review app](https://pr-986-relation-md-dire-ip-54-242-123-7.my.preview.run/c/arcadia-research) and visit the Arcadia Research community.
2. Add a new form for the Author pub type (the pub type doesn't really matter for the purposes of this test). Name it whatever you like. e.g: Screenshot 2025-03-05 at 2 49 26 PM
3. Delete all of the elements from the form. Add a single Paragraph element, and add a `:value` directive that references a Journal Article field (e.g. `arcadia-research:title`) through the `arcadia-research:contributors` relation field, e.g: Screenshot 2025-03-05 at 2 52 05 PM
4. Save the form. Set aside a link to the form's fill page using the "Copy link to live form" button at the top of the form builder.
5. Navigate to the Pubs page and find an Author pub. Navigate to its pub detail page and copy its pub ID from the URL.
6. Append the pub ID from step 5 to the form fill link from step 4, e.g. `https://pr-986-relation-md-dire-ip-54-242-123-7.my.preview.run/c/arcadia-research/public/forms/rel-attribute-test/fill?pubId=123456`. Navigate to this URL.
7. You should see the form fill page that renders the author's related Journal Article's pub value for the field you referenced in step 3.


core/lib/server/pub.ts
```diff
@@ -2325,3 +2325,43 @@ export const fullTextSearch = async (
 
 	return q.execute();
 };
+
+export const getExclusivelyRelatedPub = async (relatedPubId: PubsId, relationFieldSlug: string) => {
```
allisonking, line 2329: what does exclusively related mean here?

allisonking (commented):
the test steps worked great! I ran into something else though, which I think is likely not a bug and is just me being confused by how the refs work, but I tried adding another rel field on `arcadia-research:editors` and that seems to break things and I'm not sure why

the form builder:
https://pr-986-relation-md-dire-ip-54-242-123-7.my.preview.run/c/arcadia-research/forms/rel-test/edit

the form:
https://pr-986-relation-md-dire-ip-54-242-123-7.my.preview.run/c/arcadia-research/public/forms/rel-test/fill?pubId=d446ff22-58c6-4496-8870-3ed1027a7a72

(it's so nice we can link these now!)

core/lib/server/pub.ts
3mcd: I should add a comment! An "exclusively" related pub is not something that exists yet, but I imagine in the future we may want to have an option for relation fields that ensure only one pub value exists for that relationship on a given pub. i.e. an exclusive relationship

An example of this is a parent-child hierarchy, where a child may have only one parent

allisonking (approved):
huddled with eric, seems like the problem was that the author James McJimothy wasn't actually an editor on any journal articles!

PR #987, opened 2025-02-25
feat: add remove button to pub page

tefkah:
## Issue(s) Resolved

Can't remove pub from pub page, only from Pub list or stages page.

Just add the `PubRemoveButton` to `/c/[communitySlug]/pubs/[pubId]`

## Test Plan

1. Create new Pub
2. Go to Pub page
3. Press remove button

image


allisonking (approved):
works great! only thing is after deletion, we're left with an unauthorized page, maybe it'd be better to redirect back to the all pubs page?

image

tefkah:
oh good catch! that is kind of weird yeah, i will make it redirect

allisonking:
ah yeah I think that's ok for now! thanks for implementing it!!

PR #993, opened 2025-02-25
feat: allow for server uploads

tefkah:
## Issue(s) Resolved
Not being able to upload files from the server

While we have been able to upload files from the frontend for some time, uploading them from the server was not possible, because we were not exposing the S3 client we use for the frontend anywhere else.

This PR exposes that client, and more importantly exposes a convenience function `uploadFileToS3` which uses `@aws-sdk/lib-storage` for easy multipart file uploads, making file uploads much quicker.

## Test Plan
1. Convince yourself the added test makes sense
2. Make sure frontend uploads still work.

### DIFF
--- a/core/lib/server/assets.db.test.ts
+++ b/core/lib/server/assets.db.test.ts
@@ -1,0 +1,48 @@
+import fs from "fs/promises";
+
+import { beforeAll, describe, expect, it } from "vitest";
+
+import { mockServerCode } from "~/lib/__tests__/utils";
+import { env } from "../env/env.mjs";
+
+const { createForEachMockedTransaction } = await mockServerCode();
+
+const { getTrx } = createForEachMockedTransaction();
+
+beforeAll(async () => {
+	// check if minio is up
+
+	if (!env.ASSETS_STORAGE_ENDPOINT) {
+		throw new Error(
+			"You should only run this test against a local minio instance, not to prod S3"
+		);
+	}
+
+	const check = await fetch(env.ASSETS_STORAGE_ENDPOINT, {
+		method: "OPTIONS",
+	});
+
+	if (!check.ok) {
+		throw new Error(
+			"Minio is not running. Please setup the test environment properly by running `pnpm -w test:setup`"
+		);
+	}
+});
+
+describe("assets upload", () => {
+	it("should be able to upload a file to the minio bucket from the server", async () => {
+		const { uploadFileToS3 } = await import("./assets");
+
+		const file = await fs.readFile(new URL("./assets.ts", import.meta.url));
+
+		const url = await uploadFileToS3("test", "test.ts", file, {
+			contentType: "text/plain",
+		});
+
+		expect(url).toBeDefined();
+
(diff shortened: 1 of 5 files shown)


core/lib/server/assets.db.test.ts
```diff
@@ -0,0 +1,48 @@
...
+			contentType: "text/plain",
+		});
+
+		expect(url).toBeDefined();
+
+		const text = await fetch(url).then((res) => res.text());
+
+		expect(text).toEqual(file.toString("utf8"));
+	});
+});
```
tefkah, line 48: do we need more tests of eg larger files?

core/lib/server/assets.ts
```diff
@@ -20,9 +27,72 @@ export const generateSignedAssetUploadUrl = async (pubId: PubsId, fileName: stri
...
+
 	return await getSignedUrl(client, command, { expiresIn: 3600 });
 };
+
+/**
+ * Uploads a file to the S3 bucket using the S3 client directly
+ * @param id - id under which the file will be stored. eg for a pub, the pubId. for community assets like the logo, the communityId. for user avatars, the userId.
+ * @param fileName - name of the file to be stored
+ * @param fileData - the file data to upload (Buffer or Uint8Array)
+ * @param contentType - MIME type of the file (e.g., 'image/jpeg')
```
tefkah, line 51: whoops, this is technically outdated, i used to just have the mimetype as the last arg. ill change it

core/lib/server/assets.ts
```diff
@@ -20,9 +27,72 @@ export const generateSignedAssetUploadUrl = async (pubId: PubsId, fileName: stri
...
+		progressCallback ??
+			((progress) => {
+				logger.debug(progress);
+			})
+	);
+
+	const result = await parallelUploads3.done();
+
+	return result.Location!;
+};
```
tefkah, line 98: basically the same as our legacy implementation

https://github.com/pubpub/pubpub/blob/2dfe92a70d647278e512a91dfb060e4480739df2/server/utils/s3.ts#L93-L121

minus the ACL. I'm not sure what it is and why it would be necessary. we always seem to set it to `public-read` in legacy, but do make uploads with it? eg here
https://github.com/pubpub/pubpub/blob/2dfe92a70d647278e512a91dfb060e4480739df2/server/utils/s3.ts#L93-L121

upload seemed to work fine without it

allisonking (commented):
hmm the form file upload isn't working for me (the url localhost:9000/assets.v7.pubpub... looks wrong?)

image

will hop over to main now to see if it's happening for me there too!

allisonking:
nvm it's happening for me on main too.... is it only me?

allisonking:
> @allisonking it seems to work for me! do you have any overrides in `.env.local` for any `ASSETS_` env vars? Can you reach the minio server at `localhost:9001`?

ok don't hold this PR up for me then! I don't think I have any overrides in `.env.local`. and no I def don't have minio running, which I guess is the problem?—doing a `pnpm run dev:setup` now!

PR #995, opened 2025-02-25
feat(rm-children|2/4): remove most code related to pub children

3mcd:
## Issue(s) Resolved

Partial #952

This PR removes many references to pub children in our UI, server, APIs, and tests. It does not remove references to `parentId` or pub children in areas of the app that require the hierarchy to remain functional, e.g. the email action.

Removing the `parentId` column is handled by the final PR in this series, #1033 

## Test Plan

1. Open the [preview app](https://pr-995-rm-children-code-ip-3-82-60-71.my.preview.run/login?redirectTo=%2F).
2. Make sure everything works as expected, and there are no references to pub children in the app.


core/lib/server/pub.db.test.ts
```diff
@@ -793,8 +644,8 @@ describe("getPubsWithRelatedValuesAndChildren", () => {
 			}
 		);
 
-		// 3 root pubs, one of which has 1 child pub & 1 related pub, even though the limit is 3. This is correct behavior.
-		expect(pubs.length).toBe(5);
+		// 3 root pubs, one of which 1 related pub, even though the limit is 3. This is correct behavior.
+		expect(pubs.length).toBe(3);
```
allisonking, line 648: hmm why wouldn't this be `4`?

allisonking (approved):
image

👏 😌

core/lib/server/pub.db.test.ts
3mcd: I think the comment may be wrong? There are in fact only 3 pubs in the community, but I may need to update the test to preserve its intent... I'll look into it!

PR #996, opened 2025-02-25
Use mudder to rank form elements and related pub values

### LINKED ISSUE #963: Support relation order in DB and api
## Motivation

So developers can work with ordered relations
So users can order relations in a single field on a given pub (see #951)

## Requirements

- Queries for relationship fields across the app, including in the REST API should return related pubs in user-specified order and include order in return value

## Acceptance Criteria

- Add new column to pub values table for defining order
- Use some ordering algorithm (like the [one in Legacy](https://github.com/pubpub/pubpub/blob/78b6e47e06396f9c1f545f77f1af05405a3a76fa/utils/rank.ts)) that is faster than O(n) complexity.

kalilsn:
## Issue(s) Resolved
#963 
This PR:
- Adds rank column to both pub_values and form_elements tables
- Migrates existing form_elements and pub_values to have initial ranks
- Updates the FormBuilder to generate ranks as elements are created and moved
- Updates form selection query to order elements by rank
- Updates pub values queries to order values by rank
- Fixes a few other bugs in the form builder that I encountered during testing
	- Label changes are reflected in the form builder now
	- We now set a default config.label or config.relationshipConfig.label instead of `form_elements.label` for every element on form creation or when a new element is added to a form. This shouldn't change what gets rendered in the form itself because we also check the `form_elements.label` at that point, but this change means the default label will be populated in the element config form on the form builder. 
	- We now set the relationship component to `relationBlock` when creating a relationship element during initial form creation and when adding a new relationship form element to a form. Previously we were only setting that once the form element was edited.
	- The hover state on the delete/restore buttons on a form element is back!
	- Keyboard interactions work now! Click or use `tab` to focus an element, then press space to pick it up. Move it up and down with the arrow keys and press space to drop.
	- The "Cancel" button in the side panel component config form won't delete unsaved elements any more unless it's clicked immediately after they've been added
	- The "Save" button on the side panel won't be disabled immediately after adding a new element, so you can use that to exit the panel. Previously the only way was to make an edit or click the `x` because the cancel button removes the element.

My next PR will remove the order column from form_elements, and add the same dragging/reranking functionality to the pub_values inputs on form fill pages

## Test Plan

- Visit any form
- Reorder the elements, add some elements, remove some elements
- Save the form
- Make sure the order stays the same after you save
- Visit the live form
- Make sure the order matches the form builder


core/app/components/FormBuilder/ElementPanel/ButtonConfigurationForm.tsx
```diff
@@ -102,7 +103,11 @@ export const ButtonConfigurationForm = ({
 	const onSubmit = (values: z.infer<typeof schema>) => {
 		const index = buttonIndex === -1 ? numElements : buttonIndex;
 		update(index, {
-			order: null,
+			rank: mudder.base62.mudder(
+				elements[index - 1]?.rank ?? "",
+				elements[index + 1]?.rank ?? "",
+				1
+			)[0],
```
kalilsn, line 110: This generates a rank at the end of the form elements. It's not used right now since button elements aren't ordered in the form. But maybe the should/will be eventually? In any case, I did this because I wanted to make the rank column non-nullable. Should I use a constant string here instead of the mudder calculation?

core/app/components/FormBuilder/FormElement.tsx
```diff
@@ -63,22 +67,24 @@ export const FormElement = ({ element, index, isEditing, isDisabled }: FormEleme
 				removeElement(index);
 			}}
 		>
-			<Trash size={24} className="text-neutral-400 hover:text-red-500" />
+			<Trash size={24} className="!pointer-events-auto text-neutral-400 hover:text-red-500" />
```
kalilsn, line 70: @tefkah do you recall why we are setting `pointer-events: none` in the button component? it breaks the `:hover` pseudo-class so i ended up overriding it here.

https://github.com/pubpub/platform/commit/5eebab0b94054dd522f47e4a3081ca7f7590eb9d#diff-54e976c9e25a1c11aa59c088a36d31acd884abaeb450882b0e0b0e5076a83508R10

core/app/components/FormBuilder/FormElement.tsx
```diff
@@ -92,9 +98,10 @@ export const FormElement = ({ element, index, isEditing, isDisabled }: FormEleme
 						{restoreRemoveButton}
 						<Button
 							type="button"
+							aria-label="Edit field"
 							disabled={isDisabled || element.deleted}
 							variant="ghost"
-							className="invisible p-2 group-hover:visible"
+							className="invisible p-2 group-hover:visible group-focus:visible"
```
kalilsn, line 104: Now that keyboard interactions work, this makes sure the buttons are visible if you tab to a form element. I still haven't quite figured out how to make them appear in the tab order though

core/app/components/FormBuilder/actions.ts
```diff
@@ -54,13 +54,12 @@ export const saveForm = defineServerAction(async function saveForm(form: FormBui
 			} else if (!element.elementId) {
 				// Newly created elements have no elementId
 				acc.upserts.push(formElementsInitializerSchema.parse({ formId, ...element }));
-			} else if (element.updated || element.order !== index + 1) {
```
kalilsn, line 57: This was how we reordered elements before on the backend. Now the ranks are updated entirely on the frontend, so we can just remove this.

core/app/components/forms/elements/RelatedPubsElement.tsx
```diff
@@ -94,8 +94,11 @@ export const ConfigureRelatedValue = ({
 	onBlur?: () => void;
 	className?: string;
 }) => {
-	const configLabel = "label" in element.config ? element.config.label : undefined;
-	const label = configLabel || element.label || slug;
```
kalilsn, line 98: `slug` is not actually a field slug, it's the `name` for an `<input>` managed by react hook form, so it's not really appropriate for displaying to users even as a fallback

core/lib/server/form.ts
```diff
@@ -215,61 +207,23 @@ export const createFormInviteLink = async (props: FormInviteLinkProps) => {
...
 export const insertForm = (
-	pubTypeId: PubTypesId,
+	pubType: GetPubTypesResult[number],
 	name: string,
 	slug: string,
 	communityId: CommunitiesId,
 	isDefault: boolean,
 	trx = db
 ) => {
+	const ranks = mudder.base62.mudder(pubType.fields.length + 1);
```
kalilsn, line 218: Because we need to generate these ranks in js, I changed this function to expect a full pubType as an input, which means an additional query + roundtrip to the db before the insert. I think that's a fine tradeoff with how much simpler this looks now! But if we someday find this is too costly, mudder has no dependencies, so we could probably run it [inside postgres](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/PostgreSQL.Concepts.General.UpgradingPLv8.html).

tefkah:
> - The "Cancel" button in the side panel component config form won't delete unsaved elements any more unless it's clicked immediately after they've been added
> - The "Save" button on the side panel won't be disabled immediately after adding a new element, so you can use that to exit the panel. Previously the only way was to make an edit or click the x because the cancel button removes the element.

you are a hero

core/lib/server/form.ts
kalilsn: it's definitely not a problem for form inserts. but with pub inserts/updates i could see the roundtrip for finding the highest existing rank becoming a performance issue. mostly it's just that it sounds sick though 😄

core/app/components/FormBuilder/FormElement.tsx
kalilsn: yeah the tabindex not changing after reordering is why I kind of gave up on this for now. thanks for figuring out the other part though, I'll apply those changes!

core/prisma/exampleCommunitySeeds/arcadia.ts
```diff
@@ -128,7 +128,6 @@ export const seedArcadia = async (communityId?: CommunitiesId) => {
 			}
 
 			return {
-				value: null,
 				alsoAsChild: true,
 				pub,
 			};
```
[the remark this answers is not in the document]
kalilsn: I swear I fixed this, but I must have lost it in a rebase or something

core/app/components/pubs/PubEditor/helpers.ts
```diff
@@ -20,7 +21,7 @@ export function makeFormElementDefFromPubFields(
 					slug: field.slug,
 					schemaName: field.schemaName,
 					type: ElementType.pubfield,
-					order: index + 1,
+					rank: mudder.base62.numberToString(index + 1),
 					stageId: null,
```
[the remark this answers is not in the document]
kalilsn: I'm not opposed and I'll do this in my followup PR, but this is actually the only place I'm using `numberToString`. Do you think we need to wrap that function too? And I was thinking I would probably title the other function `findRanksBetween` since this is what we're doing almost every time:
```typescript
export const findRankBetween = ({ start = "", end = "", numberOfRanks = 1 }) =>
	mudder.base62.mudder(start, end, numberOfRanks);
```
I think that has some readability benefits (compared to `mudder.[...].mudder()` 😮‍💨) but I'm not sure it is actually going to help if we need to change it in the future, because I don't think we actually want to change most of these calls together. I think we're only going to want to change these if it looks like the rank strings are getting too long, and I think we're going to need to do that differently for different interfaces and different items. So I suspect we'd just end up adding the rest of the mudder parameters to our wrapper function.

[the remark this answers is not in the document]
tefkah: yeah true it is not really necessary for the `numberToString` one!

and yeah, it will probably have very limited benefit during a refactor, just makes me feel slightly nicer than having these specific calls everywhere, and this (very) slight hiding of the specific implementation is something im trying to learn to do more often, which is why i pointed it out here!

kalilsn:
Ok this is finally fixed up. There were some more complications with the form builder keyboard accessibility (that were affecting tests) and I also realized it was possible to do the pub value sorting entirely in the database, so I fixed that. I also had to fix the `toHaveValues` matcher to sort both the received values and the expected values, plus a few other test changes.

Issue #938, opened 2025-02-10
Paragraph type selector


gabestein:
## Motivation

So users can select between normal text, h1-6, code, blockquotes, and LaTeX in paragraphs

## Requirements

In menu bar, shows type of currently selected/cursored text and allows users to change it where legal
Allows users to use markdown shortcuts for headers (#, ##, etc.), code (```), latex/math ($, $$) and blockquotes (>)

## Mockups (if available)
![Image](https://github.com/user-attachments/assets/4bffa967-0d1e-4940-887a-e81fee7f0f89)

allisonking:
is there any reason not to copy/paste our `code` and `math` plugins from legacy to do this? implementing those seems fairly complicated, but I think I remember murmurs of maybe switching to different libraries?

PR #999, opened 2025-02-26
Paragraph selector

allisonking:
## Issue(s) Resolved
Paragraph, heading, and blockquote part of https://github.com/pubpub/platform/issues/938

Adds a paragraph/heading selector, as well as the ability to add blockquotes

## Test Plan

Run storybook in the context editor folder, and play around with making paragraphs/headings/blockquotes.

https://github.com/user-attachments/assets/cc976368-7e13-49b2-8408-35db00c97098

https://github.com/user-attachments/assets/600a427e-6a0b-47c6-a926-50d0779266ce

## Notes

The blockquote heading doesn't look great because it overlaps with the paragraph heading. not sure what should be done here? haven't looked into that code too much yet to see how we'd tweak the positioning
image


packages/context-editor/src/commands/blocks.ts
```diff
@@ -0,0 +1,89 @@
+import { lift, setBlockType, wrapIn } from "prosemirror-commands";
```
allisonking, line 1: pretty much lifted from legacy

packages/context-editor/src/commands/marks.ts
```diff
@@ -0,0 +1,32 @@
+import type { MarkType } from "prosemirror-model";
```
allisonking, line 1: also lifted from legacy—this pattern of creating toggles is much more flexible than what I originally had when it was only bold/italics

packages/context-editor/src/commands/types.ts
```diff
@@ -0,0 +1,48 @@
+import type { Mark, MarkType, Node, NodeType, Schema } from "prosemirror-model";
```
allisonking, line 1: lifted from legacy

packages/context-editor/src/commands/utils.ts
```diff
@@ -0,0 +1,27 @@
+import type { EditorState } from "prosemirror-state";
```
allisonking, line 1: lifted from legacy

packages/context-editor/src/components/MenuBar.tsx
```diff
@@ -1,65 +1,165 @@
...
+		command: heading6Toggle,
 	},
 ];
 
+const ParagraphDropdown = () => {
+	const { view } = usePluginViewContext();
+	const activeType = paragraphTypeItems.filter((item) => {
+		const { isActive } = item.command(view)(view.state);
+		return isActive;
+	})[0];
```
3mcd, line 99: Could be simplified like so:

```ts
paragraphTypeItems.find((item) => item.command(view)(view.state).isActive)
```

3mcd (approved):

PR #1000, opened 2025-02-26
Add math to context editor

allisonking:
## Issue(s) Resolved

Another part of https://github.com/pubpub/platform/issues/938

Adds math! This is a bit different from the legacy version—after talking to Katy, we decided to have an 'inline math' and a 'block math' button, more similar to 'inline code' and 'block code' in legacy, as opposed to the click-popup menu that lets you switch between the two. 

## Test Plan
Run the context editor storybook and try playing with math. hopefully it mostly acts the same as legacy (except toggling between the types is now done in the menu bar)

https://github.com/user-attachments/assets/74bf18e1-e5b3-43fb-bd94-73ca44aa867a

## Notes

I think we should write tests for these things soon, and playwright might be the way to go—but we'll have to do all the playwright setup in the context editor package probably?


packages/context-editor/src/commands/math.ts
```diff
@@ -0,0 +1,76 @@
...
+			swapNodeType.create({}, node.content),
+			true
+		);
+		dispatch(transaction);
+	}
+	return true;
+};
+
+const createMathToggle = (state: EditorState, type: MathType, dispatch?: Dispatch) => {
+	// Q: should this ever return false? (when can this func not be run?)
```
allisonking, line 35: in general these 'toggle' funcs should return false if that action cannot be performed wherever the cursor is. I didn't find a spot in legacy where you couldn't use the math button, so right now this func only ever returns true, but if there's a case where it shouldn't, we should add that here.

allisonking:
menu bar should more closely match [designs](https://www.figma.com/design/SqOTMElJn7fuPljfQL1gIY/PubPub-Platform?node-id=2382-5242&m=dev) now 
image

packages/context-editor/src/commands/math.ts
tefkah: hmm yeah you really can just put it everywhere haha 
image

the one place i've found it breaks stuff is in codeblocks (see above). even so, you are still _allowed_ to put it there, it'll just look bad. I think "looks bad/breaks your formatting" behavior is fine. if we want to be very strict we can disable the button in code blocks, but i don't think we _need_ to

tefkah: (note: the url there functions fine if you supply a custom url)

tefkah:
congrats on being PR #1000 !

packages/context-editor/src/components/MenuBar.tsx
```diff
@@ -128,35 +153,67 @@ const ParagraphDropdown = () => {
...
+		</div>
+	);
+};
+
+export const MenuBar = () => {
 	return (
 		<div
-			className="flex items-center rounded border bg-slate-50"
+			className="flex items-center rounded-t border bg-slate-50 p-4"
 			role="toolbar"
```
tefkah, line 193: Really great job with all the `MenuBar` changes, looks so nice now!

tiny tiny nitpick: the color in the designs is `gray-50` instead of `slate-50`

```suggestion
			className="flex items-center rounded-t border bg-gray-50 p-4"
```

tefkah (approved):
really great job @allisonking !

the comment about dynamic loading is mostly there for discussion. i think we should dynamically load as much as possible in the editor if we can, but it's not that big of a deal, definitely not to be implemented in this pr!

the other comments are just some tiny bugs/style changes which should not be hard to implement.

the only "big" question i had was about the gizmos for the inline math node. no clue how hard that will be to change, but would be good if it could!

otherwise again great job!

packages/context-editor/src/components/MenuBar.tsx
```diff
@@ -128,35 +153,67 @@ const ParagraphDropdown = () => {
...
+			<div className="h-6 w-px bg-gray-300" />
+		</div>
+	);
+};
+
+export const MenuBar = () => {
 	return (
 		<div
-			className="flex items-center rounded border bg-slate-50"
+			className="flex items-center rounded-t border bg-slate-50 p-4"
```
tefkah, line 192: Really great job with all the `MenuBar` changes, looks so nice now!

tiny tiny nitpick: the color in the designs is `gray-50` instead of `slate-50`

```suggestion
			className="flex items-center rounded-t border bg-gray-50 p-4"
```

tefkah, line 192: Really great job with all the `MenuBar` changes, looks so nice now!

tiny tiny nitpick: the color in the designs is `gray-50` instead of `slate-50`

```suggestion
			className="flex items-center rounded-t border bg-gray-50 p-4"
```

(got lost in the review bc i commented on the wrong line)

packages/context-editor/src/ContextEditor.tsx
```diff
@@ -19,6 +19,9 @@ import { baseSchema } from "./schemas";
 
 import "prosemirror-view/style/prosemirror.css";
 import "prosemirror-gapcursor/style/gapcursor.css";
+// For math
+import "@benrbray/prosemirror-math/dist/prosemirror-math.css";
+import "katex/dist/katex.min.css";
```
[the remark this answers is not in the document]
allisonking: oh I see, yes that makes a lot of sense! I imagine the code editor would also want something similar?

packages/context-editor/src/schemas/math.ts
```diff
@@ -0,0 +1,52 @@
...
+			getAttrs: (node) => {
+				return {
+					id: (node as Element).getAttribute("id"),
+					class: (node as Element).getAttribute("class"),
+				};
+			},
+		},
+	],
+	toDOM: () => ["math-display", { class: "math-node" }, 0],
+} satisfies NodeSpec;
```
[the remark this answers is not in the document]
allisonking: good find, this fixes my blockquote problem I meant to get back to too

before:
image


after:
image

packages/context-editor/src/schemas/index.ts
```diff
@@ -24,6 +21,7 @@ export const baseSchema = new Schema({
 		contextDoc,
 		contextAtom,
 		blockquote,
+		...math,
```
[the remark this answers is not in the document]
allisonking: great catch! I just added it—you're right that since it isn't a mark, it didn't automatically get the gizmo, so I added some logic to allow inline math to have it. we might have to do this with other non-mark inline things, but I'm not sure what those are yet, so will probably need to keep an eye on this as right now I just have it checking for inline math
image

packages/context-editor/src/ContextEditor.tsx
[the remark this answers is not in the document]
tefkah: yeah i think we had some bespoke system for the languages in v6. may have been one of the reasons for the fork

https://github.com/pubpub/pubpub/blob/1cd0cb207a4aaf877e2d6cc4ee6dfd4df61f29a0/client/components/Editor/plugins/code/languageLoaders.ts#L1-L17

packages/context-editor/src/schemas/index.ts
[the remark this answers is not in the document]
tefkah: oh awesome! glad that that wasn't that hard to add!

PR #1004, opened 2025-02-27
Google Action Updates

isTravis:
This adds:
- Better support for [10.abc] style references
- Asset re-uploading
- A couple other small fixes.


gabestein (approved):
Works for me. Amazing stuff!

PR #1005, opened 2025-02-27
chore: add pullpreview

3mcd:
## Issue(s) Resolved

N/A

This PR adds a `deploy-preview` step to our `on_pr` GitHub actions workflow that will, when labeled with a new `preview` label, deploy the containers produced by the upstream `build-all` step to temporary AWS Lightsail instances.

The changes are essentially all additive:
- A new step was added to `on_pr` that uses the [pullpreview action](https://pullpreview.com)
- A new docker compose file ([docker-compose.preview.yml](./docker-compose.preview.yml)) containing simple config changes was added. This file is not meant to be used on its own, and instead is [merged into the self-host docker compose file](https://pullpreview.com), so we're dogfooding our self hosting setup!
- pullpreview is configured to fetch images from our private ECR repository so we don't need to pollute the public GCR repository with review images

## Test Plan

Find the deployment close to the bottom of this page, open it, and confirm the app works!


docker-compose.preview.yml
```diff
@@ -0,0 +1,44 @@
...
+        depends_on:
+            platform:
+                condition: service_started
+        image: ghcr.io/pubpub/platform-migrations:latest
+        env_file: .env
+        environment:
+            DATABASE_URL: postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
+        command: ["pnpm", "--filter", "core", "reset"]
+        networks:
+            - app-network
```
tefkah, line 44: maybe an even simpler setup would be to just have the migrations container also run the reset, i don't think there's any real problem with that.

so you'd remove this, and to `platform-migrations` add (untested)

```yaml
command: ["pnpm", "--filter", "core", "migrate-docker", "&&", "pnpm", "--filter", "core", "reset"]
```

infrastructure/terraform/environments/global_aws/github_actions_iam.tf
```diff
@@ -128,4 +144,16 @@ resource "aws_iam_role_policy_attachment" "gha_attach_ecs" {
...
+
+resource "aws_iam_user_policy_attachment" "gha_user_attach_lightsail" {
+  user       = aws_iam_user.github_actions.name
+  policy_arn = aws_iam_policy.lightsail.arn
+}
+
+resource "aws_iam_user_policy_attachment" "gha_user_attach_ecr" {
+  user       = aws_iam_user.github_actions.name
+  policy_arn = aws_iam_policy.ecr.arn
+}
```
tefkah, line 159: yeah, and ideally i think we should only allow read access to ECR for that user right?

tefkah (approved):
other than the uploads not yet working, amazing!!!

PR #1014, opened 2025-03-04
Allow ordering of related pub values

### LINKED ISSUE #951: Ordering relations
## Motivation

So users can specify the order of relations they've added to specific fields

So developers can make use of relation orders in front-ends

## Requirements

- Ability to sort relationships on a per-field basis with drag-and-drop
- Relationship fields in API should return in specified order and include order in return value
- If no order is specified in relationship creation, default to last

### LINKED ISSUE #1077: adding a new related pub removes existing related pubs from table

1. A pub has a relationship field (ex: Articles) and so displays a related pub table.
2. User creates a related pub, which appears in the table
3. User adds new related pub of the same type
4. Upon saving, the related pub table only has the newest related pub, not both

Both related pubs would appear in the table

https://app.pubpub.org/c/oecs/pubs/c73299f9-53d2-4758-90b6-8900f9470fe6

### LINKED ISSUE #1079: Bug: saving pub via form removes it from stage

1. Edit a pub that is in a stage with a form
2. Click save
3. Refresh the page
4. Stage will change to "select a stage", and pub will be moved out of the stage

Pub should stay in the stage it was in, unless the stage was explicitly changed.

kalilsn:
## Issue(s) Resolved
Resolves #951
Resolves #1079 
Resolves #1077 
This PR adds the same ranking logic and UI that form elements have to related pub values. To accomplish that, I ended up having to rework the form saving logic, both on the client and the server. In the process, I also made a number of type changes:
- Tried to make the pub editor types a little stronger, since a lot of it was just using any
- Cast `pub_values.value` to `JsonValue` when selecting values in most queries
- Accept `Date` inputs to values update/insert queries, which is safe since we call `JSON.stringify(value)` before sending it to the db. This happens because we "hydrate" pub value dates from strings to `Date` objects in a number of places in the app, including in forms to allow them to pass validation.

I also fixed a few bugs:
	- We now only update the values that have actually been changed on a form, including inside of an array
	- Adding a related pub from the related pubs table won't overwrite existing pubs (#1077)
	- The initial create pub form where you select the pub type was limited to only 10 pub types, now it shows all of them
	- Saving a pub on the update form will no longer remove its stage (#1079)

## Test Plan
- Visit a pub with related values, like [James McJimothy](https://pr-1014-pubvalue-sorti-ip-54-234-137-78.my.preview.run/c/arcadia-research/pubs/9933b23c-6b9e-47b9-b1c6-b42ad5948fb1)
- Click "Add Related Pub" and create + save a pub, adding a value to the relationship at the top
- Return to  [James McJimothy](https://pr-1014-pubvalue-sorti-ip-54-234-137-78.my.preview.run/c/arcadia-research/pubs/9933b23c-6b9e-47b9-b1c6-b42ad5948fb1) and make sure you see the existing related pubs/values as well as the one you added
- Click "Update"
- Reorder, remove, and add some related values. Make sure nothing unexpected happens after you save (such as the values changing order or disappearing or the stage disappearing)!


core/app/components/pubs/PubEditor/actions.db.test.ts
```diff
@@ -110,11 +110,17 @@ describe("updatePub", () => {
...
+				values: [
+					{
+						fieldName: "Title",
+						rank: null,
+						relatedPubId: null,
+						schemaName: "String",
+						value: "new title",
+					},
+				],
+			},
```
kalilsn, line 123: This test had to be updated because an unfortunate side effect of using pub-op is that we now always call `getPubsWithRelatedValues` and return that after the update. We don't really need that because the pub update page updates from our autoRevalidated queries and we don't even look at the return value of the server action! So it would be nice to make the final query optional and/or figure out a way to use `returning` instead in pub-op eventually!

core/app/components/pubs/PubEditor/actions.ts
```diff
@@ -114,16 +120,38 @@ export const updatePub = defineServerAction(async function updatePub({
...
-		return result;
+		if (stageId) {
+			updateQuery.setStage(stageId);
+		}
+
+		const { values: processedVals }: { values: typeof pubValues } =
+			parseRichTextForPubFieldsAndRelatedPubs({
+				pubId: pubId,
+				values: pubValues as Record<string, JsonValue>,
+			});
```
kalilsn, line 137: it would probably make sense to put this call in pub-op.ts somewhere, but we'd need to make it work with normalized pubValues first

core/lib/server/stages.ts
```diff
@@ -172,8 +172,5 @@ export const movePub = (pubId: PubsId, stageId: StagesId, trx = db) => {
 			.with("leave_stage", (db) => db.deleteFrom("PubsInStages").where("pubId", "=", pubId))
 			.insertInto("PubsInStages")
 			.values([{ pubId, stageId }])
-			// Without this on conflict clause, the db errors if this function is called with the
-			// stageId the pub already belongs to
-			.onConflict((oc) => oc.doNothing())
```
kalilsn, line 177: This was the cause of the stage being removed when you saved a pub. The onConflict handling here was allowing the `leave_stage` cte to execute successfully, and the insert to quietly do nothing, which resulted in removing the pub's stage.

core/app/components/pubs/PubEditor/actions.ts
```diff
@@ -114,16 +120,38 @@ export const updatePub = defineServerAction(async function updatePub({
...
+				updateQuery.relate(slug, value, relatedPubId, {
+					replaceExisting: false,
+				});
+			} else {
+				updateQuery.set(slug, value);
+			}
+		}
+
+		for (const { slug, relatedPubId } of deleted) {
+			updateQuery.unrelate(slug, relatedPubId);
```
allisonking, line 151: so clean 🤩

core/app/components/forms/AddRelatedPubsPanel.tsx
```diff
@@ -5,12 +5,11 @@ import type { ColumnDef } from "@tanstack/react-table";
 import { useRef, useState } from "react";
 
 import type { PubsId } from "db/public";
-import { pubFieldsIdSchema, pubsIdSchema } from "db/public";
 import { Button } from "ui/button";
 import { Checkbox } from "ui/checkbox";
 import { DataTableColumnHeader } from "ui/data-table";
 
-import type { GetPubsResult } from "~/lib/server";
+import type { ContextEditorPub } from "../ContextEditor/ContextEditorContext";
```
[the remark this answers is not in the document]
kalilsn: got it. i just made this change because `GetPubsResult` is even more deprecated, but i agree we should use a new type!

core/app/components/pubs/PubEditor/PubEditorClient.tsx
```diff
@@ -76,25 +100,58 @@ const buildDefaultValues = (elements: BasicFormElements[], pubValues: ProcessedP
...
 				defaultValues[element.slug] = relatedPubValues.map((pv) => ({
-					value:
-						pv.schemaName === CoreSchemaType.DateTime
-							? new Date(pv.value as string)
-							: pv.value,
+					value: pv.value,
 					relatedPubId: pv.relatedPubId,
+					rank: pv.rank,
+					valueId: pv.id,
 				}));
```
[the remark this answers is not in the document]
kalilsn: good point, that doesn't look quite right!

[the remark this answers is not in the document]
kalilsn: the `forEach` is right: it's building the arrayDefaults and the regular default values in the same loop. the map was just leftover from before the refactor that i accidentally un-deleted!

core/app/components/forms/types.ts
```diff
@@ -86,3 +89,24 @@ export type StructuralElement = {
 export type FormElements = PubFieldElement | StructuralElement | ButtonElement;
 
 export type BasicFormElements = ButtonElement | StructuralElement | BasicPubFieldElement;
+
+export type FieldValue = {
+	value: JsonValue;
+	relatedPubId: PubsId;
+	rank: string;
+	valueId?: PubValuesId;
+};
```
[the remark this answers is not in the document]
kalilsn: I dropped the `Pub` to make it a bit shorter. I also added this prefix to a few more types and pluralized the `FormValue` types to `FormValues` since they are used as a type param that extends `FieldValues`:
image

core/app/components/pubs/PubEditor/actions.db.test.ts
tefkah: hmm good point! would not be that hard to add a `returning` thing at the end, if it's not there then we don't do the fetch. would you like to see anything in particular returned you think if `returning` is not specified?

tefkah: i guess we could also return the things that are changed, but that might get quite messy

core/app/components/pubs/PubEditor/actions.ts
tefkah: oh good point, we're indeed not doing that. should not be hard to add

core/lib/pubs.ts
```diff
@@ -47,48 +45,6 @@ export const getPubTitle = (pub: PubTitleProps): string => {
...
-		assigneeId: pub.assignee?.id ?? null,
-		assignee: (pub.assignee ?? null) as GetPubsResult[number]["assignee"],
-		pubType: pub.pubType as GetPubsResult[number]["pubType"],
-	};
-};
-
-export const processedPubsToPubsResult = (pubs: InputPub[]): GetPubsResult => {
-	return pubs.map(processedPubToPubResult);
-};
-
```
tefkah, line 91: good riddance

core/lib/server/pub.ts
```diff
@@ -326,7 +222,7 @@ export const maybeWithTrx = async <T>(
 
 const isRelatedPubInit = (value: unknown): value is { value: unknown; relatedPubId: PubsId }[] =>
 	Array.isArray(value) &&
-	value.every((v) => typeof v === "object" && "value" in v && "relatedPubId" in v);
+	value.every((v) => typeof v === "object" && v && "value" in v && "relatedPubId" in v);
```
tefkah, line 225: just javascript things™️

core/lib/server/pub.ts
```diff
@@ -999,7 +891,13 @@ export const updatePub = async ({
...
 		// Update the stage if a target stage was provided.
 		if (stageId !== undefined) {
-			await movePub(pubId, stageId, trx).execute();
+			try {
+				await movePub(pubId, stageId, trx).execute();
+			} catch (err) {
+				if (!isUniqueConstraintError(err)) {
+					throw err;
+				}
+			}
```
tefkah, line 900: what is this supposed to catch? you moving it to the stage it's already in?

core/lib/server/stages.ts
tefkah: lol.

tefkah (approved):
works and looks great! great job!

core/lib/server/pub.ts
kalilsn: yeah exactly. previously that was taken care of (incorrectly) by an `on conflict do nothing` clause. now we just catch those errors and continue since there's no need to halt just because the stage is unchanged

Issue #1003, opened 2025-02-27
Editing a pub internally breaks if related pub doesn't happen to be within the limit


allisonking:
### What went wrong, step-by-step?

https://knowledgefutures.slack.com/archives/CMAQM0BNV/p1740674465642779

1. Create a pub that has a related pub
2. To reproduce this, you can probably change the limit here to something very small such that the pub that was related no longer is returned in the query https://github.com/pubpub/platform/blob/626e97e310537e74eeb8a77289514b7319d96d83/core/app/components/pubs/PubEditor/PubEditor.tsx#L166
3. Loading the edit page for the pub will break because the app can't find the related pub

### What did you expect to happen?
The app should be able to find the related pub.

### What URL can we use to see the issue?
https://app.pubpub.org/c/arcadia/pubs/fb6347b2-ae79-4216-afdf-3dcc41906dca/edit

Add the "Pub Contributors" relation [here](https://app.pubpub.org/c/arcadia/forms/pub-default-editor/edit) and then the edit form will break

### Possible solution

1. if we have an existing pub, make sure to also pass that pub's related pubs to the context (maybe concatenate it [here](https://github.com/pubpub/platform/blob/main/core/app/components/pubs/PubEditor/PubEditor.tsx#L277) if possible?)
2. the side menu that lets you add related pubs should query/paginate rather than just using the first 30 and stopping there

Just doing 1‍⃣ should fix the bug—would need 2‍⃣ to actually be able to associate any pub with another (and not just the first 30). If it's just 1‍⃣ then this is maybe a 1 point ticket, but if doing 2‍⃣ as well then it might be 2 or 3 points.

gabestein:
Thanks for this! Let's do #1 and do #2 as a follow on.

PR #1017, opened 2025-03-04
Fix related pubs bug

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/1003

Our pub editor requires a list of pubs in a context so that it knows what pubs can be related to, and so it can render the titles of pubs that are already related. The problem is that when we query for pubs, we need to limit (in this case we limit to `30`). So a pub might be related to another pub that is not in that initial 30, in which case the app can't render the pub's title since it wasn't returned with the query and breaks.

This fixes this initial problem by concatenating an existing pub's related pubs to the pubs sent to the pub context.

## Test Plan
1. Add a relation field (can be null)
2. Add this field to a pubtype (i.e. Evaluation)
3. Add this field to the pubtype's default form too
4. Open an existing Evaluation pub (or create a new one)
5. Create a related pub via the "Add Related Pub" button on the Evaluation pub
6. Fill in any info to create the pub
7. Now add one more pub of any type (this is just to put the created related pub out of range)
8. Change the `limit` to `1`
https://github.com/pubpub/platform/blob/626e97e310537e74eeb8a77289514b7319d96d83/core/app/components/pubs/PubEditor/PubEditor.tsx#L166
9. Open the pub in step 4
10. Click "Update"
11. The pub should still render properly

### DIFF
--- a/.github/workflows/ecrbuild-all.yml
+++ b/.github/workflows/ecrbuild-all.yml
@@ -16,7 +16,7 @@ on:
         outputs:
             core-image:
                 description: "Core image SHA"
-                value: ${{ jobs.build-core.outputs.image-sha}}
+                value: ${{ jobs.build-core.outputs.image-sha }}
             base-image:
                 description: "Base image SHA"
                 value: ${{ jobs.build-base.outputs.image-sha }}
--- a/.github/workflows/on_pr.yml
+++ b/.github/workflows/on_pr.yml
@@ -34,3 +34,50 @@ jobs:
         secrets:
             AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
             AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+
+    deploy-preview:
+        permissions: write-all
+        runs-on: ubuntu-latest
+        timeout-minutes: 30
+        needs:
+            - build-all
+        steps:
+            - uses: actions/checkout@v4
+
+            - name: Copy .env file
+              run: cp ./self-host/.env.example ./self-host/.env
+
+            - name: Configure pullpreview
+              env:
+                  PLATFORM_IMAGE: ${{ needs.build-all.outputs.core-image }}
+                  JOBS_IMAGE: ${{ needs.build-all.outputs.jobs-image }}
+                  MIGRATIONS_IMAGE: ${{ needs.build-all.outputs.base-image }}
+              run: |
+                  sed -i "s|image: PLATFORM_IMAGE|image: $PLATFORM_IMAGE|" docker-compose.preview.yml
+                  sed -i "s|image: JOBS_IMAGE|image: $JOBS_IMAGE|" docker-compose.preview.yml
+                  sed -i "s|image: MIGRATIONS_IMAGE|image: $MIGRATIONS_IMAGE|" docker-compose.preview.yml
+                  sed -i "s|email someone@example.com|email dev@pubpub.org|" self-host/caddy/Caddyfile
+                  sed -i "s|example.com|{\$PUBLIC_URL}|" self-host/caddy/Caddyfile
+
+            - name: Get ECR token
+              id: ecrtoken
+              run: echo "value=$(aws ecr get-login-password --region us-east-1)" >> $GITHUB_OUTPUT
+              env:
+                  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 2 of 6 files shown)


core/app/c/[communitySlug]/pubs/[pubId]/edit/page.tsx
```diff
@@ -130,7 +130,7 @@ export default async function Page(props: {
 			}
 		>
 			<div className="flex justify-center py-10">
-				<div className="prose max-w-prose flex-1">
```
Comment (allisonking) [inline core/app/c/[communitySlug]/pubs/[pubId]/edit/page.tsx:133]: this originally came from https://github.com/pubpub/platform/pull/980 but it seems like just the `flex-1` is enough to keep the width. adding `prose` changes a bunch of styles, including making `<hr/>`s have huge margins

image

`prose` is used for markdown, but I believe the various markdown places already have `prose` on them, i.e. `StructuralElement`
https://github.com/pubpub/platform/blob/0711707979fe1bc51874d120e33d6626b3f7be32/core/app/components/FormBuilder/FormElement.tsx#L167

core/app/components/pubs/PubEditor/PubEditor.tsx
```diff
@@ -44,8 +45,10 @@ const RelatedPubValueElement = ({
 				<p className="text-sm">
 					You are creating a Pub related to{" "}
 					<span className="font-semibold">{getPubTitle(relatedPub)}</span> through the{" "}
-					<span className="font-semibold">{fieldName}</span> pub field. Please enter a
-					value for this relationship.
+					<span className="font-semibold">{fieldName}</span> pub field.
+					{element.schemaName !== CoreSchemaType.Null && (
+						<span>Please enter a value for this relationship.</span>
+					)}
```
allisonking, line 51: I just noticed it was still saying to fill out a field if the schema was null (and then it wouldn't render an input to fill out)

core/app/components/pubs/PubEditor/PubEditor.tsx
```diff
@@ -269,12 +272,17 @@ export async function PubEditor(props: PubEditorProps) {
 	const currentStageId = pub?.stage?.id ?? ("stageId" in props ? props.stageId : undefined);
 	const pubForForm = pub ?? { id: pubId, values: [], pubTypeId: form.pubTypeId };
 
+	// For the Context, we want both the pubs from the initial pub query (which is limited)
+	// as well as the pubs related to this pub
+	const relatedPubs = pub ? pub.values.flatMap((v) => (v.relatedPub ? [v.relatedPub] : [])) : [];
```
allisonking, line 277: TIL: you can use `flatMap` to get around how doing a `.map(...).filter(value => !!value)` doesn't let TS think you have `Thing[]` and TS will still think you have `(Thing | null)[]` even after the filter

https://stackoverflow.com/a/59726888

kalilsn: very hot tip, the .filter typing is so annoying!

kalilsn (approved):
Works great! Tested and confirmed the bug on main (with limit 1) and that it works on your branch

core/app/components/pubs/PubEditor/PubEditor.tsx
tefkah: i think they did fix this in a recent release, if you do `.filter(value=>value != null)` specifically!

tefkah: oh no nvm, this only applies one level deep! they fixed the "simple" case in typescript 5.5

https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates

Example (change to typescript <5.5 to see error in the simple case)

https://www.typescriptlang.org/play/?target=99&pretty=true&allowSyntheticDefaultImports=true&preserveValueImports=true&isolatedModules=true&verbatimModuleSyntax=false&ts=5.5.4&ssl=1&ssc=1&pln=2&pc=1#code/C4TwDgpgBAwg9gWzAGwgDwCoAsCWA7AcygF4oBvKLCEALnMrjoHIrlk4omBfKAHyjwBXNlwBQoSFADKOJKmz4ipClVqdW7JlB78hbUQBMIAY2QBDAE7RjcPAGdgUG3PR0AFPBeZchAJQBtAF1DE3MrJ1sHKDtZFAg6GRcFQiDRURt7R2cwEgivADoAMxxkYAgLNwA3YgA+SvzVKABCUj1kX1FURys7XOz-AAZAhuoGjlEAegmoGagAPQB+KDSMqJikXPW4opKyiuqapqbKjq6oHoAmTdjB4dVJ6dnFqCA

PR #1019, opened 2025-03-04
fix: allow file-uploads from the server > 5mb

tefkah:
- **fix: update aws sdk version to unbreak large file uploads**
- **dev: add tests with large file**

## Issue(s) Resolved
`uploadFileToS3` would fail if filesize > 5mb

Version mismatch with s3 client and the multipart fileupload helper library. 

Since they introduced this change (https://github.com/aws/aws-sdk-js-v3/issues/6810), it would try to access a checksum method which did not yet exist.

## Test Plan
1. Run tests.
2. See that the test fails on the previous version. ie, in `core`, do `pnpm i @aws-sdk/client-s3@3.645.0` and do `pnpm vitest lib/server/assets.db.test.ts`


allisonking (approved):
worked as you said it would!

Issue #1020, opened 2025-03-04
bug: Editor attributes panel is trapped in a box


gabestein:
### What went wrong, step-by-step?

1. Clicked to open some attributes in the editor (or, ~ to open the control menu)
2. Whoops, it's trapped in the editor box, with scrollbars

### What did you expect to happen?
Panel lays on top of box

### Have any screenshots? (Attach them by dragging the image here)

![Image](https://github.com/user-attachments/assets/e7ddfa30-549e-4050-a30b-ff60c6b7441b)

Issue #983, opened 2025-02-20
Upload image at cursor using Uppy


gabestein:
## Motivation

So users can add images to rich text fields

## Requirements

- Add a button to formatting bar that, when clicked, opens a popover containing an Uppy drop zone
- When a file upload completes, image is inserted at the current cursor position within the editor

## Acceptance Criteria

- Uploaded image uses prosemirror schema with all eventual fields and default values, but no user ability to set values:
   - alt text (blank string)
   - link to (blank string)
   - credit (blank 'rich text')
   - license (blank 'rich text')
   - width (int, default to 100)
   - alignment (left|center|right|block)
 - Uploaded image is added to assets.app.pubpub.org S3 bucket, prepended by the pub ID (e.g. `assets.app.pubpub.org/<pubid>/filename`
 - Uploaded image does not overwrite existing files with the same name in the bucket (this is an S3 setting, I think)
 - DOES NOT IMPLEMENT MEDIA SETTINGS EDITOR POPOVER

## Mockups (if available)
<img width="380" alt="Image" src="https://github.com/user-attachments/assets/cf0a49a1-c6b8-49fe-9895-c21805930d59" />

allisonking:
is `width` a percentage, or something like a pixel?

tefkah:
> is `width` a percentage, or something like a pixel?

in legacy it was a percentage, i think that makes the most sense

PR #1022, opened 2025-03-04
Uppy image uploading

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/983
Closes https://github.com/pubpub/platform/issues/1020

Adds a media uploader to the menu bar

## Test Plan
Open up the editor either in the app or in storybook and try uploading an image which should insert it at the cursor. Should also be able to insert multiple images at once, and should prevent files that are not images from being uploaded. 

https://github.com/user-attachments/assets/f1afa009-9147-4217-b3d5-d1ceba0b9d3c

## Notes

Also fixed https://github.com/pubpub/platform/issues/1020 by moving where we are setting the overflow https://github.com/pubpub/platform/pull/1022/commits/984ef80ca813cfe5aec41735d5e88b6d7c0863f6


packages/context-editor/src/components/ImageUploader.tsx
```diff
@@ -0,0 +1,77 @@
...
+			{/* <FileUpload
+				upload={(filename) => {
+					console.log({ filename });
+					// TODO: how best to get signed upload here?
+					return Promise.resolve("test");
+				}}
+				onUpdateFiles={(event) => {
+					console.log({ event });
+				}}
+			/> */}
```
allisonking, line 74: @tefkah do you have any recommendations for how to let the editor get a signed url? I think maybe the parent component could pass a function in? or should the editor generate it? ideally it would work both in storybook and in the app

packages/context-editor/src/components/MenuBar.tsx
```diff
@@ -199,10 +242,10 @@ const Separator = () => {
 	);
 };
 
-export const MenuBar = () => {
+export const MenuBar = ({ upload }: { upload: Upload }) => {
 	return (
 		<div
-			className="flex items-center rounded-t border bg-gray-50 p-4"
+			className="flex items-center overflow-x-auto rounded-t border bg-gray-50 p-4"
```
allisonking, line 248: our menu bar is getting longer! adds horizontal scrolling to it

https://github.com/user-attachments/assets/3d34cc47-839e-44c4-8106-8d225400cae7

packages/ui/src/customRenderers/fileUpload/fileUpload.tsx
```diff
@@ -53,7 +58,13 @@ const FileUpload = forwardRef(function FileUpload(props: FileUploadProps, ref) {
...
 			}) as FormattedFile[];
 			props.onUpdateFiles(formattedFiles);
-		});
+		};
+		uppy.on("complete", handler);
+
+		// Make sure we only have one listener at a time
+		return () => {
+			uppy.off("complete", handler);
+		};
```
allisonking, line 67: without this, subscription seemed to happen twice since the component rerendered in the app, so the 'complete' handler would fire twice, resulting in double the images being inserted in the editor!

packages/context-editor/src/components/ImageUploader.tsx
```diff
@@ -0,0 +1,48 @@
...
+				id: "",
+				class: null,
+				alt: file.fileName,
+				linkTo: "",
+				credit: null,
+				license: null,
+				width: 100,
+				align: "center",
+			};
+			insertNodeAfterSelection(view.state, view.dispatch, "image", attrs);
```
allisonking, line 30: weirdly this seems to have different behavior in storybook compared to in the app. in storybook, the files insert in the opposite order, and also insert a node between each. 

in platform:
image

storybook:
image

3mcd: 🤷 I don't think this should block your PR, but we'll see if it the behavior in prod is undesireable or non-deterministic

3mcd (approved):
Works great!

PR #1024, opened 2025-03-05
chore: use smaller google dependencies

tefkah:
- **chore: run pnpm dedupe (2190->2114 deps)**
- **chore: replace googleapis with @googleapis/drive for much smaller installs**
- **chore: align google-auth-library versions**

## Issue(s) Resolved
-

I was messing around with the bundle analyzer, and realized \>50% of all of our outputted server code are unused `googleapis`.

By importing only `@googleapi/drive` and `google-auth-library`, we see a >50% reduction in server bundle size, as well as likely having slightly faster installs.

Also i ran `pnpm dedupe` bc i just learned of it.

## Test Plan
1. @istravis could you check whether the google drive import still functions as it should? i think it should "just work", but if this breaks anything it's obv not worth it.

Outputs are the node output of `@next/bundle-analyze`.

Before:
image
	
After:
image


isTravis (approved):
Incredible - bundle size improvement PRs are one of my favorite. Great catch finding that the bulk googleapis took up that much!

Testing locally, it does seem to all work identically.

The only question I have is with `@googleapis/drive` being added as a devDependency. I don't know if we use NODE_ENV=production in production, but if so, that seems like it might cause the dependency to be missing once deployed?

tefkah:
> Incredible - bundle size improvement PRs are one of my favorite. Great catch finding that the bulk googleapis took up that much!
> 
> Testing locally, it does seem to all work identically.
> 
> The only question I have is with `@googleapis/drive` being added as a devDependency. I don't know if we use NODE_ENV=production in production, but if so, that seems like it might cause the dependency to be missing once deployed?

oh no clue why it was added as a devdep, lll add it as a normal one!

PR #1026, opened 2025-03-05
Google action support for narrative pub types

isTravis:
This adds a simple check for whether we're working with a narrative or not, and if so, skips version, discussion, etc creation.


gabestein (approved):
Looks good!

Noting for the future that I think the longer-term solution here is to refactor to add some of this to the settings of the action.

PR #1027, opened 2025-03-05
feat: allow actions to run after other action

### LINKED ISSUE #970: Trigger: when an action succeeds / fails
## Motivation

So users can chain together actions! e.g. when a pub successfully gets a DOI, it moves on to a "ready to publish" stage.
So users can run actions in a stage in order -- e.g. first pull from airtable, then pull from google drive (could also do this with moves, but would need to trigger moves)

## Requirements

- New labels on all action rule dialogues to clarify what each dropdown is selecting
- Two new event types in action rules: "When an action runs successfully..." and "When an action fails..."
- When one is selected, a new dropdown labeled "Action" appears that allows users choose from a list of configured actions in that stage that, when run, will cause the selected action to run
- Once configured, rule runs selected action when another action has succeeded or failed
- Warn the user if there is an infinite loop in the same stage and prevent them from creating the trigger

## Acceptance Criteria

- Two new rules and corresponding `Event`s should be added for the "action run succeeded" and "action run failed" events.
- Each new rule should have an additional configuration that specifies the action instance to observe for successes/failures.
- To prevent loops, action that triggered and triggered action cannot be the same

## Mockups (if available)
Image

tefkah:
## Issue(s) Resolved

Resolves #970 

This PR adds a rule which allows actions to run sequentially.

The logic is somewhat like this:

Imagine we have the rules

- If **Email** successful, run **Log**
- If **Log** successful, run **Move Pub**

This action sequence would be something like this

- Run `Email`
- `Email` successful
- Check whether any `onActionSuccessful` rules for `Email` are on the stage
- If so
   - create actionrun which is "scheduled" for `Log`
   - send a message to `jobs` to schedule a `ScheduledEvent` event for `Log` in 0 minutes with stack `[Email]`
   - 0 minutes pass
   - `jobs` receives event type `ScheduledEvent`
   - `jobs` sends an api request to core to trigger `Log`
   - `core` runs action `Log` with stack `[Email]`

the same happens when `Log` completes, the stack now being `[Email, Log]`

This stack is used to keep track of the order of operations, and on the length of the operations. If this stack exceeds length `MAX_STACK_DEPTH`, we throw an error. We also do not allow the creation of rules which would lead to exceeding this max_stack_depth, but it's good to be extra careful.

#### UI
The UI here is somewhat vague, but i did not want to spend more time changing it. If you think I should change it though (for instance by putting the "watched" action at the top) please let me know!

image

#### Jobs flow

the whole core => jobs => core flow is very hard to keep track of. this does is not helped by the nomenclature of "scheduableActions" etc.

To keep the pr small i did not want to change much wrt to the jobs flow, but it can be improved a lot.

## Test Plan

1. Go to https://pr-1027-rule-after-acti-ip-52-90-172-65.my.preview.run/c/croccroc/stages/manage?editingStageId=c6f7c69f-6763-49d7-aad0-3edb2ba7be80
2. Trigger action "Log 9". 
3. Go to action log
4. See that this successfully triggered Log 8 through 1. (if it failed it should say "Scheduled" for Log 8)

___

1. Run the "Email" action without configuring it
2. Go to action log
3. See that it fails, and that another log is triggered.

___

1. Try to add a rule like so
image
2. Observe that you can create a chain of 10 actions
3. Try to add rule like so
image
4. Observe that you are not allowed to do that.

___

1. Try to add a rule like so
image

5. Observe that you cant (the mesage is not very clear)
6. Try to add a rule like so
image
7. Observe that you can add such an action

___

Try to create a state you think is objectionable.

Changed how rules are displayed
image

New rule display (kind of weird)

image

### DIFF
--- a/self-host/docker-compose.yml
+++ b/self-host/docker-compose.yml
@@ -47,10 +47,7 @@ services:
         env_file: .env
         environment:
             DATABASE_URL: postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
-        # FIXME: remove this once it's merged into main
-        volumes:
-            - ../core/prisma/create-admin-user.cts:/usr/src/app/core/prisma/create-admin-user.cts
-        command: ["pnpm", "--filter", "core", "reset"]
+        command: ["pnpm", "--filter", "core", "migrate-docker"]
         networks:
             - app-network
 


core/actions/_lib/runActionInstance.ts
```diff
@@ -203,42 +218,64 @@ const _runActionInstance = async (
...
-						.where("status", "=", ActionRunStatus.scheduled)
-				// this should be guaranteed to be unique, as only one actionInstance should be scheduled per pub
-			)
 			.insertInto("action_runs")
 			.values((eb) => ({
-				id:
-					isActionUserInitiated || args.event !== Event.pubInStageForDuration
-						? undefined
-						: eb.selectFrom("existingScheduledActionRun").select("id"),
+				id: args.scheduledActionRunId,
```
tefkah, line 278: since im passing around the `scheduledActionRunId` anyway, i don't need to look it up anymore (which was flaky).

what this was doing at first was incorrect anyway. it was based on the (then already) invalid assumption that there would only be one action run scheduled per action instance per stage per pub.
 basically, when we just had the `pubEnteredStage` rule for a specific `actionInstanceId`, i would
- schedule that actionInstance when a pub entered a stage
- create a `scheduled` `ActionRun`
- when the action is run, i look up the `scheduled` `ActionRun` for that actionInstance and pub, and update it.

however this was very finicky. if for instance, you move a pub in and out of a stage twice, you'd end up with two scheduled actioninstances for the same pub id. then this query would fail, as `eb.selectFrom` would return two actions. 

this assumption would be even more incorrect now, as one actioninstance now can have many scheduled actionruns (Log1->Log2 and Log1->Log3)

even if i limited this, there was no guarantee i would pick out the correct action run.

therefore, passing around the scheduledActionRunId was the way to go

(btw the reason i am creating/updating the actionrun in the first place is bc i need it to set the `lastModifiedBy` if the action run ends up modifying a pub)

core/prisma/schema/schema.prisma
```diff
@@ -452,22 +460,26 @@ enum ActionRunStatus {
...
   event            Event
-  actionInstance   ActionInstance @relation(fields: [actionInstanceId], references: [id], onDelete: Cascade)
+  actionInstance   ActionInstance  @relation(fields: [actionInstanceId], references: [id], onDelete: Cascade)
   actionInstanceId String
   config           Json?
-  createdAt        DateTime       @default(now())
-  updatedAt        DateTime       @default(now()) @updatedAt
+  // as the trigger for "actionSucceeded" or "actionFailed"
+  watchedActionId  String?
+  watchedAction    ActionInstance? @relation("WatchedActions", fields: [watchedActionId], references: [id], onDelete: SetNull)
```
tefkah, line 470: this is sometimes referred to as the `triggeringActionInstanceId`. I should probably rename this to `watchedActionInstanceId`. please tell me i need to do that.

core/prisma/migrations/20250305135055_add_action_ref_to_rul/migration.sql
```diff
@@ -0,0 +1,13 @@
...
+-- AddForeignKey
+ALTER TABLE "rules" ADD CONSTRAINT "rules_watchedActionId_fkey" FOREIGN KEY ("watchedActionId") REFERENCES "action_instances"("id") ON DELETE SET NULL ON UPDATE CASCADE;
+
+DROP INDEX "rules_actionInstanceId_event_key";
+
+-- unique index for action chaining events
+CREATE UNIQUE INDEX "unique_action_chaining_events" ON "rules" ("actionInstanceId", "event", "watchedActionId") WHERE "watchedActionId" IS NOT NULL;
+
+-- unique index for regular events
+CREATE UNIQUE INDEX "unique_regular_events" ON "rules" ("actionInstanceId", "event") WHERE "watchedActionId" IS NULL;
```
tefkah, line 13: instead of having a unique index for rules of `actionInstanceId` + `event`, now there's one for `actionInstanceId + event + watchedActionInstanceId = null` and one for `actionInstanceId + event + watchedActionInstanceId`.

Maybe i could have made it more specific by only allowing  the latter index with `watchedActionInstanceId` for `event.actionSucceeded` and `event.actionFailed`. let me know if i should change that

core/prisma/schema/schema.prisma
```diff
@@ -442,6 +443,13 @@ model ActionRun {
   updatedAt        DateTime          @default(now()) @updatedAt
   PubValueHistory  PubValueHistory[]
 
+  // action run that triggered this action run
+  triggeringActionRunId String?
+  triggeringActionRun   ActionRun? @relation("triggering_action_run", fields: [triggeringActionRunId], references: [id], onDelete: SetNull)
```
tefkah, line 448: i want to display that an action was triggered by another, so i added this column. 

this is different from the initial idea of adding some kind of `stackId`, which would allow us to get all the runs in one stack.

the problem with this is that we would not be able to 
a. show the order of executions (except by time i guess)
b. see the separate threads if one action triggers multiple other actions, eg  `1->2, 1->3, 2->4, 3->5`. these would all have the same `stackId`

core/app/components/ActionUI/ActionRunForm.tsx
```diff
@@ -48,6 +48,7 @@ export const ActionRunForm = ({
 					pubId,
 					actionInstanceArgs: values,
 					communityId: community.id as CommunitiesId,
+					stack: [],
```
tefkah, line 51: when starting the action in a form, it has no stack

core/app/c/[communitySlug]/stages/manage/components/panel/actionsTab/StagePanelRuleCreator.tsx
```diff
@@ -196,6 +318,26 @@ export const StagePanelRuleCreator = (props: Props) => {
...
+										control={form.control}
+										name="watchedActionInstanceId"
+										render={({ field }) => (
+											<ActionSelector
+												fieldProps={field}
+												actionInstances={props.actionInstances}
+												label="after action..."
+												placeholder="Select action to watch"
+												key={field.value}
+												disabledActionId={selectedActionInstanceId} // Prevent self-references
```
Comment (tefkah) [inline core/app/c/[communitySlug]/stages/manage/components/panel/actionsTab/StagePanelRuleCreator.tsx:334]: maybe we could do more to show the user what combinations are not allowed (cycles, max depth, already existing rules) before the user submits, but for now i think this is fine. but if we do want better client feedback (other than a toast) let me know!

core/actions/_lib/zodTypes.ts
```diff
@@ -32,5 +32,27 @@ class StringWithTokens extends z.ZodString {
...
+// @ts-expect-error FIXME: '{ name: z.ZodString; description: z.ZodString; icon: z.ZodString; action: z.ZodString; actionInstanceId: z.ZodString; }' is assignable to the constraint of type 'T_1', but 'T_1' could be instantiated with a different subtype of constraint 'ZodRawShape'.ts(2417)
+class ActionInstance extends z.ZodObject<typeof actionInstanceShape, "strip", z.ZodTypeAny> {
+	static create = () =>
+		new ActionInstance({
+			typeName: "ActionInstance" as z.ZodFirstPartyTypeKind.ZodObject,
+			shape: () => actionInstanceShape,
+			catchall: z.never(),
+			unknownKeys: "strip",
+		});
+}
```
tefkah, line 54: i initially went a different route with the rule config, treating the action config the same as eg the config for `pubEnteredStage`.

I can still see us referring to other actions when configuring actions, so i decided to leave it in. please let me know if i should just remove it! this also goes for the `actionInstance` context in `ui`

core/lib/server/rules.db.test.ts
tefkah, line 1: would appreciate a check whether these tests are exhaustive enough! could add more!

core/playwright/actions.rules.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+		await page.goto(`/c/${COMMUNITY_SLUG}/activity/actions`);
+
+		await page.getByText("Log 1").waitFor();
+		await page.getByText("Rule (Log 2 Succeeded)", { exact: true });
+		await page.getByText("Log 2", { exact: true }).waitFor();
+
+		const success = await page.getByText("success").all();
+		test.expect(success).toHaveLength(2);
+	});
```
tefkah, line 97: the fact that there's only one test here is purely laziness. please tell me to write a few more if you think its a good idea

core/prisma/exampleCommunitySeeds/croccroc.ts
```diff
@@ -162,6 +177,105 @@ export async function seedCroccroc(communityId?: CommunitiesId) {
...
+							event: Event.actionSucceeded,
+							watchedAction: "Log 9",
+						},
+						{
+							actionInstance: "Log 1",
+							event: Event.actionFailed,
+							watchedAction: "Email 1",
+						},
+					],
+				},
```
tefkah, line 278: i added this so its easy to test the cycle/max depth detection

core/prisma/exampleCommunitySeeds/croccroc.ts
```diff
@@ -44,6 +46,7 @@ export async function seedCroccroc(communityId?: CommunitiesId) {
 					File: { isTitle: false },
 					Confidence: { isTitle: false },
 					"Published At": { isTitle: false },
+					"File Upload": { isTitle: false },
```
tefkah, line 49: for testing fileuploads it's annoying to have to add it to croccroc everytime

jobs/src/jobs/emitEvent.ts
```diff
@@ -163,6 +173,9 @@ const triggerAction = async (
 		logger.error({
 			msg: `Error trigger action ${actionInstanceId} for "${event}" event for Stage ${stageId} and Pub ${pubId}`,
 		});
+		if (e instanceof Error) {
+			logger.error({ message: e.message });
+		}
```
tefkah, line 178: if it failed to run, it would not show why

packages/ui/src/actionInstances/ActionInstanceSelector.tsx
tefkah, line 1: see my other comment. i tried a different approach using `autoform` first, these things are vestigial, but i thought maybe useful someday. let me know whether i should keep them and/or use this component in the `StagePanelRulesCreator` form instead

core/prisma/schema/schema.prisma
tefkah: i think i'll go for sourceActionInstance, i think that's the clearest! also when used in other contexts, like the action run. i initially called it `triggeringActionRunId`, which is very confusing

PR #1030, opened 2025-03-05
feat(rm-children|3/4): migrate hierarchy script

3mcd:
## Issue(s) Resolved

Partial #952

This PR adds a script that finds all pubs with parents and links them to their parent through an relation field.

The relation field is derived from the child pub's pub type name. For example, if an Evaluation pub is a child of a Submission pub, this script will relate the Evaluation pub to its parent Submission pub through an upserted `community:evaluations` relation pub field.

Please note that this script will not unparent child pubs, nor does this branch remove any code related to pub hierarchy (other than a small tweak to the arcadia seed script, which is duplicated in #995).

## Test Plan

1. Run the following command:
```
pnpm i && pnpm reset && cd core && pnpx tsx --env-file=.env.development ./kysely/scripts/migrate-hierarchy.ts
```
2. Start the app and visit the croccroc community.
3. Visit the pub detail page of a submission pub. It should be related to an Evaluation pub through a new relation field named "Evaluations":
Screenshot 2025-03-06 at 12 55 47 PM


kalilsn (approved):
Tested locally and it looks good! I appreciate that it prints out the list of fields added! My only feedback is that the test data doesn't have a lot of children, so I'm not 100% confident in running it in prod yet. Up to you if it's worth it here, but in general for scripts like this it would be great to have a "dry run" mode to see what changes are expected before making destructive changes. Since we're only adding here, it's probably not that high risk!

PR #1033, opened 2025-03-06
feat(rm-children|4/4): remove `parentId` column and cleanup remaining parent-child references

### LINKED ISSUE #952: Remove children

3mcd:
## Issue(s) Resolved

Resolves #952 

This PR cleans up any references to children that were necessary to keep Unjournal actions/templates working properly during the `rel="parent"` to `rel="unjournal:evaluations"` migration.

In addition the PR runs a migration that drops the `parentId` column from the pubs table.

## Test Plan

Click around the app. There should be no pubs with children and no references to parent/child pubs in the UI.


core/app/c/[communitySlug]/pubs/[pubId]/components/__tests__/author-pub-fixture.json
```diff
@@ -99,7 +97,6 @@
 				"children": [
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/pubs/[pubId]/components/__tests__/author-pub-fixture.json:97]: these fixtures don't need the `children` prop anymore!

core/lib/server/pub.ts
```diff
@@ -274,7 +272,6 @@ export const _deprecated_getPubs = async (
 					.innerJoin("PubsInStages", "pubs.id", "PubsInStages.pubId")
```
kalilsn, line 272: it's not in the PR so i can't comment on the lines directly, but there's an unused param related to children in this file a little above here:
https://github.com/pubpub/v7/blob/d53bc8e2e80127e0e3f6a3f8035c00a082efdca4/core/lib/server/pub.ts#L232-L238
https://github.com/pubpub/v7/blob/d53bc8e2e80127e0e3f6a3f8035c00a082efdca4/core/lib/server/pub.ts#L246

kalilsn (approved):
i found two totally inconsequential references to parents/children that you can remove, but no problems! and I'm not totally convinced we need to delete the migrate hierarchy script, since i think we'll have reasons to write js-based migrations again in the future. but it doesn't matter much! goodbye children 👋

core/lib/server/pub.ts
3mcd: Another nice catch, thank you!

core/app/c/[communitySlug]/pubs/[pubId]/components/__tests__/author-pub-fixture.json
Comment (3mcd) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/__tests__/author-pub-fixture.json:97]: Nice catch!!!

PR #1047, opened 2025-03-10
Add link support to editor

### LINKED ISSUE #944: Link button/support
## Motivation

So users can add links to other sites

## Requirements

- Configure the link destination
- Configure if link should be opened in a new tab/page
- Remove link once set without needing to highlight entire link
- Visit link

## Acceptance Criteria
- Link configuration should pop up when clicking/cursoring anywhere in link range, not just highlighting
- Link configuration changes made when cursor is anywhere in link range should apply to entire link range
- Cmd-k keyboard shortcut on a selection also creates link
- Markdown shortcut ([text](link)) renders link
- Pasting in a link automatically creates link

## Mockups (if available)

Image

Image

Image

kalilsn:
## Issue(s) Resolved
#944 
This PR adds links to the context editor:
- Adds a link button to the menu
- Adds a cmd-k shortcut to toggle a link
- Adds input rules to automatically create links from urls and from markdown link syntax
- Adds a new "pasteRules" plugin to apply those same input rules to pasted content 

It also fixes a bug in the attribute panel where editing multiple mark attributes would only save the last attribute updated (e.g. setting id and then class would only set the class).

It does not add the custom inline link menu from the issue, because I ran into severe performance issues with my approach.

## Test Plan
- Visit https://pr-1047-links-ip-52-204-131-59.my.preview.run/c/croccroc/public/forms/submission-default-editor/fill
- Try adding a link by:
    - Typing a url
    - Typing a markdown url
    - Pasting a url
    - Pasting a link
    - Highlighting some text and clicking the link button in the url or pressing cmd+k
- Try removing a link by placing the cursor inside of the link and clicking the link button or pressing cmd+k


packages/context-editor/src/commands/marks.ts
```diff
@@ -19,14 +20,37 @@ const toggleMark = (options: ToggleOptions<MarkType>) => {
 	return pmToggleMark(type)(state, dispatch);
 };
 
-export const createMarkToggle = (typeName: string) => {
+export const toggleMarkExpandEmpty = (options: ToggleOptions<MarkType>) => {
```
kalilsn, line 23: This lets the link be toggled off without having to select the entire link. Just put the cursor anywhere inside, then click the link button in the menu or press cmd+k

packages/context-editor/src/utils/links.ts
```diff
@@ -0,0 +1,43 @@
+import type { MarkType } from "prosemirror-model";
+import type { EditorState, Transaction } from "prosemirror-state";
+
+export const emailOrUriRegexBase =
+	"(?<mailto>mailto:)?(?<emailOrUri>(?:(?:(https|http|ftp)+)://)?(?:\\S+(?::\\S*)?(?<atSign>@))?(?:(?:([a-z0-9][a-z0-9-]*)?[a-z0-9]+)(?:\\.(?:[a-z0-9-])*[a-z0-9]+)*(?:\\.(?:[a-z]{2,})(:\\d{1,5})?))(?:/[^\\s]*)?)";
```
kalilsn, line 5: This is just what we were using in legacy. It matches a little too frequently, since it doesn't know which TLDs are real. We could improve that (at the cost of performance?) by using a list like https://www.npmjs.com/package/tlds. We could compromise by just taking a subset of the most common TLDs too. Thoughts?

packages/context-editor/src/commands/marks.ts
allisonking: that's really nice!

packages/context-editor/src/components/AttributePanel.tsx
```diff
@@ -63,17 +65,18 @@ export function AttributePanel({ panelPosition, viewRef }: AttributePanelProps)
...
+			...(node?.marks ?? []),
+		];
+		newMarks[index].attrs[attrKey] = value;
+		setPosition({
+			...position,
+			node: {
+				...node,
+				marks: newMarks as Mark[],
+			},
+		});
```
allisonking, line 79: oh thanks for fixing this! I did notice `class` would save sometimes and `id` wouldn't...

packages/context-editor/src/utils/links.ts
allisonking: I think we can stick with this for now, since it's easy enough to undo a link if it matches incorrectly

allisonking (approved):
nice job, it works well! it'd be great to have some tests for it too, but I know you're probably tired of looking at this 😅 maybe could be a separate ticket?

PR #1050, opened 2025-03-10
Allow adding elements to an empty form

kalilsn:
## Issue(s) Resolved
If you deleted all the elements from a form, it was impossible to add new elements to the form. The bug was easy to miss/rarely happens because we always create forms with some default fields.
https://kfg.sentry.io/issues/6368052103/?referrer=github-pr-bot

## Test Plan
- Open (or create) a form in the preview app 
- Delete every element
- Save the form
- Add a new element back to the form (it shouldn't crash)


allisonking (approved):
🚀

PR #1052, opened 2025-03-10
Add playwright to context-editor

allisonking:
## Issue(s) Resolved
Adds playwright to the context-editor so we can write tests there!

Sets up playwright in the directory, adds a few basic tests. Configures for CI

## Test Plan
Locally, try running `pnpm run playwright:test` in `packages/context-editor` and you should see the test suite run against storybook.

CI should also pass!


allisonking:
When the editor tests fail: https://github.com/pubpub/platform/actions/runs/13791638468/job/38573127504

.github/workflows/e2e.yml
```diff
@@ -95,6 +95,18 @@ jobs:
...
+              run: pnpm --filter context-editor run playwright:test
+
+            - name: Upload playwright snapshots artifact
+              if: failure() && steps.editorTests.outcome == 'failure'
+              uses: actions/upload-artifact@v4
+              with:
+                  name: playwright-snapshots-editor
+                  path: packages/context-editor/test-results
+                  retention-days: 10
+
```
allisonking, line 109: I put this here since the test had already set up all the dependencies (including playwright) but let me know if we should split it out to a different workflow instead?

tefkah: i think this is fine! the context editor tests are likely to be much quicker than the regular e2e tests as they won't have to wait for the server to respond, so running it first is a good call. 
i think we could even move it up a bit more, before we setup the db and such.

ideally, we'd run the context-editor tests in parallel with the other tests, but afaik its not trivial to run steps in parallel

tefkah (commented):
actually, i think i prefer us reducing max CI time over total CI time, so i would ideally like to have this run in a different workflow. i think you could either just copy over the minimum setup (`pnpm install`, `preconstruct build`) etc, we don't need the images iirc?

or, you could use a matrix approach like i did here, which is slightly more inefficient, but easier to reason about

https://github.com/pubpub/platform/blob/0445dc6f27afd87aa1936362289fba5c0c518582/.github/workflows/ci.yml#L16-L82

what do you think?

.github/workflows/e2e.yml
```diff
@@ -126,6 +126,9 @@ jobs:
             - name: Print container logs
               if: ${{failure() || cancelled()}}
               run: docker compose -f docker-compose.test.yml --profile integration logs
+              env:
+                  INTEGRATION_TESTS_IMAGE: ${{steps.label.outputs.core_label}}
+                  JOBS_IMAGE: ${{steps.label.outputs.jobs_label}}
```
allisonking, line 131: it seemed like this step was often erroring in CI (though it didn't affect much since this only runs if there was already a failure). I think it just needed these env vars

allisonking:
thanks for the suggestion @tefkah ! should be running as part of the matrix in `ci.yml` now!

tefkah (approved):
Great change! This'll make it much nicer to write tests for the editor.
I added a small change to also use the catalog version in core's `package.json`, just so they're always aligned.

It's good to go imo, but i did have one final caveat that would maybe be nice to add: maybe in `ci.yml` we could also spin up the integration deps for the playwright test, that way we are set up to do image upload tests for the editor as well!

allisonking:
Bumped playwright, as firefox file uploading tests were failing: https://github.com/microsoft/playwright/issues/33525

allisonking:
> It's good to go imo, but i did have one final caveat that would maybe be nice to add: maybe in ci.yml we could also spin up the integration deps for the playwright test, that way we are set up to do image upload tests for the editor as well!

👍 Added a file upload test which relies on the test setup's minio

PR #1058, opened 2025-03-11
dev: use `seedCommunity` in e2e tests

tefkah:
- **dev: add initial database seeding to playwright tests**
- **feat: make playwright load yes good**
- **fix: set dburl**
- **dev: replace a bunch of setup with db setup**
- **dev: replace a bunch of setup with db setup, + fix some form seed bugs**
- **dev: replace email test setup**
- **dev: replace externalform stuff setup**
- **fix: ensure tests can be run parallely**
- **fix: make eslint happy**
- **fix: make happy**
- **fix: restore correct label**

## Issue(s) Resolved

Slow iteration on writing tests locally, slowish execution in CI.

This PR allow playwright tests to call `seedCommunity`, which can quickly setup the test environment. This is much faster than doing it "manually", by clicking through things in the CI. It is also much more ergonomic to setup the proper environment for a test now, which might encourage us to write more e2e tests.

Additional benefits:
- Unique user creation for each tests, allows for better parallelization
- Made test communities use a random uuid rather than the current date, which allows for parallelization (otherwise, when creating communities at the same time, they would not have a unique slug)

Downsides:
- Because `seedCommunity` uses some next/react things, we need to stub those out. This was quite annoying to do for the VS Code extension, but i did manage to get it to work
-  Easy to add things in setup that we have not tested in the actual e2e tests, need to be mindful of that.
- kind of weird to have playwright do db stuff

## Test Plan
See the SPEED of these tests
image

Run the playwright tests locally! Ideally both in CLI and using the extension.


.vscode/settings.json
```diff
@@ -3,5 +3,11 @@
 		"*.css": "tailwindcss"
 	},
 	"typescript.tsdk": "./core/node_modules/typescript/lib",
-	"javascript.preferences.autoImportSpecifierExcludeRegexes": ["@prisma\\/client"]
+	"javascript.preferences.autoImportSpecifierExcludeRegexes": ["@prisma\\/client"],
+	// these variables are necessary in order to correctly run seeds
+	"playwright.env": {
+		"NODE_OPTIONS": "--import=\"#register-loader\"",
```
tefkah, line 9: this weird `#` syntax is needed because the extension run playwright in two separate directiories, `/` and `/core`. So i couldn't just have it point to `./prisma/seed/stub/register-loader.js`, bc that would not work for the one running in `/`, ughhh.

so instead, i set an `imports.#register-loader` in both `/package.json` and `core/package.json`, pointing to the same file slightly differently. this seems to work.

.vscode/settings.json
```diff
@@ -3,5 +3,11 @@
...
+		"SKIP_VALIDATION": "true",
+		"DATABASE_URL": "postgresql://postgres:postgres@localhost:54322/postgres"
```
tefkah, line 11: the `54322` is the one you get when you do `pnpm dev:setup`, which what i'm assuming you're running if you're running the e2e tests.
(`54323` is the one you get when you run `pnpm test:setup`).

unfortunately these two are atm mutually exclusive, which is very annoying but a separate problem to fix

core/package.json
```diff
@@ -4,10 +4,10 @@
...
 	"private": true,
 	"scripts": {
-		"playwright:test": "playwright test",
-		"playwright:ui": "playwright test --ui",
-		"playwright:test:dev": "TEST_DEV=true playwright test",
-		"playwright:ui:dev": "TEST_DEV=true playwright test --ui",
+		"playwright:test": "SKIP_VALIDATION=true NODE_OPTIONS='--import #register-loader' playwright test",
+		"playwright:ui": "SKIP_VALIDATION=true NODE_OPTIONS='--import #register-loader' playwright test --ui",
+		"playwright:test:dev": "SKIP_VALIDATION=true NODE_OPTIONS='--import #register-loader' TEST_DEV=true dotenv -e .env.dev -e .env.local pnpm playwright test",
+		"playwright:ui:dev": "SKIP_VALIDATION=true NODE_OPTIONS='--import #register-loader' TEST_DEV=true dotenv -e .env.dev -e .env.local pnpm playwright test --ui",
```
tefkah, line 10: wow, so beautiful

core/package.json
```diff
@@ -194,5 +194,8 @@
 		"vitest": "catalog:",
 		"yargs": "^17.7.2"
 	},
-	"prettier": "@pubpub/prettier-config"
+	"prettier": "@pubpub/prettier-config",
+	"imports": {
+		"#register-loader": "./prisma/seed/stubs/register-loader.js"
+	}
```
tefkah, line 200: see comment in settings.json for why

core/playwright.config.ts
```diff
@@ -13,10 +13,11 @@ export default defineConfig({
 	/* Retry on CI only */
 	retries: process.env.CI ? 2 : 0,
 	// multiple workers in CI is too flaky for now
-	workers: process.env.CI ? 1 : undefined,
+	workers: process.env.CI ? 2 : undefined,
```
tefkah, line 16: we can now run 2 workers!

core/playwright.config.ts
```diff
@@ -39,7 +40,7 @@ export default defineConfig({
 	// max 30 seconds per test in CI
 	timeout: process.env.CI ? 30 * 1000 : 10 * 60 * 1000,
 	/* Reporter to use. See https://playwright.dev/docs/test-reporters */
-	reporter: process.env.CI ? "github" : "html",
+	reporter: process.env.CI ? "github" : "list",
```
tefkah, line 43: don't know about you but i hate it opening my browser every time. we could change this back

core/prisma/seed/register-stub.cjs
tefkah, line 1: dont need to run these in `cjs` anymore, i figured out how to do stubs in ESM

PR #1063, opened 2025-03-11
Related pubs bug on form page

allisonking:
## Issue(s) Resolved
Follow up to https://github.com/pubpub/platform/pull/1017, I realized I made a bad assumption there, and forgot that external forms can also edit existing pubs. So the external form fill page needs pretty much the same code there.

Same solution as https://github.com/pubpub/platform/pull/1017

## Test Plan

There's a built in test case for this one!

1. In the arcadia seed, get the Author James McJimothy's pubid (should be the first pub that shows up in http://localhost:3000/c/arcadia-research/pubs)
2. Select the form "Issue Editor" http://localhost:3000/c/arcadia-research/public/forms/issue-default-editor/fill
3. Add James McJimothy's pubId to the end of this url `?pubId={pubId}`
4. The app will error on main, but be ok on this branch


kalilsn (approved):
good catch!

PR #1065, opened 2025-03-12
dev: speed up ci

tefkah:
- **fix: cache preconstruct build with turbo on ci**
- **dev: slightly more intelligently cache e2e and ci**
- **test: try to run steps concurrently**
- **chore: only run ci**

## Issue(s) Resolved

Speed up CI/CD slightly

This PR does three things to make CI slightly faster. My goal here is to decrease max CI time, sometimes at the cost of total ci time.

1. parallelize `pnpm test`, `lint`, `format`, and `type-check`. `ci` was sometimes taking longer than `build core`, and since `e2e` is dependent on both `ci` and `core`, this ended up taking longer. parallelizing them is slightly inefficient, as, without a cache, we have to download and save all the dependencies 4 times. again, tradeoff between max and total ci time.
2. Prod the `minio` for health checks more quickly during startup. Since the healthcheck interval was set to 1m30s, it would sometimes take very long for the container to be considered healthy. this can save up to a minute.
3. Cache `p:build` with turbo, and use that turbo cache in `e2e`, st `e2e` does not have to redo the preconstruct build. this can save up to 25 seconds.
4. Do not save the cache state again in `e2e`, just restore it (using `actions/cache/restore`, rather than `actions/cache`. since we don't expect the result of `preconstruct build` or `pnpm install` to differ between `ci` and `e2e`. this saves ~10/20s.

### Required actions

We need to change the status checks for branch protection from `ci` and `e2e` to `status-check`. This will allow us to merge even if we end up skipping e2e or ci.

### Feedback

Do you think this tradeoff between max and total ci time is worth it?

### Other avenues

The big speed gain we could get is if we could find a way to cache the build of `core` efficiently. I tried that a bit once again, but again saw that it's slower to cache the docker layers than it is to do so. Maybe this would change if we self-hosted the github action runners.

Wild idea: what if we _dont_ build the app in docker? what if we build it on the github actions runner, and then just copy the result over? then we could cache the `pnpm install`, the `preconstruct build`, and partially the `next build` step.

This does go against the spirit of reproducible builds though.

## Test Plan

See that the CI tests finish quicker (1 min max with cache, 2 min without), and the e2e tests

### DIFF
--- a/.github/workflows/on_pr.yml
+++ b/.github/workflows/on_pr.yml
@@ -32,12 +32,12 @@ jobs:
         if: github.event.action == 'opened' || github.event.action == 'reopened' || github.event.action == 'synchronize' || (github.event.action == 'labeled' && github.event.label.name == 'preview')
         uses: ./.github/workflows/ci.yml
 
-    # build-all:
-    #     if: github.event.action == 'opened' || github.event.action == 'reopened' || github.event.action == 'synchronize' || (github.event.action == 'labeled' && github.event.label.name == 'preview')
-    #     uses: ./.github/workflows/ecrbuild-all.yml
-    #     secrets:
-    #         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-    #         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+    build-all:
+        if: github.event.action == 'opened' || github.event.action == 'reopened' || github.event.action == 'synchronize' || (github.event.action == 'labeled' && github.event.label.name == 'preview')
+        uses: ./.github/workflows/ecrbuild-all.yml
+        secrets:
+            AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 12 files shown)


turbo.json
```diff
@@ -45,6 +45,10 @@
 		"test-run": {
 			"dependsOn": ["^topo"],
 			"outputs": [".cache/.vitestcache"]
+		},
+		"//#turbo:p:build": {
```
allisonking, line 49: should this have the `//#`?

.github/workflows/on_pr.yml
```diff
@@ -119,3 +119,13 @@ jobs:
               with:
                   source-dir: docs/out
                   action: remove
+
+    status-check:
```
allisonking, line 123: hm it seems like even with this we are hanging on a status check on this PR? 
image

although this PR doesn't have the preview tag, so not sure what's going on

tefkah: yeah this is due to our branch protection being configured to look for "CI", but now CI is split up into 4 (soon to be 5) checks which are different, so it's not recognizing it. If this is merged I want to change the check to look for "status-check" instead, which should solve this issue!

allisonking (approved):
🎉 thanks for explaining all the things!

Issue #1035, opened 2025-03-06
Use Forms to construct Pub view


gabestein:
## Motivation

So admins can specify specific views for Pubs
So users (authors, reviewers, copyeditors, etc.) can view Pubs in specific, focused contexts based on their role
So developers can determine what related fields to show users without running into graph problems 

## Requirements

- Uses the order, and label, of fields on a form to order the fields on the view page and display their labels
- If the user is a contributor (at any level), only shows the fields specified on the form
- Removes the "related pubs/section" table for contributor users
- For other users, shows the fields not on the form at the bottom of the page, in a section called "Other Fields"
- Defaults to the default form for any given type
- If no form exists for a type, create default form for pub type and use that

## Acceptance Criteria

- Fields are displayed using current form field renderers -- in the future we will return to creating read-only views

## Mockups (if available)
![Image](https://github.com/user-attachments/assets/f7ec68ed-4540-4d8a-a72f-8ce91c580a10)

tefkah:
Making read-only versions of the current form inputs might prove to be a large part of this. eg the related pub and context-editor inputs.

for the related pub inputs as views, should we allow users to view more information about those pubs, or is it just "clicking through"?

gabestein:
Agreed. We're gonna strip this down to use current renderers for now, and then return to more friendly views later.

allisonking:
a few questions:
1. What should this look like if the form has a field but there is no value? Should we only render the label, i.e.

![Image](https://github.com/user-attachments/assets/001a5bda-5e2e-4993-9450-935a6c78e98d)

(here there is no MemberId value)

2. We've been omitting the title when rendering pub values since the title is hoisted to the top header. do we still want to do this?

![Image](https://github.com/user-attachments/assets/caecbbf6-b3c4-4fea-827b-838ddb5c8f36)

In this picture, we're rendering both the title (James McJimothy) hoisted up, as well as the title because the form has the title.

allisonking:
> Fields are displayed using current form field renderers -- in the future we will return to creating read-only views

Does this refer to the current view renderer or the form renderer? i.e.

![Image](https://github.com/user-attachments/assets/54d6780c-90b9-46fd-ad61-4e8fab288e5c)

or 

![Image](https://github.com/user-attachments/assets/56d902cb-1350-4a83-93c3-7c9d7debca82)

gabestein:
> What should this look like if the form has a field but there is no value? Should we only render the label, i.e.

[...]

PR #1070, opened 2025-03-12
Use forms to render pub view

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/1035

Gets the default form and combines it with the pub values to render the pub values according to the form.

## Test Plan

1. Visit the arcadia research community and find the 'James McJimothy' pub
2. In a separate tab, find the "Author Editor (default)" form and open the form builder for it
3. Change the order of the fields in the form. Note that the field designated as the 'title' will always be hoisted to the top regardless of form order.
4. Refresh the 'James McJimothy' pub. The order should change
5. Try changing labels in the form builder and refresh the pub, which should also change
6. Note that "MemberId" is blank but still shows up on the page because it is in the form but there is no pub value for it
7. Find another form, i.e. the "Tag Editor (default)" form and copy the link to its public form
8. Put James McJimothy's pub id and edit it with the tag editor in order to give it a 'title' field. This will add a pub value to the McJimothy pub that is not part of the form
9. Go back to the pub and refresh—the 'title' should show up under 'Other fields'
10. The related pub should still render but without a form (see https://github.com/pubpub/platform/issues/1078)

image


allisonking:
I'm not sure what's the right way to check the user's permissions here. If they are a contributor, then we only want to show form values, not any of the extra values a pub might have on it. 

In the updated query, we'll be able to tell if it's an 'extra' value by if the pub value has a null `formElementId`. So that's part of it, but the other part is what to check about a contributor that tells us they have this limited permission? should we be using `form_membership`s for this? would that mean we should automatically give all contributors a membership to all default forms?

core/lib/server/pub.db.test.ts
```diff
@@ -930,6 +930,150 @@ describe("getPubsWithRelatedValues", () => {
...
+		});
+		const { community, pubs, users, pubTypes, forms } = await seedCommunity(seed);
+		const { getPubsWithRelatedValues } = await import("./pub");
+
+		// Test the builtin default form first which only has 'Title' and 'Some Relation'
+		const pub1 = await getPubsWithRelatedValues(
+			{ pubId: pubs[0].id, communityId: community.id },
+			{ withValues: true }
+		);
+		// TODO: Ideally this also returns an item for the form but with empty values...
```
allisonking, line 1020: I'm not actually sure if we want this—I imagine it'd mess up other parts of the app, if we were to return null pub values for empty form elements... which makes me think that maybe we still want to return the form separately even with the joining of form elements?

core/lib/server/pub.ts
```diff
@@ -1771,10 +1782,40 @@ export async function getPubsWithRelatedValues<Options extends GetPubsWithRelate
 								"pub_fields.schemaName",
 								"pub_fields.slug as fieldSlug",
 								"pub_fields.name as fieldName",
+								"form_elements.id as formElementId",
+								"form_elements.label as formElementLabel",
+								"form_elements.config as formElementConfig",
 							])
 							.whereRef("pv.pubId", "=", "pt.pubId")
-							// Order by most recently updated value (grouped by pub field), then rank
+							// .where((eb) =>
```
allisonking, line 1790: haven't figured out how to do this part 🥲

allisonking:
This kinda sorta works, but I think I'll need help with a few things:

1. Thoughts on joining form elements that don't have values? https://github.com/pubpub/platform/pull/1070#discussion_r1994230825 I'm worried it'd affect other parts of the app, but we do want to render the form elements, just with blank values if they aren't there
2. Made some progress on contributor permission limiting, but I can't figure out how to get the query to work... https://github.com/pubpub/platform/pull/1070#discussion_r1994231227

Other than that, things kinda work!

core/app/c/[communitySlug]/stages/components/Move.tsx
```diff
@@ -34,8 +34,9 @@ const makeSourcesAndDestinations = (props: Props) => {
 	}
 
 	const stagesById = makeStagesById(props.communityStages);
-	const sources = stagesById[props.stageId].moveConstraintSources.map((mc) => stagesById[mc.id]);
-	const destinations = stagesById[props.stageId].moveConstraints.map((mc) => stagesById[mc.id]);
+	const stage = stagesById[props.stageId];
+	const sources = stage ? stage.moveConstraintSources.map((mc) => stagesById[mc.id]) : [];
+	const destinations = stage ? stage.moveConstraints.map((mc) => stagesById[mc.id]) : [];
```
Comment (allisonking) [inline core/app/c/[communitySlug]/stages/components/Move.tsx:39]: This would crash if the user does not have access to any stages (i.e. a contributor who has been invited to a pub

packages/contracts/src/resources/site.ts
```diff
@@ -176,19 +177,48 @@ export type MaybePubOptions = {
...
+
+type ValuesWithFormElements =
+	| // With both values and form elements
+	(ValueBase & {
+			formElementId: FormElementsId;
+			formElementLabel: string | null;
+			formElementConfig:
+				| { label?: string }
+				| { relationshipConfig: { label?: string } }
+				| null;
```
allisonking, line 205: maybe there's a better way to get this config type?

core/lib/pubs.db.test.ts
```diff
@@ -0,0 +1,266 @@
...
+			{ withExtraPubValues: adminCan, expected: 3 },
+			{ withExtraPubValues: editorCan, expected: 3 },
+			{ withExtraPubValues: contributorCan, expected: 2 },
+		].forEach(({ withExtraPubValues, expected }) => {
+			const pubWithForm = getPubByForm({ pub, form, withExtraPubValues });
+			expect(pubWithForm.values).toHaveLength(expected);
+		});
+	});
+
+	it("should respect form order", async () => {
```
allisonking, line 232: I wanted to add another test that makes sure the pub values not in the form still stay in order, but I'm not sure how to do this in a transaction. I think @kalilsn may have run into this problem too, was there a good workaround?

allisonking: oh that's a good point! I had been trying to control the insertion time so that I could say the order was a static value, but you're right that I should just be able to compare with whatever `getPubsWithRelatedValues` returns to get the same result!

core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx
```diff
@@ -5,24 +5,60 @@ import type { ReactNode } from "react";
...
+};
+
 const PubValueHeading = ({
 	depth,
 	children,
 	...props
 }: React.HTMLAttributes<HTMLHeadingElement> & { depth: number }) => {
+	// For "Other Fields" section header which might be one lower than any pub depth
+	if (depth < 1) {
+		return <h1 {...props}>{children}</h1>;
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:60]: I'm not sure I completely understand the situation where this case happens, but I think it should probably be an `<h2>` since there's already an `<h1>` on the page for the pub title. and if it needs to be lower than a depth 0 pub value heading, the others can be shifted up a level (case 0 -> h3 etc.)

packages/contracts/src/resources/site.ts
kalilsn: feels like there should be but idk. thanks for writing this out!

core/lib/pubs.ts
```diff
@@ -89,14 +92,124 @@ export const processedPubsToPubsResult = (pubs: InputPub[]): GetPubsResult => {
...
+	}>["values"] = [];
+	for (const slug of slugsNotInForm) {
+		for (const value of valuesByFieldSlug[slug]) {
+			valuesNotInForm.push(value);
+		}
+	}
+
+	const newValues = [...valuesWithFormElements, ...valuesNotInForm];
+
+	return { ...pub, values: newValues };
```
kalilsn, line 214: nice work! reading this function makes me appreciate how terse sql is, but i think we'll be glad you did it this way!

kalilsn (approved):
i tested this both as an admin and a contributor and it worked great! i think the headings could probably use an adjustment, but i'm not 100% sure. and i think the label issues should probably be their own issue so i'm clicking approve!

core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx
Comment (kalilsn) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:60]: actually it also looks like the heading order has been reversed now. nested pubs are using `h3` for the labels and the root pub is using `h4` on https://pr-1070-pub-view-form-ip-34-201-149-207.my.preview.run/c/arcadia-research/pubs/1a7f3f27-a8d6-40cb-8c87-b2afe2bb7188

core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx
```diff
@@ -5,24 +5,60 @@ import type { ReactNode } from "react";
...
+		const config = value.formElementConfig;
+		if (config) {
+			configLabel = "label" in config ? config.label : undefined;
+			relationshipConfigLabel =
+				"relationshipConfig" in config ? config.relationshipConfig.label : undefined;
+		}
+		formElementLabel = value.formElementLabel;
+	}
+	return formElementLabel || configLabel || relationshipConfigLabel || defaultLabel;
+};
```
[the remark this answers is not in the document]
Comment (allisonking) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:51]: hmm no I think you're right—I'm just getting all jumbled by these labels! lemme fix that, and I'll see about the form builder too

core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx
Comment (allisonking) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:60]: hmmm I think maybe we shouldn't use `depth` here, or maybe I don't understand `depth`. you're right that the `h3`s and `h4`s have become reversed for Jimothy, but they are still right for a journal article i.e. https://pr-1070-pub-view-form-ip-34-201-149-207.my.preview.run/c/arcadia-research/pubs/00ba5f00-eada-4a3f-9da6-daade13dadbc

https://github.com/user-attachments/assets/682bde6f-52d7-4924-b399-93a00d7a638c

maybe we should manually track `depth` instead of using what the pub returns, what do you think?

Comment (kalilsn) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:56]: strange! i don't see why they would be different.

> maybe we should manually track depth instead of using what the pub returns, what do you think?

manually meaning increment it in the components? that makes sense to me! though it's a bit concerning if the depth from the query is wrong since we're trying to use that to limit which pubs we display

Comment (allisonking) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:56]: hm yeah I'm not sure why `depth` seems to return `3` for both the root pub and the related pub for jimothy 😕 maybe @tefkah knows?

Comment (allisonking) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:56]: oh okay I think I know why—

Jimothy has a cycle going on, being both the root pub and a related pub at depth 3. We make a map of the pubs here:
https://github.com/pubpub/platform/blob/211c7e8ce2a3fe176756ca2fa8425b8a71f4e325/core/lib/server/pub.ts#L1772

And the depth 3 jimothy overwrites the depth 1 jimothy, and we end up with depth 3 jimothy returned as the root pub. I'm not quite sure what the right thing to do here is, maybe make the map index to be by `pubId` and by depth? not sure if that will get us into some weird cycles though

Comment (tefkah) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:56]: huh. yeah that should not happen, although im not sure if it's really a problem, as they should be the same pub. I think there should be some kind of check that prevents this from happening (the basic: is this pub already in the map). i think it's okay to leave, if you can fix it quickly i think it's worth it, otherwise we should file an issue!

Comment (tefkah) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:56]: as for the actual issue it causes (the headings): i don't think we should decrement the depth of the heading automatically per level. i havent yet looked at the code yet, but i think choosing one `h` level for the related pubs and eg h1 for the top level one is good enough. is a bit strange to keep decrementing it (what if you go 7 pubs deep, should you have an h8?). 

i think we should solve the accessibility here by making the other html of these nested pubs more semantic, eg wrapping them in `article` or something. but i think that is putting too much energy into something we're going to change pretty soon anyway (hopefully!)

core/prisma/migrations/20250313200801_form_is_default_unique_constraint/migration.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  Warnings:
+
+  - A unique constraint covering the columns `[isDefault,pubTypeId]` on the table `forms` will be added. If there are existing duplicate values, this will fail.
+
+*/
+-- CreateIndex
+CREATE UNIQUE INDEX "forms_isDefault_pubTypeId_key" ON "forms"("isDefault", "pubTypeId") WHERE "forms"."isDefault" is true;
```
tefkah, line 8: wild that we didnt have this yet, good catch!

core/prisma/migrations/20250313193528_add_see_extra_pub_values_capability/migration.sql
```diff
@@ -0,0 +1,37 @@
...
+    (
+        'editor'::"MemberRole",
+        'community'::"MembershipType",
+        'seeExtraPubValues'::"Capabilities"
+    ),
+    (
+        'admin'::"MemberRole", 
+        'community'::"MembershipType", 
+        'seeExtraPubValues'::"Capabilities"
+    );
```
tefkah, line 37: related note: i think it would be nice if we could have some way we could declare these capabilities in code and have them auto added to a migration if we change them. i already found it hard to keep track of them when they were in one migration file, now that they are starting to split into many it's going to be even harder! ill spend some time thinking about this (would also really like a solution for migrating stored procedures and triggers)

core/lib/pubs.db.test.ts
```diff
@@ -0,0 +1,524 @@
...
+});
+
+describe("getPubByForm", () => {
+	it("should be able to get a pub with form elements", async () => {
+		const { seedCommunity } = await import("~/prisma/seed/seedCommunity");
+		const { pubs, community, pubTypes } = await seedCommunity(seed);
+		const { getPubByForm } = await import("./pubs");
+		const { getPubsWithRelatedValues } = await import("./server/pub");
+		const { getForm } = await import("./server/form");
+
```
tefkah, line 141: sweet, thanks for all the tests!

for the future: you could make all these imports a bit nicer while retaining the auto rollback by creating a small helper function that does the imports, eg

```ts
const getAllImports = async () => {
    	const { seedCommunity } = await import("~/prisma/seed/seedCommunity");
		const { pubs, community, pubTypes } = await seedCommunity(seed);
		const { getPubByForm } = await import("./pubs");
		const { getPubsWithRelatedValues } = await import("./server/pub");
		const { getForm } = await import("./server/form");

        return { seedCommunity, /*...*/ }
}

core/lib/pubs.db.test.ts
tefkah, line 1: thanks for including all these tests!

even though that file is already quite big, could you put these tests in `lib/server/pub.db.test.ts` instead? at first i thought that it was here bc it's testing functionality from `/lib/pubs.ts`, but that doesn't seem to be the case, so i think only having one `pub(s).db.test.ts` is maybe easier to keep track of!

also maybe we shouldn't have a `lib/server/pub.ts` and `lib/pubs.ts` haha, so confusing

core/app/c/[communitySlug]/stages/components/Move.tsx
Comment (tefkah) [reply core/app/c/[communitySlug]/stages/components/Move.tsx:39]: nice, good catch

core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx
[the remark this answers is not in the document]
Comment (tefkah) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:47]: yeah this should probably be the `config.[whatever we called the relatonship config].label` right? 
do we still want to show the `inner` label here?

tefkah (approved):
great work! other than some very minor organizational comments it looks great!

core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx
Comment (kalilsn) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:56]: ok it seems like this bug was already present so i wouldn't worry about it on this PR. but i do think we should keep decrementing heading levels.

i would love to just solve this with some kind of appropriate semantic element to represent a nested structure, but AFAIK for screen readers, there's still nothing that replaces the the h1-6 hierarchy in the document outline. and [the overwhelming majority of screen reader users read the headings first](https://webaim.org/projects/screenreadersurvey10/#finding)! if there's an alternative we should be doing here, i'd love to know about it though. my understanding mostly comes from [this css-tricks article](https://css-tricks.com/document-outline-dilemma/), which is from 2018 (it's long and not important to read for this but very interesting IMO)

>(what if you go 7 pubs deep, should you have an h8?)
i'm not worried about hitting h7 (or the even more mythical h8) here because we're only rendering related pubs to depth 2 anyways. and if we have to collapse all of the lowest level pubs to h6, that's still ok.

Comment (allisonking) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx]: >  I think there should be some kind of check that prevents this from happening (the basic: is this pub already in the map).

👍 done!

for the headings, I've changed it so the "Other fields" heading won't be 'higher' than `h2`, and also wrapped all the pubvalues in `article`, and kept the general pub heading incrementing for now

core/lib/pubs.db.test.ts
allisonking: oh yes this file does test `lib/pubs.ts`, not `lib/server/pub.ts`, which is why I made it a separate file.

I agree having `lib/server/pub.ts` and `lib/pubs.ts` is confusing... my impression was that `lib/server/pub.ts` had all db related functions, while `lib/pubs.ts` had more just helper functions on pubs. Since `getPubByForm` doesn't actually call the db, I put it in `lib/pubs.ts`. though perhaps confusingly, its test _does_ use the db to set up the data...

PR #1081, opened 2025-03-18
fix: fix rendering of markdown with related member value

### LINKED ISSUE #935: action log claims email success when email cannot be sent

~~~1. A community with pre-existing email actions (prior to commit #909 ) has an a email sent when a pub moves into a stage 
2. Pub moves into that stage, triggering the action
3. The action is run with success according to the action log
4. Recipient does not receive email~~~

1. An email action contains a reference to a MemberId of a related pub  `:value{field="unjournal:evaluation-manager" firstName lastName rel="unjournal:evaluations"}`
2. Email action is run
3. Error appears
4. In the Action Log, email status is marked as a success. When hovered over, `{
  "cause": {
    "name": "AssertionError"
  },
  "error": "",
  "title": "Failed to Send Email"
}`

Action log would display "failure" for the email

### What browser/operating system were you using?
Firefox

## Optional
Email does send if the token does not include the relationship (ex: `:value{field="unjournal:evaluation-manager" firstName lastName}`

Davit//Unjournal

### LINKED ISSUE #1074: MemberID tokens with relationships cause emails to fail

1. User has an email action using a MemberID token (ex: `:value{field="unjournal:evaluation-manager" firstName rel="unjournal:evaluations"}`
2. Upon "Run", email fails to send.
3. If the token includes `fullName`, email sends successfully, but prints the MemberID value instead of the user's fullName
4. Action Log reports the following
```
{
  "cause": {
    "name": "AssertionError"
  },
  "error": "",
  "stack": [
    "3226d079-6190-4643-a385-ac7bc95b7ee9"
  ],
  "title": "Failed to Send Email"
}
```

Token would be replaced by specific value. Looks like this is specific to tokens with a relationship field.

https://app.pubpub.org/c/unjournal/pubs/d6918895-1ed1-4082-98b5-419612f34f2f

David//Unjournal

3mcd:
## Issue(s) Resolved

Resolves #1074
Resolves #935 (I think @tefkah fixed this in a recent PR but I confirmed that failing email actions now report a red failure badge)

This PR fixes an issue where markdown templates fail to render when a `:value` directive for a MemberID field was used, specifically when `rel` is used.

## Test Plan

1. Create a simple community with Submission and Evaluation pub types, where the Submission has an "evaluation manager" field (MemberId) and an "evaluations" field (relation field).
2. Create a submission pub and an evaluation pub. Give the submission pub an evaluation manager.
3. Relate the evaluation pub to the submission pub through the "evaluations" field.
4. Run an email action against an evaluation pub. The email action should have a directive like `:value{field="c:evaluation-manager" firstName lastName rel="c:evaluations"}`
5. Verify that the email sends successfully and renders the submission's evaluation manager's full name.


allisonking (approved):
verified it works on forms! https://pr-1081-email-rel-membe-ip-34-224-84-52.my.preview.run/c/ak-test/public/forms/evaluation-default-editor/fill?pubId=18384fee-93db-479e-8beb-2451caaf9df6

if you have time, would be great to add an e2e test for this case too!

Issue #945, opened 2025-02-10
List support


gabestein:
## Motivation

So users can add ordered and unordered lists to their pubs

## Requirements

- Create/switch between ordered and unordered lists
- Support for multiple levels of sub-lists that can themselves be ordered or unordered

## Acceptance Criteria

- One button for unordered lists, one button for ordered lists
- Typing enter once at the end of a list creates a new list item
- Typing tab creates a new sub-list
- Typing shift-tab goes to previous level (or back to P)
- Typing -+space or 1.+space automatically creates ordered or unordered list
- Typing enter in a list and then delete creates a new paragraph within the previous list item, so that users can create block elements within a list

## Mockups (if available)

<img width="796" alt="Image" src="https://github.com/user-attachments/assets/312b0d56-7d25-460e-b600-e02b74849071" />

<img width="788" alt="Image" src="https://github.com/user-attachments/assets/a746cc98-99ee-4b3a-854e-a3abedc69c93" />

PR #1082, opened 2025-03-18
Lists in editor

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/945

Adds lists, both bullet and ordered, to the editor.

## Test Plan

I set up a form with a rich text editor in the preview environment here to play with: https://pr-1082-editor-lists-ip-3-80-61-184.my.preview.run/c/croccroc/public/forms/editor-default-editor/fill

https://github.com/user-attachments/assets/5ca02bb1-1741-4302-8a35-235c644f5c85

## Notes

Because of the way we have a button to open the AttributePanel for each block, the bullet points end up being spaced pretty far apart. 

image

And sublists have their bullets kind of hidden among the borders. I tried to move the bullet markers 'into' the list item, but this caused other formatting issues I haven't been able to quite figure out yet. But if we know we want these to look a certain way, I'll give it another shot!

Also, it seems like using tab/shift+tab to indent/deindent lists might not be the most accessible. I did add it since it's in the ticket requirements and it seems like lots of editors do this. The alternative is Cmd+[ and Cmd+] (which also works here)

### DIFF
--- a/packages/context-editor/package.json
+++ b/packages/context-editor/package.json
@@ -103,6 +103,7 @@
 		"prosemirror-keymap": "^1.2.2",
 		"prosemirror-model": "^1.24.1",
 		"prosemirror-schema-basic": "^1.2.3",
+		"prosemirror-schema-list": "^1.5.1",
 		"prosemirror-state": "^1.4.3",
 		"prosemirror-suggest": "^3.0.0",
 		"prosemirror-transform": "^1.10.0",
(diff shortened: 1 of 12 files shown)


3mcd (approved):
Nice tests! I was pleased to see the tests made up the majority of your PR 🥳 

It works well. The visual noise is pretty bad of course but we will tackle styling of "standard" marks/nodes like bold/italics/lists/etc soon!

PR #1104, opened 2025-03-24
fix: update next to fix CVE

tefkah:
## Issue(s) Resolved

There was a Next CVE that's fixed in the 15.2.3 (https://nextjs.org/blog/cve-2025-29927)

while technically not impacting us as we don't do auth in middleware, probably still good to update.


kalilsn (approved):
🔐

PR #1112, opened 2025-03-25
feat: render submit buttons on edit page

### LINKED ISSUE #1041: Contributor forms should use configured submit buttons
## Motivation

So admins/editors can give contributors the ability to implicitly move pubs/actions, while still restricting their field access

e.g. A legal review contributor can be given access to all pubs on a stage, but only to see the contents, and then, via one of the form buttons, decide whether to "approve" or "send back with feedback" a pub, thus moving the pub to a new stage and triggering an action.

## Requirements

- When a contributor user updates a pub via the update page, it should render both a save button AND the submission buttons configured by the form via which they are accessing the page
- The save button continues to update the pub in place, as it works today
- When a contributor updates the pub via one of the configured submit buttons, it runs the post-submit action specified by the form and displays the post-submission message

3mcd:
## Issue(s) Resolved

Resolves #1041

This PR renders a form's submit buttons on the pub update/edit page when the logged in user is a pub contributor.

In addition the PR hides the stage select dropdown on the pub update/edit page when the logged-in user does not have `movePub` capability.

## Test Plan

1. Add a submit button to the default form of an existing pub type in any community. Configure the submit button with a stage to move into when clicked.
2. Create a pub of this pub type using the create pub button on either the pub/workflows pages or the stage editor panel.
3. Submit the form using the normal Save button. You should see the pub has no stage (or the stage you selected in the form).
4. Submit the form using the new submit button you added to the default form in step 1. You should see the pub now has the stage you specified in step 1.

### DIFF
--- a/core/app/components/pubs/PubEditor/PubEditorClient.tsx
+++ b/core/app/components/pubs/PubEditor/PubEditorClient.tsx
@@ -327,14 +327,19 @@ export const PubEditorClient = ({
 				buttonElements,
 			});
 
-			const newStageId = stageIdFromForm ?? stageIdFromButtonConfig;
-			const stageIdChanged = newStageId !== stageId;
+			const newStageId =
+				stageIdFromForm !== stageId
+					? stageIdFromForm
+					: stageIdFromButtonConfig !== stageId
+						? stageIdFromButtonConfig
+						: undefined;
+
 			let result;
 			if (isUpdating) {
 				result = await runUpdatePub({
 					pubId: pubId,
 					pubValues,
-					stageId: stageIdChanged ? newStageId : undefined,
+					stageId: newStageId,
 					formSlug,
 					continueOnValidationError: autoSave,
 					deleted,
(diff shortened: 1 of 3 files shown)


core/app/components/pubs/PubEditor/PubEditorClient.tsx
```diff
@@ -367,12 +368,10 @@ export const PubEditorClient = ({
...
-				formInstance.reset(
-					{},
-					{
-						keepValues: true,
-					}
-				);
+				formInstance.reset({
+					...formValues,
+					stageId: newStageId ?? stageId,
+				});
```
3mcd, line 374: Previously the selected stage would disappear when the form was reset.

core/app/components/pubs/PubEditor/PubEditorClient.tsx
```diff
@@ -327,8 +327,9 @@ export const PubEditorClient = ({
 				buttonElements,
 			});
 
-			const newStageId = stageIdFromForm ?? stageIdFromButtonConfig;
+			const newStageId = stageIdFromButtonConfig ?? stageIdFromForm ?? undefined;
```
kalilsn, line 330: It's a slightly awkward experience for admins that actually update the stage select since clicking the submit button will ignore what they've selected there now. But I see that you'd probably have to remove the stageId from the default values to make the other order work, so I think this is fine for now!

kalilsn (approved):
🚀

Issue #1002, opened 2025-02-27
Update email action to invite users who don't have accounts


gabestein:
## Motivation

So an admin can send emails with forms to people who create discussions and thus are not (yet) members. In this case, it's to create a discussion verification flow for users where we're collecting email addresses on the front-end and want to create memberships for them as part of that flow.

The other cool thing this does is essentially allows you to create custom magic link flows for different workflows (via API or otherwise). E.g., you can have someone: 
1. Enter their email address on a public-facing site, which creates a pub
2. Have that creation trigger a form invite that simply asks them to verify their account
3. Have submitting that form trigger another form invite to fill out the original form

## Requirements

- If an email action containing an invite is sent to an email address field (as opposed to a member)...
   - If the member for that email already exists, it sends it to that user and allows them to login with the invite link as normal and then redirects them to fill out the form as usual
   - If the member does not exist, it creates an account for them and then sends a link that redirects them to fill out the invited form with the new created account
- Invite can only be used to fill out form from newly created account/email

## Acceptance Criteria

- Action flow: 
   - Action detects no account/member when run, and creates new account for specified email without sending a normal invite email
   - Action creates a form permission for the form/pub pair and the new account
   - Action rewrites form link from direct link to the form to a typical account setup invitation link, adding a `redirectTo=` parameter that redirects to the form with pub ID
   - User clicks on link, and is prompted to finish setting up account
   - After account setup, setup page reads redirect param and forwards logged in user to pub form
 - Update account setup page to accept redirect (if doesn't already)
 - Documentation update to Logging In, Email Action, Membership

gabestein:
Let's discuss. Could also see accomplishing this by creating users via API or even action — maybe having a special function that can take an email field and either creates a new user from that field (and sets a member field to that user) or, if the user already exists, sets a member field to that user?

gabestein:
Another thought: can we solve this to some degree by implementing public forms? That is, a form that prompts a user to login or create an account, and then once they've done so, allows them to submit it? We need this for ICOR as well.

gabestein:
@tefkah we need your thoughts on this one -- the context is in the motivation. Basically, we want to be able to accept a discussion pub from someone with only an email address, but then give them the ability to see/edit the pub they created, and in this case, to trigger a workflow that sends them a form to verify the discussion.

gabestein:
[...]

PR #1114, opened 2025-03-26
feat: allow users to sign up and fill out public forms

### LINKED ISSUE #1051: Allow users to create or login to accounts to fill out forms set to public
## Motivation

So communities can accept e.g. submissions from the public (with minimal spam)

For example, ICOR will want a public submissions form where anyone can submit a project.

## Requirements

- When a form is set to "public" in the ui...
   - Visiting the form page while logged in as any user will load it
   - Visiting the form page not logged in will prompt the user to either login or create an account
- Upon submitting the form, if the user is not already a member of the community, they will be given a Contributor membership to the community and a Contributor membership to the resulting pub

## Acceptance Criteria

- Enable public signup if not already built -- but only in context of filling out a form, not on other pages (for now)

## Extra work

- Change private/invite-only/public dropdown to private <-> public switch. Context here is we are ultimately going to get rid of the concept of "invite-only" forms, and the only setting that matters is whether anyone with an account can fill out a form. But we don't need to make the UI change right now.

### LINKED ISSUE #1053: Add contributor pub membership when user creates pub
## Motivation

So contributors who create new pubs (via public form, related fields, or internal create form) can edit them later using the form they used to create it

## Requirements

- When a user creates a new pub, they are automatically given a contributor membership to it, with a form permission for the form they used to fill it out.

## Acceptance Criteria

- Should create new membership for the user/pub, with formid of form used to create pub

tefkah:
## Issue(s) Resolved

Resolves #1053
Resolves #1051

There is now the possibility for people to sign up to communities without being invited.

Communities that have at least one public form now automatically enable signups at `/c/{communitySlug}/public/signup`. There currently is no way to disable this aside from settings your forms to private.

Users who go to a public form, will now be redirected to the signup page, where they can sign up to the community. Once complete, they will be redirected to the form they tried to access.

Separately from all this: once a user creates a pub using the PubEditor, they will be added as a Contributor to the pub. This allows the user to see the pub in the pub list.

This PR also removes the "Invite Only" access type. Forms which were set to this will be set to `private` instead.

#### Redirect to the created pub

Currently there is no easy way to direct the user to their created pub. I added a new option to the `:link` token that allows you to link to a pub. It goes like `:link(page='currentPub' text='View your pub')`. This will link to the pub that the user just created. By setting the postsubmit screen like i do here

https://github.com/pubpub/platform/blob/1e622788da94c05597f27a586f5df2040bd40bea/core/playwright/formAccess.spec.ts#L126-L131

the user will see a screen like this (note how it links to a direct pub)

Image

We should probably have some redirect action instead, see #1115 for more details.

#### Nicer button

I added a button that automatically does broadly correct loading state showing, see:

https://github.com/user-attachments/assets/e3ff9dcc-9860-4b9e-85a1-9c788a0bfc40

The code is here:

https://github.com/pubpub/platform/blob/8b86f1f18dd978de78d5622e7c0ede14d322066c/core/app/components/SubmitButton.tsx#L1-L176

I think it's nice and we should use it in more places.

## Test Plan

As always, convince yourself of the truth and accuracy of the added tests.

1. Go to https://pr-1114-public-form-u-ip-54-221-104-228.my.preview.run/c/croccroc/public/forms/review/fill while not logged in.
2. Be presented with the signup form, create an account.
3. You should be redirected back to the form.
4. Fill out form and submit.
5. Click the link
6. See that you are redirected to your pub, and see that you are a contributor.

Do maybe the same again after changing the form to private (and logging out). You should not be able to see the form/signup page at all.

Demo:

https://github.com/user-attachments/assets/7db159a4-5cad-42a2-973c-b84a6055ef0b

I changed the spacing of button config form from this

image

to this

image


core/lib/authentication/errors.ts
tefkah, line 1: ideally we would define most errors for server actions like this i think, rather than having a bunch of random `return { error: 'xxxx' }` everywhere

core/lib/__tests__/transactions.ts
tefkah, line 1: i thought i wanted to use this in a playwright test (to do a rollback), which is why i moved it. didn't actually want to turns out, bc if you update things in a transaction, ofc the app will not reflect those changes before you commit

core/app/components/Signup/PublicSignupForm.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export function PublicSignupForm(props: { communityId: CommunitiesId; redirectTo?: string }) {
+	const runSignup = useServerAction(publicSignup);
+
+	const searchParams = useSearchParams();
+	const redirectTo = props.redirectTo ?? searchParams.get("redirectTo") ?? undefined;
+
+	const handleSubmit = useCallback(
+		async (data: SignupFormSchema) => {
+			// TODO: this is not very nice UX, we should wait a sec and show a loading state
```
tefkah, line 21: since the signup function already redirected users i didn't really want to change it, but we should figure out some nice pattern here

core/app/components/Signup/LegacySignupForm.tsx
tefkah, line 1: this is the signup form for users who are invited using the email action to complete their signup. i think we should phase this out and handle those invites the same as invites where we do not create an account yet (#1002)

core/app/components/Signup/JoinCommunityForm.tsx
tefkah, line 1: users who are already logged in should not sign up again, they can just join the community immediately. 
both this and the signup form (or just that page) should show some branding in the future.

core/app/components/FormBuilder/ElementPanel/SelectAccess.tsx
```diff
@@ -8,23 +8,17 @@ import { Contact, Lock, Users } from "ui/icon";
 import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from "ui/select";
 
 const iconsAndCopy = {
-	[FormAccessType.inviteOnly]: {
-		Icon: Contact,
-		description: "Accessible via URL with tracked submissions",
-		name: "Invite Only",
-		help: "Community members & invited contributors can submit",
-	},
```
tefkah, line 16: As per the extra work in #1051 , we now only have private and public forms

core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/page.tsx
Comment (tefkah) [inline core/app/c/(public)/[communitySlug]/public/forms/[formSlug]/fill/page.tsx:1]: this logic is getting very complex. in my next pr i will create some kind of centralized place to handle this, like a `Service` class or something

config/tsconfig/nextjs.json
```diff
@@ -3,7 +3,6 @@
 	"display": "Next.js",
 	"extends": "./base.json",
 	"compilerOptions": {
-		"plugins": [{ "name": "next" }],
```
tefkah, line 6: this removal doesn't really do anything other than make me feel better.

core/app/components/SubmitButton.tsx
```diff
@@ -0,0 +1,183 @@
...
+	successText?: string;
+	errorText?: string;
+
+	// button props
+	className?: string;
+	onClick?: () => void;
+	type?: "button" | "submit" | "reset";
+};
+
+export const SubmitButton = ({
```
allisonking, line 33: 🤩

core/prisma/schema/schema.prisma
```diff
@@ -680,3 +683,46 @@ model MembershipCapabilities {
...
+  status    String    @default("active")
+  revokedAt DateTime?
+
+  InvitedByUser        User?      @relation("invited_by", fields: [invitedByUserId], references: [id], onDelete: SetNull)
+  invitedByUserId      String?
+  InvitedByActionRun   ActionRun? @relation("invited_by_action_run", fields: [invitedByActionRunId], references: [id], onDelete: SetNull)
+  invitedByActionRunId String?
+
+  @@map(name: "invites")
+}
```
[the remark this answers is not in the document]
tefkah: have taken it out

allisonking:
I tested using the preview environment, though I wasn't able to see the pub once I hit submit on the form. I saw the 'Go see your pubs here' link, and clicked here, but then got redirected to sign in. I tried to sign in but it didn't work (this might just be because I didn't remember my temp password right). the pub I made is there though: https://pr-1114-public-form-u-ip-54-221-104-228.my.preview.run/c/croccroc/pubs/84e66eea-563f-4c2c-ba90-fcfd1f781cea

I'll try locally too to see if maybe it's just something is up with the preview env

allisonking:
well it seems to work locally! although I can't seem to change a form to private so haven't been able to test the 'no forms should show up' state yet

https://github.com/user-attachments/assets/d7ece2f8-cda3-425a-867d-c2afac67bfa8

tefkah:
huh that's strange, ill have a look!

tefkah:
The "not being able to change access types" is a bug in the formbuilder, you need to change something else about the form in order for it to change. ill put up a quick pr which addresses that (and some other thing), but i don't want this pr to be even bigger!

allisonking (approved):
ah okay great! I confirmed that I can make private if I also add another form element, and then the not-signed-in user can't see it 👍 nice work!

Issue #1006, opened 2025-02-27
Paginate related pubs interfaces/queries in forms


gabestein:
## Motivation

So users can attach any pub as a related pub in a form.

## Requirements

- Updates the form relationship input config to allow the admin to specify which types of pubs can be added to the field. If no types are specified, users can select from all types.
- As a follow on to #1003, updates the interface of related pub form fields so that users can select from any pub of the types specified in the input config.
- Orders pub list by most recently updated
- Removes currently non-functional asc/desc/hide buttons from table

## Acceptance Criteria

- Uses the action config-style multi-select component to select pub types (see mocks)
- Makes sure list is deduplicated

## Extra work (but seriously if you could it's so small and drives us crazy)

- Update multi-select component so that clicking outside of the multi-select dropdown closes it.
- Remove the "x" close button and have the carrot flip (extra extra credit)
- Fix the bug where the dropdown doesn't scroll via the mouse

## Mockups (if available)

![Image](https://github.com/user-attachments/assets/d7fb60b6-b7c2-4925-87c9-dbc8411ca818)

eastofesten:
nice to have if these are `orderBy` some value

tefkah:
> Uses the search interface from https://github.com/pubpub/platform/pull/919 to allow users to search pubs to add to the field. The search should only return results from the pub types specified in the input config

this cannot filter by pubtypes yet, but should be very easy to add.
notably the fulltextsearch functionality also does not yet do any permission stuff, but i think that's fine for now right?

allisonking:
> Updates the form relationship input config to allow the admin to specify which types of pubs can be added to the field. If no types are specified, users can select from all types.

This is to add a 'pub type' selector to the form builder, right? I did a first pass of that in https://github.com/pubpub/platform/pull/900 but @tefkah made a good point that we should discuss how we want to link pubtypes here. the 'easiest' way, in that the form builder is already set up to do this, is to plop it in `config` but we probably actually want a real FK here, so maybe a new column in the `form_elements` table? Thomas also suggested making this an array of possible pubtypes to stay flexible for the future—any reason not to do this?

allisonking:
[...]

PR #1116, opened 2025-03-26
Paginate tables either server side or client side

allisonking:
## Issue(s) Resolved
Part of https://github.com/pubpub/platform/issues/1006

Adds https://github.com/sadmann7/shadcn-table to paginate a server side table, but also adapts it so that we can paginate on the client too (by issuing queries via react query).

## Test Plan

Try adding/removing related pubs to James McJimothy: https://pr-1116-paginate-tables-ip-54-84-241-40.my.preview.run/c/arcadia-research/pubs/d8180174-bbfc-472d-9ad0-1bcf6f83aa62/edit

And also to a new form that has a related field. View the network tab to verify that the paging is happening on each page click. Sorting and "row count" should work too

image


core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts
```diff
@@ -350,7 +350,7 @@ const handler = createNextHandler(
 				const { user, community } = await checkAuthorization({
 					token: { scope: ApiAccessScope.pub, type: ApiAccessType.read },
 					// TODO: figure out capability here
-					cookies: false,
+					cookies: true,
```
Comment (allisonking) [inline core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts:353]: I'm not sure what's supposed to go here but I imagine it shouldn't just be `true` 😅

core/app/c/[communitySlug]/pubs/page.tsx
```diff
@@ -33,6 +34,11 @@ export default async function Page(props: Props) {
 
 	return (
 		<>
+			<PubsDataTableServer
+				communityId={community.id}
+				userId={user.id}
+				searchParams={searchParams}
+			/>
```
Comment (allisonking) [inline core/app/c/[communitySlug]/pubs/page.tsx:41]: Plopped this in the all pubs page just as an example. Seems to work decently well, with the caveat that the table is only rendering title atm. would probably be more complicated once we want it to start rendering action menus

image

core/app/components/DataTable/PubsDataTable/PubsDataTableClient.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+interface WithData {
+	total: number;
+	data: ProcessedPub<{ withPubType: true }>[];
+}
+type PubsDataTableProps = {
+	perPage: number;
+} & (WithPromises | WithData);
+
+export const PubsDataTable = ({ perPage, ...props }: PubsDataTableProps) => {
```
allisonking, line 27: made this table able to accept promises or data. it was difficult to get react query to return a promise to keep the interface exactly the same

Seems to work okay with the query params, although I'm not sure we want to use query params here, maybe just state instead? query params would make it difficult to render two of the same component on the page (not that we are doing that for pub tables anywhere atm)

image

packages/ui/src/data-table-paged/components/data-table-pagination.tsx
```diff
@@ -0,0 +1,92 @@
+import type { Table } from "@tanstack/react-table";
```
allisonking, line 1: from here down, mostly copied from https://github.com/sadmann7/shadcn-table

core/app/components/DataTable/PubsDataTable/PubsDataTableClient.tsx
tefkah: yes i think we should use state for the react-query variation. although i think it could be fine if we used the `shallow: true` in that case (this prevents the page from reloading all the data on searchParam change). if we do though, we should prefix the parameters with something st we can distinguish different usages of the component on the same page.

but my gut feel is to just leave it out, just use state for the react query version!

tefkah: the main consideration i would have: in which situations would we use the client component version, _and_ would want the state of the page to be preserved on eg reload. currently i don't see that many cases for it

core/app/components/DataTable/PubsDataTable/PubsDataTableClient.tsx
```diff
@@ -0,0 +1,90 @@
...
+			},
+		},
+	});
+
+	const pubs = data?.body;
+	if (!pubs) {
+		return null; // TODO: empty state
+	}
+
+	// TODO: figure out how we should get total here
```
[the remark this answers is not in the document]
tefkah: oo i kind of like passing that through a header! or else we could return `{ data: ProcessedPub[], total: number }` or something, but that would be kind of different from how we do everything else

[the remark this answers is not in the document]
tefkah: i think for now a header is good, maybe for a `v1` version of the api we can think of a more holistic approach

core/app/components/DataTable/PubsDataTable/PubsDataTableClient.tsx
```diff
@@ -0,0 +1,89 @@
...
+				withRelatedPubs: false,
+				withStage: true,
+				withValues: false,
+				withLegacyAssignee: true,
+			},
+			params: {
+				communitySlug: community.slug,
+			},
+		},
+	});
```
tefkah, line 80: this currently is not being paginated right? or is that included in `filterParams`?

core/app/components/DataTable/PubsDataTable/columns.tsx
```diff
@@ -0,0 +1,72 @@
...
+				<Checkbox
+					checked={row.getIsSelected()}
+					onCheckedChange={(value) => row.toggleSelected(!!value)}
+					aria-label="Select row"
+					className="translate-y-0.5"
+				/>
+			),
+			enableSorting: false,
+			enableHiding: false,
+		},
```
tefkah, line 48: idk why but in all the screenshots you shared the checkbox column seems so wide, is that bc of the translate?

tefkah:
some more thoughts (mostly ui related)

- do you think there's a way to "natively" support the pubtype selector we have for the relatedpubs table in this table? its probably just a filter you can select, but i think that would be nicer to have that native to the table itself rather than it being an extra component we need to wrap it with
- if not filtered by pubtype, we should show the pubtype in the table i think
- the "Add related pubs" panel does become quite cramped this way. can we just make it a little bigger by default?
- should probably have some option to only show the "select" checkboxes if there's something you can do with it
- maybe would be cool to be able to pass some styling options easily, eg to not show the lines. that way this component could also easily serve as a plain "publist"
- having some idea of what pubValue columns would look like would be nice. those would probably need to be set dynamically from eg the pubtype, but could also be from a form or something.

but these are nice to haves, not need to haves!

core/app/components/DataTable/PubsDataTable/PubsDataTableClient.tsx
allisonking: yes, included in `filterParams`! I just wrapped it in `getFilterParamsFromSearch` so that both the server and the client side versions could get the same parameters

core/app/components/DataTable/PubsDataTable/columns.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { dateFormatOptions } from "~/lib/dates";
+import { getPubTitle } from "~/lib/pubs";
+
+interface GetColumnsProps {
+	setRowAction?: React.Dispatch<
+		React.SetStateAction<DataTableRowAction<ProcessedPub<{ withPubType: true }>> | null>
+	>;
+}
+
+export function getColumns({
```
allisonking, line 21: I think once we use this table more, we should pass different columns in. but here's an example of the simple columns we need for adding related pubs (just the name and `updatedAt` and checkboxes)

core/app/components/forms/elements/RelatedPubsElement.tsx
```diff
@@ -227,40 +238,33 @@ export const RelatedPubsElement = ({
...
-				return acc;
-			},
-			{} as Record<string, ContextEditorPub>
-		);
-	}, [pubs]);
-
-	const linkedPubs = fields.map((f) => f.relatedPubId);
-	const linkablePubs = pubs
-		// do not allow linking to itself or any pubs it is already linked to
-		.filter((p) => p.id !== pubId && !linkedPubs.includes(p.id));
```
allisonking, line 243: note: our table's behavior has changed a bit (I think for the better!) where instead of filtering out already linked pubs, we still show them in the table, but they are checked now. however, we don't have an interaction for removing from this table at the moment... i.e. unchecking a box doesn't 'delete' the value right now. I don't think that'd be hard to add though

core/app/components/DataTable/PubsDataTable/PubsDataTableServer.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { PubsDataTable } from "./PubsDataTableClient";
+import { getFilterParamsFromSearch, searchParamsCache } from "./validations";
+
+type Props = {
+	communityId: CommunitiesId;
+	userId: UsersId;
+	searchParams: { [key: string]: string | string[] | undefined };
+};
+
+export const PubsDataTableServer = async ({ communityId, userId, searchParams }: Props) => {
```
allisonking, line 13: Not currently used anywhere, but should be ready to go when needed, with possible adjustments to `getColumns`.

core/app/components/forms/AddRelatedPubsPanel.tsx
```diff
@@ -94,10 +40,10 @@ export const AddRelatedPubsPanel = ({
 				<Button
 					type="button"
 					data-testid="add-related-pub-button"
-					onClick={handleAdd}
+					onClick={handleUpdate}
 					className="flex-1 bg-blue-500 hover:bg-blue-600"
 				>
-					Add
+					Update
```
allisonking, line 46: Note: I changed this panel's behavior so that it 'updates' rather than only 'add's. previously, the table only showed pubs that weren't already linked. Now, since we are just showing all the pubs from `getPubsWithRelatedValues`, we show all pubs, and the ones that are linked are checked. I think this behavior makes a little more sense, but I did have to add logic for removing from based on unselecting rows

allisonking:
I think this is in an okay state now. Still doesn't do a good chunk of https://github.com/pubpub/platform/issues/1006 but we are getting closer! Didn't implement all of @tefkah 's suggestions since things were already getting large for this PR, but I think the pub type filtering at least will be done in my follow up PR as part of https://github.com/pubpub/platform/issues/1006

core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts
Comment (tefkah) [reply core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts:354]: hmm true might actually be fine i think! we filter the pubs by the user anyway, so i don't think allowing blanket access to this endpoint is that bad. i think maybe we should at least check that the user is a member of the community (i think currently `cookies: true` just means that the user must be logged in). as extra work, could you maybe add an option for cookies to be set to "community-member", and if it is, it will check in `checkAuthorization` if the user object returned from `getLoginData` has a membership for the current community on it?

Comment (tefkah) [reply core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts:354]: (dont do this for this pr, you can do that in the follow up if you want)

tefkah:
small issue, maybe to add for next PR: could you filter out the current pub from the shown list? currently this is not possible to do through the api, so just doing it on the client is fine

core/app/components/forms/AddRelatedPubsPanel.tsx
tefkah: yes, i noticed this, very nice! that's indeed how i would expect it to work

core/playwright.config.ts
```diff
@@ -38,7 +38,7 @@ export default defineConfig({
 		},
 	],
 	// max 30 seconds per test in CI
-	timeout: process.env.CI ? 30 * 1000 : 10 * 60 * 1000,
+	timeout: process.env.CI ? 30 * 1000 : 60 * 1000,
```
tefkah, line 41: 🙏

packages/contracts/src/resources/site.ts
```diff
@@ -37,6 +37,8 @@ import {
 import type { Json, JsonValue } from "./types";
 import { CreatePubRequestBodyWithNulls, jsonSchema } from "./types";
 
+export const TOTAL_PUBS_COUNT_HEADER = "x-total-pubs";
```
tefkah, line 40: nice, thanks for explicitly exporting the header name. very annoying we can't set this in `ts-rest` tho

tefkah (approved):
looks really good, great job!

allisonking:
Made the current pub disabled:

image

PR #1129, opened 2025-04-01
Add pubtypes to related pub form config for filtering

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/1006


core/lib/server/form.ts
```diff
@@ -64,7 +69,12 @@ export const getForm = (
 							"pub_fields.slug",
 							"pub_fields.isRelation",
 							"pub_fields.name as fieldName",
+							eb.fn
+								.jsonAgg(eb.ref("_FormElementToPubType.B"))
+								.filterWhere("_FormElementToPubType.B", "is not", null)
+								.as("relatedPubTypes"),
 						])
+						.groupBy(["form_elements.id", "pub_fields.id"])
 						.$narrowType<FormElements>()
```
allisonking, line 78: something is going wrong in this `narrowType`, and probably has to do with how I'm specifying the `relatedPubTypes` in `FormElements`. I can't seem to get the type union there to work right. here's what `relatedPubType` resolves to currently:

image

or if I change 
https://github.com/pubpub/platform/blob/6a8e9486dfd76b995b695bbde4a4edd9d5abafef/core/app/components/forms/types.ts#L42

to `relatedPubTypes: PubTypesId[]` then the error becomes slightly better

image

but `relatedPubTypes: PubTypesId[] | null` is the 'right' type, I think

packages/ui/src/multi-select.tsx
```diff
@@ -134,23 +136,14 @@ export const MultiSelect = React.forwardRef<HTMLButtonElement, MultiSelectProps>
...
 				open={isPopoverOpen}
-				// open={isPopoverOpen}
-				// onOpenChange={(open) => {
-				// 	console.log("GOING TO", open);
-				// 	setIsPopoverOpen(open);
-				// }}
 				onOpenChange={(open) => {
-					if (isPopoverOpen && !open) {
-						return;
-					}
```
allisonking, line 145: this was making the popover not close when clicking outside of it. I'm not sure if removing this breaks something else though... I played around and didn't find anything though

gabestein: !!! (someone who knows should comment but omg, tyty, this wastes so much of my time)

core/app/components/FormBuilder/actions.ts
```diff
@@ -16,6 +23,23 @@ import { autoRevalidate } from "~/lib/server/cache/autoRevalidate";
 import { findCommunityBySlug } from "~/lib/server/community";
 import { defineServerAction } from "~/lib/server/defineServerAction";
 
+const upsertRelatedPubTypes = async (
```
allisonking, line 26: this is maybe not the most efficient but I think it's the easiest to reason about. let me know if there's a better way to do this though, since this requires a few db calls

core/app/components/FormBuilder/actions.ts
```diff
@@ -42,18 +66,32 @@ export const saveForm = defineServerAction(async function saveForm(form: FormBui
...
 		(acc, element, index) => {
 			if (element.deleted) {
 				if (element.elementId) {
 					acc.deletes.push(element.elementId);
 				}
 			} else if (!element.elementId) {
-				// Newly created elements have no elementId
-				acc.upserts.push(formElementsInitializerSchema.parse({ formId, ...element }));
+				// Newly created elements have no elementId, so generate an id to use
+				const id = crypto.randomUUID() as FormElementsId;
```
allisonking, line 82: note: now setting the id here instead of letting the db to do it so that we can have access to the id when creating the pub type relationship too

core/app/components/FormBuilder/ElementPanel/ComponentConfig/RelationBlock.tsx
```diff
@@ -42,6 +45,35 @@ export default ({ form }: ComponentConfigFormProps<InputComponent.relationBlock>
...
 			/>
+			<FormField
+				control={form.control}
+				name="relatedPubTypes"
+				render={({ field }) => (
+					<FormItem>
+						<FormLabel>Pub Type</FormLabel>
+						<MultiSelect
+							className="bg-white"
+							value={[]}
```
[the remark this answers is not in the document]
allisonking: you're right! I had copied this from another component that was using `MultiSelect` and I think neither is right—`value` doesn't do anything here, it's really `defaultValue` that does

PR #1130, opened 2025-04-01
Handle errors in memoize

kalilsn:
## Issue(s) Resolved
Hopefully the regular 503 errors from large query responses: see https://knowledgefutures.slack.com/archives/CMAQM0BNV/p1743537155616299

There's a maximum size to data we can store with unstable_cache. This PR adds a try/catch around our unstable_cache calls and logs when an error is thrown

## Test Plan
Deploy it and see if @isTravis can still break the app with a single request

### DIFF
--- a/core/lib/server/cache/memoize.ts
+++ b/core/lib/server/cache/memoize.ts
@@ -2,6 +2,7 @@
 
 import { cache } from "react";
 import { unstable_cache } from "next/cache";
+import * as Sentry from "@sentry/nextjs";
 
 import { logger } from "logger";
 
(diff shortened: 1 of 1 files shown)


core/lib/server/cache/memoize.ts
```diff
@@ -93,46 +94,60 @@ export function memoize<P extends unknown[], R>(cb: Callback<P, R>, opts?: Memoi
...
+				const fn = { name: cb.name, body: cb.toString() };
+				logger.error({
+					msg: "⚠️ Memoize: unstable_cache failed, using uncached function",
+					error,
+					args,
+					fn,
+					additionalCacheKey,
+				});
+				Sentry.captureException(error, {
+					extra: { args, fn, additionalCacheKey },
```
kalilsn, line 144: We don't actually have any info to identify the original calling function here when this is being called with autocache (`cb` will be an anonymous function wrapping the actual query fn) , but the `additionalCacheKey` should include the compiled sql which will help us narrow it down. if we really wanted to get the name, we'd need to create an error in `autoCache`, parse the calling fn name from the stack trace, then pass it to `directAutoOutput` and set it as a sentry tag there. felt like overkill for now

Issue #1037, opened 2025-03-06
Contributor Permissions Update


gabestein:
## Motivation

So admins can restrict contributors to viewing/editing pubs using specific forms -- implicitly, this means they can also only see/edit pubs of the types of forms they have access to.

## Requirements

- Pub-Level Contributors:
   - Can view/edit pubs they’ve explicitly been given access to via forms they’ve explicitly been given access to. If no form is specified for the pub, they can only view the pub, using the default form.
    - Can add other contributors to pubs via the forms they have access to
    - Can create/add/see related pubs via fields explicitly added to the form they have access to
    - Can delete/archive pub
- Stage-Level Contributors
   - Can do everything a pub-level contributor can but for all pubs in the stage via forms they’ve explicitly been given access to at the stage. If no forms are specified, they can only view pubs in the stage, using the default form.
   - If they have access to multiple forms, can switch between them using the switcher in #1036 
- Community-Level Contributors
   - Can create new pubs via the forms they've explicitly been given access to at the community level.
   - CANNOT see/view all pubs with forms they've been given access to (still feels a little awk but let's go with it)

kalilsn:
> * Pub-Level Contributors:
>   * If forms are specified:
>     * Can add other contributors to pubs (a bit awkward because it's on the view for now)

should this be "Can add other contributors to *the pub*"? if not, does it mean all pubs with the pub type(s) of the form(s)?

>     * Can create/add/see related pubs via fields explicitly added to the form they have access to

create/add shouldn't be a problem. seeing related pubs is still tricky, since it makes calculating whether a user has view access a potentially quite expensive query. right now we're just handling that by embedding the related pubs into a the pub details page, to depth 2. if that's not sufficient anymore, we could allow configuration of the depth/whether to embed related pubs at all at the form_element level as part of the new view configuration

> * Community-Level Contributors
>   
>   * If no forms are specified, can only see/edit pubs they have access to via pub or stage memberships
>   * If forms are specified, can create new pubs via specified forms at the community level ONLY (e.g., cannot create a pub via a "review form" if they only have access to it in a stage)

when a contributor creates a pub, should we give them some kind of membership to it? maybe that's a new setting on the submit buttons or the form itself?

gabestein:
> should this be "Can add other contributors to the pub"? if not, does it mean all pubs with the pub type(s) of the form(s)?

Yup, good catch, thanks! Updated.

[...]

PR #1134, opened 2025-04-03
Add form selection to memberships

### LINKED ISSUE #1038: Contributor Form Access Membership Selector
## Motivation

- So admins/editors can give contributors access to pubs via specific forms at the community, stage, and pub level, and when inviting new users

## Requirements

- Adds combobox for attaching form to a contributor's membership at every level
- Adds explanatory text, differing slightly by level (see mockups/written text below)
- Searchable combobox allows user to add any forms that match the type to the membership -- or none in case of pub level

## Mockups (if available)
Actual text to use in description:

Image

**Pub Level**
Select the forms via which this member can edit and view this Pub. If no form is selected, they will only be able to view the Pub, and will only see fields added to the default Pub form for this type. 

**Stage Level**
Select the forms via which this member can edit and view Pubs in this stage. If no form is selected, they will only be able to view Pubs in this stage, and will only see fields added to the default Pub form for a each Pub type.

**Community Level**
Selecting forms will give the member the ability to create Pubs in the community using the selected forms. If no forms are added, the contributor will not be able to create any Pubs, and will only be able to see Pubs they have access to either directly or at the stage level.

kalilsn:
## Issue(s) Resolved
Resolves #1038
#1037 
This PR makes it possible to associate forms with a contributor membership to a pub, stage, or community when adding a member. This PR does not yet add those options to the MemberSelect forms (used in actions and member select fields), but they have been updated so they will keep working with the new schema. It also removes the form_memberships table, and replaces those memberships with either pub contributor or stage contributor memberships. 

As part of this change, I updated the database schema/queries so users can now have multiple memberships to the same pub/stage/community, provided the formId varies between them. In the future we should make it possible to update a user's memberships through the form, but for now, I've just made sure the existing UI for displaying members continues to work and only renders a user once. 

Memberships should continue to work as they did before, since I haven't yet updated `userCan` to operate differently. A followup PR will handle the rest of that and close out #1037.

image


core/lib/server/form.ts
```diff
@@ -83,43 +84,72 @@ export const userHasPermissionToForm = async (
...
+					.$if(Boolean(props.formSlug), (eb) =>
+						eb
+							.innerJoin("forms", "forms.id", "community_memberships.formId")
+							.where("forms.slug", "=", props.formSlug!)
+					)
+					.$if(Boolean(props.formId), (eb) =>
+						eb.where("community_memberships.formId", "=", props.formId!)
+					)
+					.select("community_memberships.id")
+			).executeTakeFirst()
```
kalilsn, line 143: Separately handling the case where there's no pubId lets us separately authorize creating pubs (community contributor membership with this form) and updating a specific existing pub (pub contributor membership with this form). We weren't doing that before, so everyone who was invited to edit a specific pub could also create new pubs with that form. Merging this PR will mean that all users who were invited to fill out forms via the email action can no longer use the create form. If that's a problem, let me know and we should discuss how to migrate them!

At the moment, the way to grant a contributor access to use a specific form for pub creation is to use the members page to grant them a community contributor membership with the desired form. We should probably add markdown syntax to differentiate between inviting a user to fill out a create form vs updating a specific pub to make the email action usable.

core/playwright/externalFormInvite.spec.ts
```diff
@@ -224,11 +225,26 @@ test.describe("Inviting a new user to fill out a form", () => {
...
 		const unauthorizedPubsPage = new PubsPage(newPage, "croccroc");
 		await unauthorizedPubsPage.goTo();
-		expect(await newPage.url()).toMatch(/\/settings$/);
+		newPage.waitForURL(/\/settings$/);
+		// expect(newPage.url()).toMatch(/\/settings$/);
 
-		// Creating a pub without a pubId should work
-		const createPage = decodedUrl.replace(`pubId%3D${community.pubs[0].id}`, "");
-		await newPage.goto(createPage);
+		// Creating a pub without a pubId should not work
```
kalilsn, line 231: We were explicitly testing the ability for a user invited to edit a specific pub to _also_ create new pubs. Now this tests the opposite.

core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts
```diff
@@ -319,6 +319,27 @@ const manuallyParsePubFilterQueryParams = (url: string, query?: Record<string, a
...
+
+				if (!form) {
+					return { status: 404, body: "Form not found" };
+				}
+				return {
+					status: 200,
+					body: form,
+				};
+			},
+		},
```
Comment (kalilsn) [inline core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts:342]: This isn't actually used, I just realized it's here. It was going to be for the form switcher but it turns out that's also not going to be necessary so I'll remove it

core/app/c/[communitySlug]/stages/manage/components/panel/StagePanelSheet.tsx
```diff
@@ -28,6 +28,7 @@ export const StagePanelSheet = (props: Props) => {
 	return (
 		<Sheet open={open} onOpenChange={onOpenChange}>
 			<SheetContent className="max-h-100vh overflow-y-auto sm:max-w-md">
+				<SheetTitle className="sr-only">Edit Stage</SheetTitle>
```
Comment (kalilsn) [inline core/app/c/[communitySlug]/stages/manage/components/panel/StagePanelSheet.tsx:31]: I added this as sr-only since it was to fix an accessibility warning, although I think it would be reasonable to have a visible title here too

kalilsn:
@tefkah Ok I finally actually fixed the tests after switching to the multi-select component. LMK if there are other issues!

core/app/c/[communitySlug]/stages/manage/components/panel/StagePanelSheet.tsx
Comment (tefkah) [reply core/app/c/[communitySlug]/stages/manage/components/panel/StagePanelSheet.tsx:31]: thanks! that error was driving me crazy

core/app/components/Memberships/MembersList.tsx
```diff
@@ -13,9 +17,23 @@ export const MembersList = <T extends TargetId>({
...
+					dedupedMembers.set(member.id, m);
+				}
+			}
+		}
+		return dedupedMembers;
+	}, [members]);
 	return (
 		<>
-			{members.map((user) => (
+			{[...dedupedMembers.values()].map((user) => (
```
tefkah, line 36: not a blocker as i think we should just get this in, but i think it would prefer it if we just returned a single membership from the db with an optional `forms` property holding the forms the user has access to. i think that's nicer than having to do something like this on the client! but let's just get this in first, we can do that when we get to it (unless ofc you think we shouldnt do it at all)

core/lib/db/queries.ts
```diff
@@ -57,6 +57,7 @@ export const getStageMembers = cache((stageId: StagesId) => {
 			.innerJoin("users", "users.id", "stage_memberships.userId")
 			.select(SAFE_USER_SELECT)
 			.select("stage_memberships.role")
+			.orderBy("users.id")
```
tefkah, line 60: i don't really care, but is this necessary? kind of a weird thing to order on imo

core/app/components/Memberships/MembersList.tsx
kalilsn: I agree, this is ugly and a temporary fix. How about you review my PR for #1138 when it's up and see if you still want me to change the query?

kalilsn: actually what i really mean is: that sounds like it's probably a good idea and i will most likely do it in #1138 😆

tefkah (approved):
looks good!

i have one optional suggestion: since this is a pretty big change to permissions, and i think it would be good if we started doing an (internal) changelog again, could you maybe create a `changelog.mdx` in `docs` with just a single entry like

```md
### <date>/<pr number> Form ids on memberships

Contributor Memberships now can have form ids. This means
- Community: user is able to create pubs with that form
- Stages: user is able to create and edit pubs in this stage with that form
- Pub: user is able to edit and view pub with that form 

You can have >1 contributor membership if they have forms.
```

only if you think this is also a good idea!

core/prisma/migrations/20250402192249_add_form_id_to_all_memberships/migration.sql
```diff
@@ -0,0 +1,106 @@
...
+    "formId" IS NULL;
+
+-- CreateIndex
+CREATE UNIQUE INDEX "stage_memberships_stageId_memberGroupId_formId_key" ON "stage_memberships" ("stageId", "memberGroupId", "formId")
+WHERE
+    "formId" IS NOT NULL;
+
+CREATE UNIQUE INDEX "stage_memberships_stageId_memberGroupId_key" ON "stage_memberships" ("stageId", "memberGroupId")
+WHERE
+    "formId" IS NULL;
```
tefkah, line 106: sorry, i forgot about this: could you also add constraints that only `role = contributor` can have `formId`?

kalilsn: good call!

core/lib/db/queries.ts
kalilsn: Yeah this was the reason: new members were being added to the panel at a random point. But ordering by id isn't much better, so I changed it to createdAt

PR #1135, opened 2025-04-03
fix: improve formbuilder logic

### LINKED ISSUE #1143: Bug: cannot make form public
1. Tried to make a form public
2. Saving reported successful, but on refresh get the "change not saved" error and when the page reloads, form is again set to private and I am not prompted to login or signup when visiting the `/fill` page while logged out.

Form should save as public and I should be prompted to login or signup when visiting the `/fill` page for the public form

tefkah:
- **fix: improve client side logic for formbuilder**
- **fix: be able to submit access changes on their own**

## Issue(s) Resolved

Resolves #1143

Roughly does 3 things:
1. Moves the form builder diff calculation on the client, that way it's easier to only send up changes that are relevant, and we disable the submit button if no changes have been made
2. Updates `access` if that is the only thing changed. Previously also the formelements needed to have been moved/deleted/added in order for a change in access to go through
3. slight beauty enhancements

### Moving diff to client

This was put as TODOs in the code in some places, and it made my life easier, so i did.

The main change to the diffing algorithm is here:

https://github.com/pubpub/platform/blob/23a047bdf927b2c8a4d10f42a41626b6976a66b1/core/app/components/FormBuilder/FormBuilder.tsx#L155-L164

May be a better way of doing this but i thought this was fine for now (i expect the order of keys to be relatively stable).

Basically, if nothing except maybe the id/updatedAt of the form element has changed, we do not consider it a change.

> [!NOTE]
> Caveat: A "rank" change is still considered a change. If two elements are swapped back and forth, their rank may still have been changed. I find it hard to check for this case as i'll admit i don't fully understand the ranking implementation, maybe @kalilsn you have some insight on how to ignore swapping back and forth cases.

> [!WARNING]
> Moving the diffs to the client now does carry slightly more risk for weird stuff happening, but not really that much more than before

### Update logic changes

Since all the changes now arrive as separate entities, we can more easily only update form accesstype if that has changed. 

### QoL changes

- When hovering over the drag handle, we now show `cursor-drag` rather than `cursor-pointer`
- When dragging, we now show `cursor-dragging`, rather than `cursor-pointer`
- When dragging, the dragged element no longer sometimes slides under other elements when dragged.

Demo:
(slightly hard to see bc recording my screen apparently makes the cursor almost always be the default one)

https://github.com/user-attachments/assets/c39cf47f-0c8c-4545-8e8c-1934732c2a9e


core/app/c/[communitySlug]/forms/[formSlug]/edit/page.tsx
```diff
@@ -28,14 +28,7 @@ export default async function Page(props: {
...
 		communitySlug: string;
 	}>;
-	searchParams: Promise<{
-		unsavedChanges: boolean;
-	}>;
 }) {
-	const searchParams = await props.searchParams;
-
-	const { unsavedChanges } = searchParams;
-
```
Comment (tefkah) [inline core/app/c/[communitySlug]/forms/[formSlug]/edit/page.tsx:38]: this logic does not need to happen server side, we can just do it client side. this is much faster

core/app/components/FormBuilder/FormBuilder.tsx
```diff
@@ -107,13 +114,98 @@ type Props = {
...
+					// check whether the element is reeeaally updated minus the updated field
+					const { updated: _, id: _id, ...elementWithoutUpdated } = element;
+					const { updated, id, ...rest } =
+						defaultValues.elements.find((e) => e.elementId === element.elementId) ?? {};
+
+					const defaultElement = rest as Omit<FormElementData, "updated" | "id">;
+
+					if (JSON.stringify(defaultElement) === JSON.stringify(elementWithoutUpdated)) {
+						return acc;
+					}
```
tefkah, line 164: very sophisticated diffing. please recommend a better approach if you know one

core/app/components/FormBuilder/FormBuilder.tsx
```diff
@@ -107,13 +114,98 @@ type Props = {
...
+};
+
 export function FormBuilder({ pubForm, id, stages }: Props) {
-	const router = useRouter();
-	const pathname = usePathname();
-	const params = useSearchParams();
-	const form = useForm<FormBuilderSchema>({
-		resolver: zodResolver(formBuilderSchema),
-		values: {
+	const [isChanged, setIsChanged] = useIsChanged();
```
tefkah, line 205: this is a small custom `nuqs` hook reads from/updates the `isChanged` query param. using nuxt we can do this without a round trip to the server, which we want.

https://github.com/pubpub/platform/blob/415de673452e7952872c1d341ebfc01f3da8054c/core/app/components/FormBuilder/useIsChanged.tsx#L1-L13

core/app/components/FormBuilder/SaveFormButton.tsx
```diff
@@ -18,7 +21,7 @@ export const SaveFormButton = ({ form, className, disabled }: Props) => {
 			form={form}
 			type="submit"
 			data-testid="save-form-button"
-			disabled={disabled}
+			disabled={disabled != null ? disabled : !isChanged}
```
tefkah, line 24: maybe i should just remov the `disabled` prop entirely and make it `!isChanged` by default

core/app/components/FormBuilder/FormBuilder.tsx
```diff
@@ -107,13 +114,98 @@ type Props = {
...
+							formId: formValues.formId,
+							id: element.elementId,
+						})
+					); // TODO: only update changed columns
+					if (element.relatedPubTypes) {
+						// If we are updating to an empty array and there were related pub types before, we should clear out all related pub types
+						if (
+							element.relatedPubTypes.length === 0 &&
+							defaultElement.relatedPubTypes?.length
+						) {
```
tefkah, line 178: i added this `defaultElement.relatedPubTypes?.length` check, bc otherwise if you swapped a relation element back and forth it would assume something had to be deleted, which didn't make sense

allisonking (approved):
woo this is awesome! approving, though I don't know about the ranking part either though, so maybe wait for Kalil to weigh in on that part?

Issue #948, opened 2025-02-10
Strikethrough support/button


gabestein:
## Motivation

So users can add strikethrough text

## Requirements

- Apply/unapply strikethrough text at cursor/highlight

## Mockups (if available)

<img width="796" alt="Image" src="https://github.com/user-attachments/assets/be5352d1-8e90-48e7-b3a8-e35154c23f79" />

<img width="780" alt="Image" src="https://github.com/user-attachments/assets/24798f47-abed-44de-96ee-33e0284b830c" />

allisonking:
this isn't part of the AC (maybe for a good reason!) but wanted to make a note that I started adding the markdown shortcut `~stuff~` which works ok, but is a weird interaction with the context menu

https://github.com/user-attachments/assets/f351b13e-607a-475e-b0f7-c2510a965cac

so I'm going to remove it, but it'd be easy to put back if we wanted it

PR #1136, opened 2025-04-03
Editor: strikethrough

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/948


packages/context-editor/src/schemas/strike.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export default {
+	attrs: {
+		id: { default: null },
+		class: { default: null },
+	},
+	parseDOM: [
+		{ tag: "s", getAttrs },
+		{ tag: "strike", getAttrs },
+		{ tag: "del", getAttrs },
```
kalilsn, line 18: according to mdn `<del>` is used to mark a deletion in a diff (i.e. track changes/suggested edits), which we might eventually support. so we may not want to coerce those to `<s>` elements, which would happen if you pasted content with a `<del>` element in. but this is probably fine for now!

kalilsn (approved):
good call on not including the markdown shortcut!

packages/context-editor/src/schemas/strike.ts
allisonking: oh that's good to know! I copied these from legacy but what you said makes sense. I'll keep it in mind for the future!

PR #1147, opened 2025-04-07
Refactor to use prosemirror-adapter-react for attribute menu

kalilsn:


packages/context-editor/src/ContextEditor.tsx
```diff
@@ -155,15 +170,19 @@ function UnwrappedEditor(props: ContextEditorProps) {
...
-			view.current.setProps({
-				editable: () => !props.disabled,
-			});
-			const tr = view.current.state.tr
-				.setMeta(reactPropsKey, { ...props, suggestData, setSuggestData })
-				.setMeta(attributePanelKey, { panelPosition, setPanelPosition });
-			view.current?.dispatch(tr);
-		}
+
+		setTimeout(() => {
```
kalilsn, line 174: adding this fixes a `flushSync` error. i don't really know why, and avoiding that kind of nonsense does seem like a strong pro of using the other view library

packages/context-editor/src/components/AttributePanel.tsx
```diff
@@ -1,110 +1,93 @@
...
+import { InlineLinkMenu } from "./InlineLinkMenu";
 
 const animationTimeMS = 150;
 const animationHeightMS = 100;
 
-export interface AttributePanelProps {
-	panelPosition: PanelProps;
-	viewRef: React.RefObject<EditorView | null>;
-}
+export function AttributePanel() {
```
kalilsn, line 17: i made smaller changes to this component than were probably necessary:
- got rid of its local state and used the parent's (maybe a mistake?). ideally it wouldn't depend on the `node` in the parent component state being up to date at all and could instead use prosemirror as the source of truth (perhaps just from the selection, or maybe directly in the plugin state). it should probably also submit all of its values at once instead of using these onchange handlers on individual inputs.
- moved it inline in the doc (also maybe a mistake)
- conditionally render the special link menu (not really functional, i wouldn't worry about it)

packages/context-editor/src/components/AttributePanelToggleWidget.tsx
```diff
@@ -0,0 +1,102 @@
...
+			return;
+		}
+		view.dispatch(
+			view.state.tr.setSelection(
+				new TextSelection(
+					view.state.doc.resolve(pos)
+					// view.state.doc.resolve(getPos() + node.nodeSize)
+				)
+			)
+		);
```
kalilsn, line 34: this was from an attempt to make the attribute panel work based on the selection, rather than manually setting the `node` in the panelProps. it's not really working at the moment so you'll probably want to remove this

packages/context-editor/src/components/AttributePanelToggleWidget.tsx
```diff
@@ -0,0 +1,102 @@
...
+		if (currentPubId === pubId) {
+			buttonText = `~${currentField.name}`;
+		} else {
+			buttonText = `/${currentTypeName}`;
+		}
+	} else {
+		buttonText = `${node.type.name}${node.type.name === "heading" ? ` ${node.attrs.level}` : ""}`;
+	}
+
+	return (
```
kalilsn, line 87: This is where I would make the buttons for lists/any nested blocks more readable. Perhaps we could group them by depth? Or make them show on some kind of inline hover when they're not at the top level of the document. The way we handle this in legacy for suggested edits may be helpful to look at, if that's something you're tackling here.

packages/context-editor/src/plugins/attributePanel.ts
```diff
@@ -1,19 +1,195 @@
...
+							prevEditorState,
+							editorState,
+							getToggleWidget
+						),
+					};
+				},
+			},
+			props: {
+				decorations(state) {
+					// return this.getState(state)?.decorations;
```
kalilsn, line 53: Uncommenting this should use our state logic to only rerender decorations in parts of the doc that change instead of crawling the whole doc each time, [as recommended by marijn](https://discuss.prosemirror.net/t/any-tips-for-caching-improving-performance-of-decorations/8044/1). But even without that, memoization makes this fast enough to work for now.

packages/context-editor/src/plugins/attributePanel.ts
```diff
@@ -1,19 +1,195 @@
...
+			}
+		});
+		return DecorationSet.create(editorState.doc, decorations);
+	}
+	let mappedDecorations = prevDecorationSet.map(tr.mapping, editorState.doc);
+
+	for (const stepMap of tr.mapping.maps) {
+		stepMap.forEach((oldStart, oldEnd, newStart, newEnd) => {
+			mappedDecorations = mappedDecorations.remove(mappedDecorations.find(newStart, newEnd));
+			editorState.doc.nodesBetween(newStart, newEnd, (node, pos, parent, index) => {
```
kalilsn, line 179: I haven't tested this, so if you do use it, I'd make sure this is using the right combo of old/new positions and the respected editorstate vs previouseditorstate. Also there's a chance these are just the totally wrong functions to use 😓

packages/context-editor/src/plugins/attributePanel.ts
```diff
@@ -1,19 +1,195 @@
...
+				// 	setPanelPosition({
+				// 		...panelPosition,
+				// 		isOpen: !panelPosition.isOpen,
+				// 		isLink,
+				// 		pos: range?.from || pos,
+				// 		node,
+				// 	});
+
+				// 	return true;
+				// },
```
kalilsn, line 96: I was using this in an attempt to make clicking into any mark open its menu, while also allowing the cursor to stay inside the editor, rather than having to click the tiny button and unfocusing the editor. But since the attribute panel depends on the node including the entire text in the mark and the pos pointing to the beginning of the mark, this needs some tweaking.

Issue #1121, opened 2025-03-27
Verify user emails after signup


tefkah:
## Motivation

Currently we don't verify users emails after/during signup at all, which is bad.
It wasn't really a problem before, as we invited users through their email already, so it wasn't that important.

But now with #1114 landing we for the first time will have public signups to pubpub. We need to check that users own the email they fill in there.

## Requirements
- After a user signs up, PubPub sends them an email with a randomly generated code to verify their email address
- When user has not yet verified, they are prompted to verify when trying to access any page while logged in with a page telling them we've sent them a verification email and they should check for it, and a button that allows them to re-send the verification, which creates a new token and invalidates the old one.
- After verification, they are forwarded to the page they were originally trying to access
- If the token has expired, user is shown a page telling them it's expired and a button that allows them to receive a new token.
- Forgot password also functions as verification, in case where user signs up, never completes verification, and deletes verification email

## Acceptance Criteria

- Remind @eastofesten: documentation update to Logging In
- When a user signs themselves up via /c/${community}/public/signup, update the publicSignup server action to Create an auth token of type verifyEmail
- Create an auth token of type verifyEmail that expires in 2 hours
- Send a verification email with a link that contains the token. Similar to sendForgotPasswordMail https://github.com/pubpub/platform/blob/main/core/lib/authentication/actions.ts#L132
   - If user visits the link after the token has expired, page should say the token has expired and contain a button labeled "Resend Verification Link" that, when clicked, repeats the creation of token and sending of verification link
- Add a new page /verify which checks the verification token. Similar to (user)/reset/page.tsx https://github.com/pubpub/platform/blob/main/core/app/(user)/reset/page.tsx
- Once verified, forward to the original page they were trying to access
- Add checks to other pages to make sure the user is verified
- Updates forgot password action to verify a user's email if it is not already verified (in case where user forgets verification and returns to signup months later)
- On other pages where a user is added via an email, `isVerified` is automatically set to true so those users do not need to go through the verification flow

allisonking:
To be a more granular/check my understanding of the auth setup:

1. When a user signs themselves up via `/c/${community}/public/signup`, update the `publicSignup` server action to
    1. Create an auth token of type `verifyEmail`    
        1. does this token expire?
    2. Send a verification email with a link that contains the token. Similar to `sendForgotPasswordMail` https://github.com/pubpub/platform/blob/main/core/lib/authentication/actions.ts#L132
[...]

PR #1151, opened 2025-04-08
Verify user emails

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/1121

Adds a verify email flow.

## Test Plan
I don't think this will work in the preview environment since we need to use emails.

Happy path:
1. In a browser that is not signed in, visit a public form i.e. http://localhost:3000/c/croccroc/public/forms/review/fill
2. You should see the public sign up page. Fill it out 
3. You should get to a page that says you need to verify your email
4. Open up inbucket at localhost:5432/monitor
5. You should have gotten an email with the subject 'Verify your email'. Click this link
6. You should be redirected to the public form, and a toast saying you've verified your email should appear at the bottom.

Other things to test:
* Request another verification email. This should invalidate the first one, and you should get a second email with a new token
  * You can try using the invalidated link and you will see a token invalid message, request another one
* After step 3 above, instead of verifying your email, try to navigate to another part of the app. You should always be sent to the `/verify` page, but the "Forgot password" flow is still accessible

https://github.com/user-attachments/assets/8a4c51e5-12ef-4d55-afd3-0993a709644f


core/app/(user)/magic-link/route.ts
```diff
@@ -16,30 +17,8 @@ const redirectToURL = (
...
-		});
-		return NextResponse.redirect(url, opts);
-	}
-
-	// invalid redirectTo, redirect to not-found
-	return NextResponse.redirect(
-		new URL(`/not-found?from=${encodeURIComponent(redirectTo)}`, env.PUBPUB_URL),
-		opts
-	);
+	const url = createRedirectUrl(redirectTo, opts?.searchParams);
```
allisonking, line 20: Moved this to a helper file since I found I needed it elsewhere (mostly to easily tack on search params). but I think @tefkah 's lil routes lib would be a lot nicer

core/app/(user)/magic-link/route.ts
```diff
@@ -129,6 +108,14 @@ export async function GET(req: NextRequest) {
 
 	const { user: tokenUser, authTokenType } = tokenSettled.value;
 
+	if (!tokenUser.isVerified) {
```
allisonking, line 111: by putting this here in the `magic-link` route, we automatically mark users who go through this route as verified since they would've come from their email. this lets us get the requirement "Forgot password also functions as verification, in case where user signs up, never completes verification, and deletes verification email" for free

core/app/(user)/verify/ResendVerificationButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+		const result = await sendVerifyEmail({ email, redirectTo });
+		if ("error" in result) {
+			setStatus("error");
+		} else {
+			setStatus("success");
+		}
+	};
+
+	return (
+		<SubmitButton
```
allisonking, line 31: very handy button!!

core/app/layout.tsx
```diff
@@ -23,7 +24,9 @@ export default function RootLayout({ children }: { children: React.ReactNode })
 					<NuqsAdapter>
 						<TooltipProvider>
 							{children}
-							<Toaster />
+							<Suspense>
+								<RootToaster />
+							</Suspense>
```
allisonking, line 29: hopefully this is okay—I wasn't sure how else to get the toast after redirect to work. it needs `Suspense` since `RootToaster` needs to `useSearchParams` in order to figure out whether to render the 'verified' toast

core/app/page.tsx
```diff
@@ -1,23 +1,36 @@
...
 	const lastVisited = cookieStore.get(LAST_VISITED_COOKIE);
 	const communitySlug = lastVisited?.value ?? user.memberships[0]?.community?.slug;
 
 	if (!communitySlug) {
-		redirect("/settings");
+		redirect(createRedirectUrl("/settings", params).toString());
 	}
 
-	redirect(`/c/${communitySlug}/stages`);
+	redirect(createRedirectUrl(`/c/${communitySlug}/stages`, params).toString());
```
allisonking, line 35: the regular pages weren't passing on search params. I needed it to so that `?verified=true` could get passed along to arbitrary pages

core/prisma/seed/seedCommunity.ts
```diff
@@ -95,6 +95,10 @@ export type UsersInitializer = Record<
 			isSuperAdmin?: boolean;
 			slug?: string;
 			existing?: false;
+			/**
+			 * @default true
```
tefkah, line 99: good default

core/lib/server/email.tsx
```diff
@@ -113,6 +111,36 @@ export function passwordReset(
...
 
+export function verifyEmail(
+	user: Pick<Users, "id" | "email" | "firstName" | "lastName">,
+	redirectTo?: string,
+	trx = db
+) {
+	return buildSend(async () => {
+		const magicLink = await createMagicLink(
+			{
+				type: AuthTokenType.generic,
```
tefkah, line 122: could you maybe add a small comment here why this needs to be `AuthTokenType.generic` vs `AuthTokenType.verifyEmail`? I think it's bc `/magic-link` will set their email to be verified, but i think that's easy to forget!

tefkah (approved):
i think it looks great! very elegant solution of making magic-links set the verified to `true`.

I made some optional suggestions which don't really need to be implemented if you don't feel like it, but there's one situation we do still need to cover.

if a user is invited through the email action, they are prompted to fill in their information+create a password. they can pick a different email during signup. i think in that case, we should force them to verify their email. this happens here:
https://github.com/pubpub/platform/blob/a54eb9b77fd202346153a294e3775c9dc9eb9bfb/core/lib/authentication/actions.ts#L493-L529

i think we'll get rid of this route as a special case soon, but we should still cover it for now!

greate work!

allisonking:
Thanks @tefkah , did not realize the legacy signup needed this too! added, and also implemented using nuqs, great idea!


https://github.com/user-attachments/assets/8b92e659-888c-4f7c-8274-6acf2036a243

PR #1154, opened 2025-04-10
fix: remove some warnings when running next dev

tefkah:
- **fix: fix annoying prettier warning**
- **fix: fix sentry external packages warning**
- **fix: don't double import the editor config**

## Issue(s) Resolved

When running the next dev server and sometimes on build, we would get the warnings:

- `package prettier/import-in-the-middle/require-in-the-middle can't be external`
- ``data-[state=open]:slide-in-from-top-[48%]` is ambiguous and matches multiple utilities.

The cause of the 1st one is that next has some default list of packages it does not bundle, like `prettier` and those `in-the-middle` packages, but some of our dependencies (`react-email` and `sentry` respectively) use those as deps. Next recommends we install those deps. But that then becomes a problem when the dependencies misalign.

The second one is caused by us importing the tailwind plugins twice, specifically the animation one. once through `ui`, and a second time through `context-editor`, which also imports `ui`. 

1. Override `react-email`'s prettier dependency to be the same as our one.
2. update `import/require-in-the-middle` so that sentry stops whinging
3. This one i'm the least certain on, i just removed the import of `context-editor`s tailwind config. It did not look like it caused any problems, but we should check on the preview.

## Test Plan
1. Go to preview env
2. Check if the context editor still behaves as expected.


allisonking (approved):
oh thank goodness that prettier warning was really getting to me yesterday!

I'm a little surprised the editor seems to work ok without the shared config thing. from what I remember it was there so that if we use a tailwind thing in the editor that isn't in the core app, that would still work. I remember this was a problem with I think the SuggestPanel but looking in the preview it looks fine to me 🤷‍♀️ actually maybe it only needed `packageSource("context-editor")` to make it work? (original commit: https://github.com/pubpub/platform/pull/732/commits/32e25116ba5118ccf71116c841c0a6afded98674)

Issue #1140, opened 2025-04-03
Don't allow http action runs, email sending, and image uploading on sandbox


gabestein:
## Motivation

So users don't use the sandbox to do spam or other nefarious things

## Requirements

- Does not allow http actions to be run (they should be able to be setup)
- Does not allow email actions to be run
- Does not allow member invites to be sent via email (either via email action or via member dashboard)
- Does not allow file uploads to occur.
- When disabled action is run, alert says "The action was triggered successfully, but will not run in the sandbox."

## Acceptance Criteria

- Use env variables to disable action runs, with an eye towards potentially making this programmable for self-hosters down the line
- Runs of disabled actions shouldn't add new row in action run log

eastofesten:
does this need to prevent file uploads generally, or just images in the editor?

3mcd:
File uploads generally, I think

PR #1158, opened 2025-04-14
Global feature flags via environment variables

3mcd:
## Issue(s) Resolved

Partial #1140

This PR introduces a simple pattern for implementing global (i.e. not community-level) feature flags.

One would create a new feature flag by adding a tuple to the new [`flagSchema`](https://github.com/pubpub/platform/pull/1158/files#diff-cee86778a1a51877651456193b7e1b5e7f257b8d066456b39544acfa0dcbe3afR30) zod type in `env.ts`, where the first element of the tuple is a unique flag name, and the second element is the flag "args" or options. Default flag states/args are defined using Zod's `.default` method.

Turning a flag on or off is as simple as starting the app with `FLAGS=flag-name:off`. You can toggle multiple flags using a comma separated list, e.g. `FLAGS=flag-1:off,flag-2:on`.

Some flags may take slightly more complex options than just "off" or "on". For example, the `disabled-actions` flag has a transformer that parses the following flag expression:

```env
disabled-actions:http+email
```

into

```json
["http", "email"]
```

The state and/or arguments of a flag can be retrieved on the server like
```ts
env.FLAGS.get("disabled-actions") // ["http", "email"]
env.FLAGS.get("uploads") // false
```

The flags I've introduced here don't do anything yet. I intend to utilize them in the `em/disabled-actions` branch, and a small follow-up branch to disable file uploads and member invites in our sandbox environment.

## Test Plan

1. Start the app with `FLAGS=uploads:off,invites:off`. It should start successfully! (These flags don't do anything yet).
2. Try running the app with an unsupported flag turned on or off. The app should fail to start with an environment error.
3. Try running the app with an unsupported flag state, e.g. `FLAGS=uploads:pancakes`. The app should fail to start again.

## Notes

Setting a feature flag currently doesn't have any effect, this PR just introduces the pattern.

### DIFF
--- a/core/lib/env/env.ts
+++ b/core/lib/env/env.ts
@@ -1,65 +1,13 @@
-// @ts-check
 import type { ZodTypeAny } from "zod";
 
 import { createEnv } from "@t3-oss/env-nextjs";
-import { z, ZodError } from "zod";
+import { z } from "zod";
 
-import { actionSchema } from "db/public";
+import { createFlags, flagsSchema } from "./flags";
 
 const selfHostedOptional = (schema: ZodTypeAny) => {
 	return process.env.SELF_HOSTED ? schema.optional() : schema;
 };
-
-const flagStateToBoolean = (flagState: string, ctx: z.RefinementCtx) => {
-	switch (flagState) {
-		case "on":
-		case "true":
-			return true;
-		case "off":
-		case "false":
-			return false;
-	}
-	ctx.addIssue({
-		code: z.ZodIssueCode.custom,
-		message: "Invalid flag state",
-		fatal: true,
-	});
-	return z.NEVER;
-};
-
-const flagSchema = z.union([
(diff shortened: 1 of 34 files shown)


core/lib/env/env.ts
3mcd, line 1: All of the important changes are in this file!

allisonking (approved):
worked as promised! although just a note that the check does not kick in on server reload. if I make my .env.local invalid, I'll get a message on the server that says "Reload env: .env.local" but the server will keep working fine. I think this is okay, but I guess means we can't dynamically swap out feature flags without restarting the server?

3mcd:
@allisonking I'm not totally sure what's causing the issue with reloading. Because we'll probably have a more robust solution in the future (i.e. one that supports per-community settings), I think I'm okay with the limitation of a server restart for global feature flags for now!

Issue #1132, opened 2025-04-02
Switch to react-prosemirror


gabestein:
## Motivation

So devs can more  easily manage prosemirror-react interactions

## Requirements

- Introduce [react-prosemirror](https://github.com/handlewithcarecollective/react-prosemirror) as our way of reconciling the two libraries' lifecycles
- Refactor all existing editor features to work with react-prosemirror
   - Plugins
   - Menus (toolbar, ~ menu, clickable class/settings menus)
   - Manual dom manipulation (e.g. structure decorations and carrots) as react components, not manual dom nodes
   - Recursive atom renderer

kalilsn:
https://github.com/pubpub/platform/pull/1147 here's my branch. It's a mess so let me know if you'd like to talk through it at all. I'm happy to explain my thought process/what issues I ran into. At the moment it's performant (as far as I know at least) and the main problem is just that the forms (the attribute menu and inline link menu) aren't working properly which can most likely be fixed by refactoring them so they submit all of their values instead of one input at a time.

PR #1159, opened 2025-04-15
Use react-prosemirror

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/1132

Integrates `react-prosemirror` to handle react + prosemirror interactions. 

## Test Plan

Try out the editor and make sure things still mostly work! https://pr-1159-react-prosemirro-ip-23-22-146-0.my.preview.run/c/croccroc/public/forms/review/fill

 The AttributePanel is buggy and I think the approach for finding an "active node" should be done differently, but I didn't want to spend a lot of time on it before we have settled on the actual pattern for how we want to open the panel, as that will affect how we determine the 'activeNode'

See comments throughout PR!


core/app/components/ContextEditor/AtomRenderer.tsx
```diff
@@ -1,21 +1,27 @@
-import { useNodeViewContext } from "@prosemirror-adapter/react";
```
allisonking, line 1: The `ContextAtom` here (and in `stories/AtomRenderer`) is a node view, and so must fulfill the node view props as described in the react-prosemirror docs https://github.com/handlewithcarecollective/react-prosemirror?tab=readme-ov-file#building-node-views-with-react.

This isn't too different from what we already had using `@prosemirror-adapter/react`—instead of passing a node to the component, we pass the full `nodeProps` which we basically get for free because we register this as a node view. Then, instead of being able to grab `selected`, we use the `react-prosemirror` hook `useIsNodeSelected()`

packages/context-editor/src/components/AttributePanel.tsx
```diff
@@ -1,221 +1,253 @@
...
+		activeNode,
+		position: activeNodePosition,
+		setActiveNode,
+		setPosition: setActiveNodePosition,
+	} = useEditorContext();
+
+	/**
+	 * This determination of the 'activeNode' is prone to bugs. We should figure
+	 * out a better way to do it.
+	 **/
```
allisonking, line 47: right now this mostly works, but there are edge cases such as if the user is typing **bold** and then they unbold, the attribute panel might stay open

https://github.com/user-attachments/assets/9b09be1a-e19b-41e1-b790-9b3a46b54aba

the most important thing is that if a paragraph has content, and we try to edit the paragraph's attributes, the panel closes automatically. this is because this bit of code thinks we are in the next node, which is no good! I will try to fix this soon

here it is working in other cases though

https://github.com/user-attachments/assets/5ca9c15e-3ffe-425d-891c-23bf6d672b4e

packages/context-editor/src/components/AttributePanel.tsx
```diff
@@ -1,221 +1,253 @@
...
+	if (Object.keys(nodeAttrs).length === 0 && nodeMarks.length === 0) {
+		return null;
+	}
+
 	// Marks will automatically show names, so it is only the 'inline' types
 	// that are not marks that need to be specifically rendered
-	const showName = node.type?.name === "math_inline";
+	const showName = activeNode?.type?.name === "math_inline";
 
 	return (
```
allisonking, line 151: below should mostly be the same as before

packages/context-editor/src/components/MenuBar.tsx
```diff
@@ -208,21 +210,30 @@ const paragraphTypeItems: MenuItem[] = [
 ];
 
 const ParagraphDropdown = () => {
-	const { view } = usePluginViewContext();
-	const activeType = paragraphTypeItems.find((item) => item.command(view)(view.state).isActive);
+	const itemCommand = useEditorEventCallback((view, item: MenuItem) => {
```
allisonking, line 213: rather than change the structure of the commands, I'm wrapping them in the `useEditorEventCallback` so that they have access to the `view`

packages/context-editor/src/components/StructureDecoration.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+import React, { forwardRef } from "react";
+import { useEditorState } from "@handlewithcare/react-prosemirror";
+
+import { reactPropsKey } from "../plugins/reactProps";
+import { useEditorContext } from "./Context";
+
+const getBlockName = (node: Node) => {
+	const state = useEditorState();
+	const { pubTypes, pubId, pubTypeId } = reactPropsKey.getState(state);
```
allisonking, line 12: we could likely put this in the context instead if we wanted

packages/context-editor/src/components/SuggestPanel.tsx
```diff
@@ -1,12 +1,36 @@
...
+	setSuggestData: any;
+};
+export default function SuggestPanel({ suggestData, setSuggestData }: Props) {
+	const { isOpen, selectedIndex, items, filter } = suggestData;
 	const [position, setPosition] = useState([0, 0]);
+
+	/**
+	 * In order to get the suggestions to the plugin, we pass props through
+	 * reactPropsKey which the plugin can then access.
+	 */
```
allisonking, line 19: this was the only part I couldn't get to be totally independent of passing state around plugins. I don't think `react-prosemirror` offers anything for passing state between plugins and react components, so I'm using the same `reactPropsKey` method we were using before. The `contextSuggestion` plugin uses `prosemirror-autocomplete` so I'm not very confident in tweaking it

packages/context-editor/src/plugins/structureDecorations.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export default () => {
+	return new Plugin({
+		props: {
+			decorations: (state) => {
+				const decorations: Decoration[] = [];
+				state.doc.descendants((node, pos) => {
+					if (node.type.isBlock) {
+						// TODO: is there a better key we can use?
+						decorations.push(widget(pos, BlockDecoration, { key: `node-${pos}` }));
```
allisonking, line 16: this file isn't all too different from the original `structureDecorations.ts`. It changes rendering the widget via React thanks the `widget()` from `react-prosemirror`. That is `StructureDecoration.tsx` instead of the JS `wrapWidget` function from before

packages/context-editor/src/stories/AtomRenderer.tsx
```diff
@@ -1,13 +1,16 @@
...
+import { useIsNodeSelected } from "@handlewithcare/react-prosemirror";
 import { CsvToHtmlTable } from "react-csv-to-table";
 
-export default function ContextAtom({ nodeProp }: { nodeProp: Node }) {
-	const { contentRef, node, selected } = useNodeViewContext();
+export const AtomRenderer = forwardRef<
+	HTMLDivElement,
+	NodeViewComponentProps & { selected?: boolean }
+>(function AtomRenderer({ nodeProps, selected, ...props }, ref) {
+	const { node } = nodeProps;
```
allisonking, line 11: See https://github.com/pubpub/platform/pull/1159/files#r2047460927 for explanation of this change.

packages/context-editor/src/components/AttributePanel.tsx
allisonking: there does seem to be a big with the updating mark logic that also seems to exist with the editor prior to this. I haven't been able to figure it out yet, but it seems like the mark attrs will sometimes apply to _all_ mark attributes. this does not appear to be the case if you already have a document loaded and are applying marks there though (i.e. it works ok in the ContextEditor > Primary story)

https://github.com/user-attachments/assets/4aa367db-cd74-40ac-b884-c77a6afb38a5

packages/context-editor/src/ContextEditor.tsx
```diff
@@ -82,103 +60,55 @@ const initSuggestProps: SuggestProps = {
...
-			const tr = view.current.state.tr
-				.setMeta(reactPropsKey, { ...props, suggestData, setSuggestData })
-				.setMeta(attributePanelKey, { panelPosition, setPanelPosition });
-			view.current?.dispatch(tr);
-		}
-		/* It's not clear to me that any of the props need to trigger this to re-render.  */
-		/* Doing so in some cases (onChange for the EditorDash) cause an infinite re-render loop */
-		/* Figure out what I actually need to render on, and then clean up any useMemo calls if necessary */
-	}, [props, suggestData, panelPosition]);
+		props.onChange(editorState);
```
allisonking, line 84: moved the `onChange` call to this `useEffect` rather than using the `onChange` plugin. this is possible now because our editor state is hoisted up.

I originally tried to put `onChange` inside of the `setEditorState` call on L95 but this resulted in a "bad set state" error and I'm not sure I totally know why

3mcd (approved):
I wrote up a whole thing that got lost when my wifi went out 😩 So I'll just say: looks good, although I have some small, unimportant questions about the `reactPropsKey` plugin.

Tested in the preview environment and the editor worked well!

PR #1162, opened 2025-04-15
chore: prune dangling docker images in preview environments

3mcd:
## Issue(s) Resolved

N/A

Amends the PullPreview pre-script to also clean up unused/dangling docker images on each deploy.

## Test Plan

Not sure... I think as long as the deployments no longer fail due to out-of-space errors we should be good to go!


3mcd:
Interestingly we only got `Total reclaimed space: 0B` on that change. Maybe due to something caching something layers something. Gonna try an application code change.

3mcd:
Got `Total reclaimed space: 3.187GB` that time. Nice!

kalilsn (approved):
looks great! i might add a `df -h` line to the script as well so we can see if there are other things that are slowly filling the disk, but i think pruning the images will probably fix the problem

PR #1166, opened 2025-04-16
chore: sandbox seeds

### LINKED ISSUE #1141: Write sandbox seeds
## Motivation

So users have some examples of relatively fleshed out communities to work with.

## Requirements

- Admin user account that is a member of two communities
- Seed 1 named "Legacy" that uses Thomas's existing Arcadia seed/model, but without any references to Arcadia, and with "content" fields as rich content types
- Seed 2 named "Starter" that uses the "CrocCroc" seed and contains just a simple pub with (at least) title and rich content field, and two stages, "Draft" and "Published"

## Acceptance Criteria

- Don't boil the ocean, feel free to copy existing seeds (as long as specific references to Arcadia and Crocodiles are stripped out)

3mcd:
## Issue(s) Resolved

Resolves #1141

This PR renames the CrocCroc seed to Starter, and the Arcadia seed to Legacy. The unjournal seed was removed.

Slugs and other string identifiers were updated to remove all references to Arcadia.

The big rule chaining example was removed from the Starter seed just to simplify the community for the sandbox. We can add it back later if necessary!

Finally, all of the `*:content` fields were converted from String->RichText.

## Test Plan

1. Run the seed, e.g. `pnpm --filter=core reset`. It should succeed.
2. Poke around the seed communities. Ensure the content fields are rich text!


gabestein (approved):
Looking good. Both took a look at the files and played with the preview.

One very nit-picky non-blocking thing: the legacy logo is weirdly skewed in the menu and should probably be changed to a square file.

PR #1171, opened 2025-04-18
RD: legacy migration

tefkah:
- **feat: create legacy structure separate from seed**
- **fix: fix some type errors**
- **feat: extremely basic importer**
- **fix: pass trx to getAllPubTypesForCommunity**
- **feat: delete existing stuff before reattempting migration**
- **feat: much better parsing, import entire archive**
- **refactor: make creating pubtypes with default forms easier**
- **refactor: improve passing through of trx**
- **refactor: create separate place for creating pubtype with form**
- **fix: worse validation**
- **feat: wow sort of importing works**
- **feat: add ability to target by value rather than id for many pub-op operations**
- **feat: add tests for by value operations**
- **fix: make relateByValue also properly accept multiple values**
- **fix: improve setting logic**
- **feat: import collections**
- **feat: import pages**
- **fix: update pnpm-lock.yaml for reasons**
- **fix: allow non-destructive update**
- **feat: allow you to actually set a size for a column in a datatable holymoly**
- **fix(pubOp): make updateByRelatedValue work**
- **fix(ui): fix checkbox alignment in tables**
- **feat(ui): add sticky header option for datatable**
- **feat(migration): add more granular undo**
- **feat(autoCache): correctly handle autocache in transactions**
- **refactor(pub-op): allow pub-op to not return things**
- **feat(migration): create top-level journal + connections**
- **feat(pubOp): tentatively add support for batch operations**
- **feat(pubOp): add tests for batch operations**
- **chore(pubOp): add seed**
- **feat(PubOp): make resolving non-target values more efficient**
- **fix(trx): do not always throw transaction**
- **feat(api): search pubtypes by name**
- **feat(api): search pubtypes by name array**
- **feat(site-builder): functioning site-builder**
- **feat(site-builder): add site building action**
- **refactor(ui): move submit button to ui package**
- **feat(site-builder): add astro configs for prettier and vscode**
- **feat(site-builder): add env vars for sitebuilder**
- **fix(pubOp): fix target resolution**
- **chore(migration): update server action to log out error**
- **feat(site-builder): add working action**
- **chore: new deps**
- **chore(migration): clean up migration code a bit**
- **fix(site-builder): do not depend on ui and use shared dockerfile**
- **chore(site-builder): keep builds dir**
- **chore(site-builder): keep more dirs**
- **fix(site-builder): run successfully in docker**
- **chore(site-buider): better docker folders**
- **feat(site-builder): use ts-rest with hono, use different thing with thing**
- **feat(site-builder): stream, and allow of uploading directory**
- **feat(site-builder): make building more programattic and slightly easier to paralellize**
- **fix(site-builder): include .env.development**
- **fix(site-builder): fix things**
- **feat(site-builder): add sitebuilder auth and sitebuilder specific tokens**
- **chore: fix env imports**
- **feat(site-builder): add infra configuration**


config/prettier/astro.js
tefkah, line 1: st we can `prettier` `.astro` files

core/actions/_lib/zodTypes.ts
```diff
@@ -53,6 +53,16 @@ class ActionInstance extends z.ZodObject<typeof actionInstanceShape, "strip", z.
...
 
+class FieldName extends z.ZodString {
+	static create = () =>
+		new FieldName({
+			typeName: "FieldName" as z.ZodFirstPartyTypeKind.ZodString,
+			checks: [],
+			coerce: false,
+		});
+}
+
```
tefkah, line 64: not _really_ used for the sitebuilder action, but could be used in the future.

core/actions/buildJournalSite/action.ts
```diff
@@ -0,0 +1,28 @@
...
+import { fieldName } from "../_lib/zodTypes";
+import { defineAction } from "../types";
+
+const schema = z.object({
+	token: z.string().describe("The token for the site builder"),
+	articles: fieldName(),
+	collections: fieldName(),
+	mainColor: fieldName(),
+	accentColor: fieldName(),
+});
```
tefkah, line 16: these are not actually used, but we should do so in the future (or find some other way to configure it)

core/actions/buildJournalSite/run.tsx
```diff
@@ -0,0 +1,122 @@
...
+				</p>
+			</div>
+		),
+		data: {
+			...data,
+			url: rewrittenDataUrl,
+			s3FolderUrl: rewrittenS3FolderUrl,
+		},
+	};
+});
```
tefkah, line 122: ill put this near the site builder service as well, but here's how it roughly works:

1. action gets site builder api token
2. action checks pings `api/health`
3. site builder responds "ok"
4. action initiates build, authenticates using `Bearer <site builder token>`
5. site builder pings `pubpub/.../site/health/check` with same token
6. action checks whether that token is indeed a site-builder token, says "yup"
7. site builder continues with build
8. ... build stuff
9. site builder finishes, upload artifact to url
10. action renders out toast saying: here's your site sire.

tada! 

you can also choose to have the site builder upload every individual page to eg `<s3Url>/site/.../pub/pub-1.html` etc. we cannot serve the site from there using minio, this would require some proxy service (nginx and caddy have (3rd party) s3 modules that could do that)

i feel like the authentication handshake is kinda solid: it relies on `core` knowing where the sitebuilder is, and sitebuilder to know how to authenticate. maybe it would be safer to introduce another token, but since we would need to send the sitebuildertoken back and forth anyway (since the sitebuilder needs to build sites for many communities) i thought this could work. please be hackerman and find some flaw in this!

tefkah: also note i allowed actions runs to be `.tsx` so we can just use `JSX` to template reports. not sure whether i like that yet, it's quite useful tho. leave a thumbs up/down whether you think its useful

core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts
```diff
@@ -271,6 +275,55 @@ const parseQueryWithQsMiddleware: RequestMiddleware = (req) => {
...
+					// 		return {
+					// 			status: 401,
+					// 			body: {
+					// 				ok: false,
+					// 				code: "HAS_NO_READ_PERMISSIONS",
+					// 				reason: `Site builder token has no read permissions for ${permission.scope}, which is required. Please contact support.`,
+					// 			},
+					// 		};
+					// 	}
+					// }
```
Comment (tefkah) [inline core/app/api/v0/c/[communitySlug]/site/[...ts-rest]/route.ts:318]: this was just extremely cautious but ended up causing problems, ill just remove it and we just trust that the sitebuilding token is good

core/app/c/[communitySlug]/settings/tokens/ExistingToken.tsx
```diff
@@ -15,7 +16,16 @@ export const ExistingToken = ({ token }: { token: SafeApiAccessToken }) => {
...
+					{token.isSiteBuilderToken ? (
+						<InfoButton>
+							<p>
+								This token is a site builder token. It has read-only access to all
+								content in the community. It cannot be revoked.
+							</p>
+						</InfoButton>
+					) : (
+						<RevokeTokenButton token={token} />
+					)}
```
Comment (tefkah) [inline core/app/c/[communitySlug]/settings/tokens/ExistingToken.tsx:28]: we should be able to regen it though, in case it gets leaked

- [ ] add ability to regenerate site builder token

core/lib/env/env.ts
```diff
@@ -50,6 +50,8 @@ export const env = createEnv({
 		DATACITE_REPOSITORY_ID: z.string().optional(),
 		DATACITE_PASSWORD: z.string().optional(),
 		SENTRY_AUTH_TOKEN: z.string().optional(),
+		SITE_BUILDER_ENDPOINT: selfHostedOptional(z.string().url()),
+		SITE_BUILDER_API_KEY: selfHostedOptional(z.string()),
```
tefkah, line 54: should be removed, we use the sitebuilder token instead

core/package.json
```diff
@@ -119,6 +119,7 @@
 		"nuqs": "catalog:",
 		"openapi3-ts": "^4.1.2",
 		"oslo": "^1.2.1",
+		"p-map": "^7.0.3",
```
tefkah, line 122: its just so dang handy!

basically `promise.all` but with concurrency

core/prisma/migrations/202504181210121_add_site_builder_token/migration.sql
```diff
@@ -0,0 +1,29 @@
...
+)
+INSERT INTO api_access_permissions ("apiAccessTokenId", scope, "accessType" )
+    SELECT new_access_tokens.id, 'community'::"ApiAccessScope", 'read'::"ApiAccessType" FROM new_access_tokens
+    UNION ALL
+    SELECT new_access_tokens.id, 'pub'::"ApiAccessScope", 'read'::"ApiAccessType" FROM new_access_tokens
+    UNION ALL
+    SELECT new_access_tokens.id, 'stage'::"ApiAccessScope", 'read'::"ApiAccessType" FROM new_access_tokens
+    UNION ALL
+    SELECT new_access_tokens.id, 'member'::"ApiAccessScope", 'read'::"ApiAccessType" FROM new_access_tokens;
+
```
tefkah, line 29: i wanted to create a sitebuilding token for each community, but this meant basically reimplemnting the api token creation in postgres. maybe we can just reuses this and have postgres always do it idk. 👎/👍 to vote

core/prisma/schema/schema.prisma
```diff
@@ -579,6 +580,9 @@ model ApiAccessToken {
   PubValueHistory PubValueHistory[]
   InviteHistory   InviteHistory[]
 
+  // has unique index with isSiteBuilderToken = true and communityId
+  isSiteBuilderToken Boolean @default(false)
+
```
tefkah, line 585: i know eric things rando boolean columns are cringe, but i couldnt think of anything better!

docker-compose.preview.yml
```diff
@@ -5,27 +5,30 @@ services:
...
-    #         ASSETS_UPLOAD_KEY: preview-different
-    #         ASSETS_UPLOAD_SECRET_KEY: preview-different123
+
+    minio-init:
+        restart: on-failure
+        environment:
+            MINIO_ROOT_USER: minioadmin
+            MINIO_ROOT_PASSWORD: minioadmin
+            ASSETS_UPLOAD_KEY: preview-different
+            ASSETS_UPLOAD_SECRET_KEY: preview-different123
```
tefkah, line 25: i think i managed to fix minio-init causing the preview to fail by not having it as the last thing in the chain, but idk it does work now, but browser based uploads will not work, as you can't really put `minio` on a `subpath` eg `https://..../assets`. S3s signing algorithm doesnt like it

infrastructure/terraform/modules/ecr-repositories/main.tf
```diff
@@ -39,6 +39,15 @@ resource "aws_ecr_repository" "pubpub_v7_jobs" {
...
 
+resource "aws_ecr_repository" "pubpub_v7_site_builder" {
+  name                 = "pubpub-v7-site-builder"
+  image_tag_mutability = "MUTABLE"
+
+  image_scanning_configuration {
+    scan_on_push = false # can set this to true if we want
+  }
+}
+
```
tefkah, line 50: one more repository please

packages/contracts/src/resources/site-builder.ts
```diff
@@ -0,0 +1,51 @@
...
+		path: "/health",
+		summary: "Health check",
+		description: "Health check",
+		responses: {
+			200: z.object({
+				status: z.literal("ok"),
+			}),
+		},
+	},
+});
```
tefkah, line 51: tiny little contract for the sitebuilder so we can easily use ts-rest for actions as well

self-host/docker-compose.yml
```diff
@@ -118,23 +120,17 @@ services:
...
+        depends_on:
+            minio:
+                condition: service_healthy
+        image: minio/mc:latest
+        env_file: .env
+        entrypoint: ["/bin/sh", "/minio-init.sh"]
+        volumes:
+            - ./minio-init.sh:/minio-init.sh
+        networks:
+            - app-network
```
tefkah, line 133: i cant remember which part of this made it work

site-builder/astro.config.mjs
```diff
@@ -0,0 +1,39 @@
...
+		// 	maxConcurrent: 10,
+
+		// 	pages: (pathname) => {
+		// 		if (pathname.includes("/pub") && !pathname.includes("/release")) {
+		// 			return true;
+		// 		}
+		// 		return false;
+		// 	},
+		// }),
+	],
```
tefkah, line 37: i think well have a different solution for pdfs right?

site-builder/docker-compose.yml
tefkah, line 1: for locally testing whether it works as a docker container

tefkah: out of date, as it's pointing to a no longer existing Dockerfile

site-builder/server/astro.ts
```diff
@@ -0,0 +1,18 @@
...
+export const buildAstroSite = async (astroConfig?: Partial<AstroInlineConfig>) => {
+	try {
+		const config = { ...defaultConfig, ...astroConfig } satisfies AstroInlineConfig;
+		await build(config);
+		return true;
+	} catch (error) {
+		console.error(error);
+		return false;
+	}
+};
```
tefkah, line 18: super nice you can run astro programitcally like this!

tefkah: im not fixing that spelling

site-builder/server/server.ts
```diff
@@ -0,0 +1,586 @@
...
+import { PassThrough } from "stream";
+
+import { S3Client } from "@aws-sdk/client-s3";
+import { Upload } from "@aws-sdk/lib-storage";
+import { serve } from "@hono/node-server";
+import { initClient } from "@ts-rest/core";
+import { fetchRequestHandler, tsr } from "@ts-rest/serverless/fetch";
+import archiver from "archiver";
+import { Hono } from "hono";
+import mime from "mime-types";
```
tefkah, line 15: i chose to use `hono` as it seems like what the cool kids are using, but it doesn't work thaaat nicely with `ts-rest`.

tefkah: requires some manual setup, that's it

site-builder/server/server.ts
```diff
@@ -0,0 +1,586 @@
...
+
+import { siteApi } from "contracts";
+import { siteBuilderApi } from "contracts/resources/site-builder";
+
+import { buildAstroSite } from "./astro";
+
+const env = await import("../src/lib/env/server").then((m) => m.SERVER_ENV);
```
tefkah, line 22: i did a dynamic import for a reason, but i can't recall it

site-builder/server/server.ts
```diff
@@ -0,0 +1,586 @@
...
+		results.push(...batchResults);
+	}
+
+	console.log(
+		`Directory upload complete. Uploaded ${results.length} files to s3://${bucket}/${prefix}`
+	);
+
+	// Return the base URL of the uploaded content
+	return `${env.S3_ENDPOINT}/${bucket}/${prefix}`;
+};
```
tefkah, line 238: the idea was: what if you could just upload all the files to `bucket/prefix` and now you have a site! `minio` dont work like that, and our s3 also doesnt, so maybe remove this in the future

site-builder/server/server.ts
```diff
@@ -0,0 +1,586 @@
...
+				reject(err);
+			});
+
+		// Append files from the directory to the archive
+		archive.directory(sourceDir, false);
+
+		// Finalize the archive - this is when data actually starts flowing
+		archive.finalize();
+	});
+};
```
tefkah, line 393: yum streams. im glad i can let robots write this code for me now, i really hate working with this

site-builder/server/server.ts
```diff
@@ -0,0 +1,586 @@
...
+app.all("*", async (c) => {
+	return fetchRequestHandler({
+		request: new Request(c.req.url, c.req.raw),
+		contract: siteBuilderApi,
+		router,
+		options: {
+			//
+		},
+	});
+});
```
tefkah, line 579: this is how you use ts-rest with `hono`, although i think this instantiates the entire route structure on every request, which is obviously a bit slow.

site-builder/server/server.ts
```diff
@@ -0,0 +1,586 @@
...
+					body: { success: false, message: "Build failed" },
+				};
+			}
+
+			console.log("Build successful");
+
+			let uploadResult: string | undefined;
+			let s3FolderUrl: string | undefined;
+			let s3FolderPath: string | undefined;
+			let error: Error | undefined;
```
tefkah, line 477: i tried to make the builds slightly parallelizable across communities, and i think it works? i think `buildAstroSite` should spin up its own process and use diff env vars across them, but i have yet to test it. it builds the stuff inside of different directories at least, but i reuses the same directory across different builds for the same community. that could probably be better

site-builder/server/server.ts
```diff
@@ -0,0 +1,586 @@
...
+				return {
+					status: 200,
+					body: {
+						success: true,
+						message: "Site built and uploaded successfully",
+						url: uploadResult,
+						timestamp: timestamp,
+						// We can't determine the exact file size without saving to disk
+						// or collecting that data during archiving/upload, so we provide estimated values
+						fileSize: 0, // Required by the API contract, but we don't know the exact size
```
tefkah, line 520: i think we can likely get the filesize from s3, i was just lazy

site-builder/src/lib/client/queries.ts
```diff
@@ -0,0 +1,125 @@
...
+			withRelatedPubs: opts?.withRelatedPubs ?? true,
+		},
+	});
+
+	if (journal.status !== 200) {
+		throw new Error("Failed to fetch journal");
+	}
+
+	return expect(journal.body?.[0]);
+};
```
tefkah, line 48: so obviously we shouldnt do it like this, but pass in the id of the journal or something in the action (it is the pub that the action is run on).
none of this should be treated as how we should _actually_ do it

site-builder/src/pages/[slug].astro
```diff
@@ -0,0 +1,93 @@
...
+				params: {
+					slug,
+				},
+				props: { page, journal, header },
+			};
+		})
+		.filter((page) => !!page);
+
+	return paths;
+}
```
Comment (tefkah) [inline site-builder/src/pages/[slug].astro:53]: i really like this about astro, we only need to fetch that data once and then can use it for each page!

however, we might need to do it differently: if you have a sufficiently large community this requires keeping every pub in memory (i think) in order to build the entire site, which is will obviously take it down very quickly.

instead we should probably write the results to disk or something (stream it ideally) and then read it for each page. thisll be slower but easier to scale.

ive tested the same approach with `next`, which is much slower for doing this (20sec builds vs 2s on astro)

tefkah (commented):
hopefully github will nest all these reviews:

these comments are part of the site builder. ill put the comments on the migration in a separate review

tefkah (commented):
here are all the comments related to the migration!

3mcd (approved):

PR #1175, opened 2025-04-22
feat: dead simple realtime update framework

tefkah:
## Issue(s) Resolved

-

Adds very simple notifcation framework, which works as follows:

1. on an interesting table, add a new trigger which calls `notify_change` somehow. you can use `notify_change_generic` for tables with a `communityId`, but you need to write a small wrapper around `notify_change` to retrieve a communityId for tables which do not have it (could also easily be solved by giving everything a communityId)
2. add `<SSERevalidator>` element with the name of an interesting table on a page where you want to update the contents of the page when something changes
3. this establishes an `SSE` connection to `/api/v1/c/[communitySlug]/sse`, which initializes a db client, which listens to the channels `change_${communityId}_${table}`.
4. when a table is updated, postgres calls `notify_change` , which will notify database clients listening to `change_${communityId}_${table}`
5. `/sse` sends a payload to with the changes to the client
6. the client simply calls `router.refresh` on any data (debounced a bit, and checks whether the data is actually different)

that's it!

You can also use the `useSSEUpdates` manually and manually do things when the data changes, but you probably don't want to do both that and use `<SSERevalidator/>`

### Necessary improvements

- Provide some better way of filtering what data the client can receive from `/sse`. currently they receive ALL updates from the table, but that can leak sensitive data. 
- What tables to listen to should also not only be determined by the client sending arbitrary query params, we should do some authorization checks here in the future
- 

## Test Plan

1. Open two tabs, go to the same stage with the same action
2. in one, open the "Pubs" tab, on the other the "Actions" panel
3. run any action on a pub
4. a little icon should appear next to the action in tab 2, which if you hover over should provide some info about how the action was run
5. run action again
6. the icon in tab 2 should gain a little blue circle, indicating that new data is available.


3mcd:
@tefkah This is really exciting. Nice work!

Could you provide a deeper explanation that explains why the revalidations are excessive with this implementation, and an example of what the API would look like for a page to "subscribe" to table notifications?

I think because I am more used to stateful, bidirectional RTC, I'm having a hard time wrapping my head around why you use the database as an event source rather than the application.

Couldn't the server emit the SSE events each time `runActionInstance` succeeds (or fails)? The action log table is currently a mostly passive part of the app, but with this approach, it's now tied closely to application behavior/logic.

tefkah:
@3mcd thank you! 

good questions, i think most of the reasons are more psychological than things im really committed to, so here's some mix of reasons why use the db rather than the server

- we were already using the db as an event source (pub enter stage rule, the graphile worker stuff), so i didn't think this was particularly weird. i would just add this trigger to all tables we think are interesting (and probably change `for each row` to `for each statement`)
- the idea that the server could trigger these events just straight up didnt occur to me lol, bc im used to thinking of next api routes as serverless and stateless. ofc  that seems kinda silly now bc this SSE stream is long-lived to begin with and therefore clearly not serverless/ephemeral, but i guess i was still approaching it from that paradigm, ie there's no real inter-server communication
- that said, i do still think this approach is simpler than notifying the api route from the server, as does not introduce a new kind of data flow. everything is still `db -> server -> client` and `client -> server -> db`. with the approach you outlined, we would have a newish kind of event on the server, which i dont think we really have yet (maybe we do and i overlooked something). 

As for the API of a page listening to specific table updates: I already implemented that! see 

https://github.com/pubpub/platform/blob/cc718398d536baa8210607a5f93e5c5313306028/core/app/c/%5BcommunitySlug%5D/stages/manage/page.tsx#L104-L106

~this `listen` prop then gets used here, just a very simple check on the client to see whether to do anything on receiving an event. I suspect it might also be possible to specify on the server~

https://github.com/pubpub/platform/blob/cc718398d536baa8210607a5f93e5c5313306028/core/lib/notify/useSSEWithRevalidation.ts#L41-L66

after thinking about it, it makes more sense for the server to filter out the notifications the client gets.

i now set up a connection with `/sse` like `/sse?listen=action_runs` (you can pass multiple)

https://github.com/pubpub/platform/blob/0d1598ce7f607550909069eaae965fe6c69b41a5/core/lib/notify/useSSEWithRevalidation.ts#L32-L45

and then filter those out on the server

https://github.com/pubpub/platform/blob/0d1598ce7f607550909069eaae965fe6c69b41a5/core/app/api/v0/c/%5BcommunitySlug%5D/sse/route.ts#L40-L72

that way the client doesn't need to potentially handle a ton of events.

tefkah:
Re your comment

> The action log table is currently a mostly passive part of the app, but with this approach, it's now tied closely to application behavior/logic.

i don't think this is such a big problem, and i wouldn't say it's "closely tied": the app still works without the `action_runs` table, we basically just provide an "auto refresh" when its data changes, which is kind of what we want. it doesn't really influence the core logic of the "app" i'd say bc we still have this unidirectional data flow from the db->server->client, just with an occasional auto refresh. we could disable the db trigger at any time and the app would continue to function as it did.

maybe said differently: i think treating the changing of the action_runs table as significant in the app (rather than being "passive" like  you said) is fine, bc (in this example) it's exactly what we're interested in. we want to see the changes in the action_run state (again in this example), so it makes sense to listen to them.

with this framework, that is what we do: we just listen to changes in tables, and refresh the data if needed. now, maybe this is not always what we want, but i think it's simpler conceptually then adding a bunch of custom hooks in different places. eg what if we want to show changes in stages, memberships, pub_values etc, things that are changed not through `runActionInstance`? now we need to remember where we put all those hooks, which from my exp doing that in v6 with cache invalidation is kinda annoying. 

but this does really depend on how we plan to expand these updates!

3mcd:
I think as long as we keeep it simple, like action run statuses, notifications, that sort of thing... then this pattern will scale fine. I am hesitant about making the client aware of specific tables rather than "topics" like you might subscribe to in a message queue, socket.io server, etc.

What isn't clear to me is how we might implement an SSE for some kind of event that happens as a consequence of more than one table being updated. Or an event that would pass through some ephemeral data (like a pointer position-- obviously we'd use WebSockets for that kind of thing but just an example), without storing it in a table. But I don't think this is a real concern right now because our domain entities are pretty much 1:1, and our concepts are fairly encapsulated/lightweight.

Issue #1169, opened 2025-04-17
Implement editor links custom attributes menu


gabestein:
## Motivation

So users can add links to rich context intuitively.
So developers have a template on which to build future context menus.

## Requirements

Replaces existing attributes menu with a custom one for configuring link options, including:
- Configure the link destination
- Configure if link should be opened in a new tab/page
- Remove link once set without needing to highlight entire link
- Visit link
- Advanced options for setting id and class

## Acceptance Criteria

- Updates the general menu behavior for all nodes/marks such that:    
   - Menu opens when user clicks on the node/mark
   - Menu DOES NOT open when user merely hovers or curses over the node/mark
   - Menu should close when the user focuses on another element on the page
- Specific to the link mark:
    - Menu should open on creation of the link (via button click or cmd/ctrl+k), with the URL input element in the menu focused such that a user can directly paste in a copied link -- this is a special case for link marks
- Adds tests for link functionality added in #944 (in addition to any new required tests)

## Mockups (if available)
https://www.figma.com/design/SqOTMElJn7fuPljfQL1gIY/PubPub-Platform?node-id=2388-5355&t=hEdQ8HQc3TED0Bah-1

![Image](https://github.com/user-attachments/assets/8e603e1f-e385-4d6f-9b0a-f784e419dcb9)

gabestein:
> Are these behaviors meant to apply to all marks/nodes? i.e. how much of this is unique to the link menu vs for all marks/nodes that open up the attribute panel?

Updated the ticket in response. In general, let's use the Legacy heuristic:

- Menu opens when you click on the mark/node OR click on the menu item.
- Menu does not open just when cursing/hovering over the mark/node.

The link element should have one special case, though, which is that the menu should open when you first create the mark, and the URL input should get focus, so that you can paste a link directly in.

> Where should the menu pop up? Currently it always off to the right margin but I think there was talk about putting it next to the cursor or elsewhere?

Let's keep it right margin for now. We may want to special-case the link node in the future to be below/to the right of the cursor, a la Legacy and other text editor link features, but for now I think it's okay not to as we do a little more work on design philosophy.

allisonking:
Thanks Gabe!

> Menu opens when you click on the mark/node OR click on the menu item.

[...]

PR #1177, opened 2025-04-22
Link menu

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/1169

Implements the link editor. Also makes improvements to the attribute panel so it opens/closes in a more intuitive way.

## Test Plan

Try playing with the link menu here! https://pr-1177-link-menu-ip-54-83-75-247.my.preview.run/c/starter/public/forms/review/fill 

https://github.com/user-attachments/assets/14052e85-6a9b-4b9f-b239-bdc340f9c287

## Notes

Some of the tests seem to fail in firefox/webkit :( will look into it more tomorrow!


packages/context-editor/src/components/AttributePanel.tsx
```diff
@@ -41,19 +42,22 @@ export function AttributePanel({ menuHidden }: { menuHidden: boolean }) {
 		setPosition: setActiveNodePosition,
 	} = useEditorContext();
 
-	/**
-	 * This determination of the 'activeNode' is prone to bugs. We should figure
-	 * out a better way to do it.
-	 **/
```
allisonking, line 47: it's still a bit more bespoke than I'd like, but at least there are tests for it now so should hold up okay 🤞

packages/context-editor/src/components/AttributePanel.tsx
```diff
@@ -114,6 +118,35 @@ export function AttributePanel({ menuHidden }: { menuHidden: boolean }) {
 		}
 	);
 
+	/** Bulk update version of updateMarkAttr */
+	const updateMarkAttrs = useEditorEventCallback(
```
allisonking, line 122: At first I had the menu calling the old `updateMarkAttr` which updates one attr at a time (the same way the original attribute panel did it). However, I realized that now that we have proper forms, we also have proper validation, and so ideally instead of propagating the data directly to the prosemirror editor state, we can validate it first.

for example, the Link popover has a URL field, so we can use our URI validator to make sure it's a proper URL. Then, only once it's valid, do we propagate it to the editor state.

I made the form submit `onBlur` since it's fewer calls than `onChange`, but we could always go back to `onChange` if that's preferred

packages/context-editor/src/ContextEditor.tsx
```diff
@@ -98,8 +102,12 @@ export default function ContextEditor(props: ContextEditorProps) {
 						</div>
 					)}
 					<ProseMirrorDoc />
-					<AttributePanel menuHidden={!!props.hideMenu} />
-					<SuggestPanel suggestData={suggestData} setSuggestData={setSuggestData} />
+					<AttributePanel menuHidden={!!props.hideMenu} containerId={containerId} />
```
3mcd, line 105: Instead of passing a container ID to the `AttributePanel`, could you instead pass a ref to the container div as a prop? Better yet, what about computing the values you need in the `ContextEditor` (like the `offsetTop`/`clientWidth`) and passing those as scalar props?

packages/context-editor/src/components/AttributePanel.tsx
```diff
@@ -284,6 +344,7 @@ export function AttributePanel({ menuHidden }: { menuHidden: boolean }) {
 					transition: position.top === 0 ? "" : `right ${animationTimeMS}ms linear`,
 				}}
 			/>
-		</>
+		</>,
+		container
```
3mcd, line 348: I see now that the reference to the container DOM node is necessary here for the portal. Could you help me understand why a portal is used here?

packages/context-editor/src/components/menus/LinkMenu.tsx
```diff
@@ -0,0 +1,142 @@
...
+	});
+
+	const handleSubmit = (values: FormSchema) => {
+		const { openInNewTab, ...rest } = values;
+		const attrs = { ...rest, target: values.openInNewTab ? "_blank" : null };
+		onChange(attrs);
+	};
+
+	return (
+		<>
```
3mcd, line 73: I think this is an unnecessary React fragment.

3mcd (changes requested):
Just a couple minor questions!

packages/context-editor/src/components/AttributePanel.tsx
3mcd: Ah! Just saw your comment about the nested form issue. Such a frustrating thing. I guess we don't really need to use proper `<form>` elements anywhere since our submit handlers are generally invoking JS code on submit (i.e. don't use the native `method` or `action` attributes for the forms). But I guess it's a good thing to use form elements semantically?

Anyways, the portal solution seems like a decent one, I'd just prefer that we pass a ref instead of querying for the element by ID.

3mcd (approved):

PR #1178, opened 2025-04-23
feat: add batch operations to PubOp, expand/fix *byValue capabilities

tefkah:
## Issue(s) Resolved

-

The goal is to split off some work from the migrations pr to make it more reviewable in the long term.

This PR does two things that are necessary for #1171 

1. Allow batch updating/creating/upserting of pubs. that way in order to create N pubs, rather than needing `N x M` calls to the db, where `M` is the number of calls pubop makes for creating one pub, we now only need `M` calls. Much more efficient.
2. Fix and expand the `byValue` pattern established by `PubOp.upsertByValue` and `PubOp.updateByValue` to actually work, and also be available for `relateByValue`. This allows you to do `pubOp`s of which you may only know a specific value (in the case of the migration the `legacy-id`), rather than the id. This is very useful for doing imports.

## Test Plan

- Convince yourself of the tests
- See for instance the `createCollections` fn in the migrations pr on how both are to be used

https://github.com/pubpub/platform/blob/187f0148bd7c220b275a32164ccb1fcdd8b7eb11/core/lib/server/legacy-migration/legacy-migration.ts#L875-L1067

Explanatory video

https://github.com/user-attachments/assets/1d0febec-3fa5-44b8-b737-8e3fb26f24fb

### DIFF
--- a/core/lib/server/pub-op.ts
+++ b/core/lib/server/pub-op.ts
@@ -543,15 +543,19 @@ abstract class PubOpBase {
 
 			const fallbackOnNotFound = operations.get(target)?.mode === "upsert";
 
-			if (pubIds.length === 0 && !fallbackOnNotFound) {
-				throw new PubOpError(
-					"INVALID_TARGET",
-					`No pub found for target: ${target.slug} = ${target.value}`
-				);
-			}
-
 			if (pubIds.length === 1) {
 				targetMap.set(target, pubIds[0]);
+			}
+
+			if (pubIds.length === 0) {
+				if (!fallbackOnNotFound) {
+					throw new PubOpError(
+						"INVALID_TARGET",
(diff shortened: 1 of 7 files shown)


core/lib/server/pub-op.ts
```diff
@@ -114,6 +115,11 @@ type SetOptions = {
 	 * ```
 	 */
 	deleteExistingValues?: boolean;
+
+	/**
+	 * If true, nullish values will not be set.
+	 */
+	ignoreNullish?: boolean;
```
3mcd, line 122: So this flag will turn a `.set(slug, null)` into a no-op? Why delegate that responsibility to `PubOp` instead of the caller?

3mcd (approved):
Tests look great. The changes here are pretty noisy and difficult to follow, but I did my best to look over them.

PR #1180, opened 2025-04-23
ci: run context-editor playwright test later and make fields test more consistent

tefkah:
- **fix(ci): run playwright at e2e**
- **fix(e2e): install correct playwright deps**

## Issue(s) Resolved
The context-editor e2e tests often take longer than building core, with it taking the longest of all the CI + build steps, holding off deploying the preview environments and the core e2e tests.

The context-editor e2e tests now run at the same time as the core ones, st it wont hold up deploying previews.

Also they are now cached by turbo, so can sometimes be skipped.

## Test Plan

- see ci time go down a lil bit


allisonking (approved):
nice! lgtm though it looks like the upload isn't quite working yet

Issue #957, opened 2025-02-10
Add support for uploaded media settings


gabestein:
## Motivation

So users can add uploaded media (images, videos, etc.) to their pubs. We should figure out what the state of the art strategy is.

## Requirements

- Upload media -- should support:
   - Images (jpg, png, svg, webp, gif...any other common formats?)
   - audio (rendered via native <audio> container with mp3, wav, ogg, mp4, flac), video (via native <video> container with mp4, ogg, and webM), and "files", which is anything else, and can be rendered as a card that links to the file
   - "files" -- anything else, rendered as simply a card that links to the file. If the upload ends up being a file, the "link to" option should become disabled.
- Specify alignment
- Specify width/height
- Specify alt text when applicable
- Specify a caption that can be edited in the document - rich text with all inline properties and ability to reference other Pubs, but no alignment, block properties, media/attachment, or "transclusion"
- Specify a source that can be edited in the document — text with bold, italic, links  (if simpler, can just support all features for now rather than needing a "minimal" editor)
- Specify a license that can be edited in the document — text with bold, italic, links (if simpler, can just support all features for now rather than needing a "minimal" editor)
- Specify other accessibility/alternative options, e.g. captions or fallback images for audio/video

## Acceptance Criteria

- Will need to solve problem of how to conditionally enable editable nodes for caption, source, and license
- Caption, source, and license should be editable in the document once enabled
- Likely uses Uppy for uploading and light image editing
- Media should upload to our S3 bucket, prepended by the Pub ID
- Users should be able to drag files into their docs and create media
- Users should be able to paste in media and have it upload as usual
- Add @eastofesten as reviewer for documentation

## Mockups (if available)
https://www.figma.com/design/SqOTMElJn7fuPljfQL1gIY/PubPub-Platform?node-id=2348-81470&t=Fh6donJBbzs2vAW5-1

![Image](https://github.com/user-attachments/assets/e3d94726-d1df-4224-9d73-959845a7464c)
![Image](https://github.com/user-attachments/assets/a4d8b876-c108-4f12-9356-5d435ce73f4b)
![Image](https://github.com/user-attachments/assets/365aa2f0-a560-4348-bda2-3c533d6db586)
![Image](https://github.com/user-attachments/assets/666f2a39-498d-4320-836e-d8f9790ba21d)

allisonking:
A good chunk of the upload work was done in https://github.com/pubpub/platform/issues/983

[...]

PR #1187, opened 2025-04-26
Uploaded media settings

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/957 but there are a few things we decided should be done in follow-up PRs/tickets:

* Other media types (audio, video)
* Drag and drop into the editor to upload
* Paste into the editor to upload

Adds the upload menu with toggles that allow for adding/removing subnodes of a figure.

## Test Plan

Since we don't have upload in the preview environment, run the context-editor in storybook (in `packages/context-editor`, run `pnpm run storybook`), or run the app on a form that has a rich text editor.

Try uploading media and setting the attributes, including toggling caption, license, credit, and writing text to it.

https://github.com/user-attachments/assets/77c59df7-86f5-4635-ab6a-54bb5bbc0a33

## Notes

As you can see from the video, there is some funkiness in our decorations when you edit a node on top of another node (i.e. if you try to write a caption above a license). I believe this is a bug with the library we are using, and have filed an issue https://github.com/handlewithcarecollective/react-prosemirror/issues/51 

Also the `parseDOM` isn't totally working right now, but we aren't using it anywhere yet so can always come back to it, perhaps using `getContent` instead of relying on the content expression


packages/context-editor/src/stories/ContextEditor.stories.tsx
```diff
@@ -71,3 +76,64 @@ export const Blank: Story = {
...
+				<ContextEditor {...args} onChange={setState} />
+				<pre className="text-xs" data-testid="prosemirror-state">
+					{state ? JSON.stringify(state?.doc.toJSON(), null, 2) : "{}"}
+				</pre>
+			</>
+		);
+	},
+};
+
+export const ParseDOM: Story = {
```
allisonking, line 107: for trying out `parseDOM`.... the `<figure>` elements don't parse the way I want them to yet

packages/context-editor/src/components/menus/LinkMenu.tsx
```diff
@@ -113,27 +110,9 @@ export const LinkMenu = ({ mark, onChange }: LinkMenuProps) => {
...
-									onCheckedChange={(checked) => {
-										field.onChange(checked);
-									}}
-								/>
-							</FormControl>
-							<FormMessage />
-						</FormItem>
-					)}
-				/>
+				<MenuSwitchField name="openInNewTab" label="Open in new tab" />
```
3mcd, line 113: nice!

3mcd (approved):
Nice work!

Like we discussed I may replace your `toggleNodesInFigure ` function with my version from the tables branch. I'll remove those attributes too in favor of looking at the document state to determine the state of the toggle switches.

packages/context-editor/src/stories/ContextEditor.stories.tsx
3mcd: that's ok! I'll take a look at `parseDOM` in the table branch because I'll need to solve it for tables, too. I'll probably end up using `getContent`, but we'll see.

Issue #1108, opened 2025-03-24
Get basic prosemirror tables working in the editor


gabestein:
## Motivation

So users can add a basic table

## Requirements

- Button that adds a 2x2 table that a user can edit (and any other "default" table features)
- Ability to enable title and caption via context menu, same as #957
- Ability to edit title and caption in editor nodes, same as #957

## Acceptance Criteria

- Uses https://github.com/ProseMirror/prosemirror-tables or https://www.remirror.io/docs/extensions/react-tables-extension
- Add @eastofesten as reviewer for documentation

PR #1190, opened 2025-04-28
feat: simple editor tables

3mcd:
## Issue(s) Resolved

Partial #1108, will add a follow-up PR for title and captions when #1177 lands.

TODO: tests

- Adds the prosemirror-tables table nodes to our schema.
- Adds the prosemirror-tables `tableEditing` plugin to our plugin list.
- Adds a new editor MenuBar button that will insert a fixed 3x3 table into the document immediately after the cursor position.
- Adds some basic styling for tables.
- Adds some basic Tab-key navigation for tables.

I also did some (fairly tame) refactoring of the way we track the active node that hopefully is a bit easier to understand and a little more fault-tolerant.

## Test Plan

- Add a table to the editor [while editing any pub](https://pr-1190-pm-tables-ip-18-212-226-214.my.preview.run/c/starter/pubs/c27107ee-ffa3-43e8-84d5-23997a3bf914/edit) and experiment with inserting content in the fixed 3x3 cells.
- Toggle the caption and title via the table's parent figure's attribute panel and try to add rich content to either.
- Insert media somewhere (ideally outside of the table) and toggle the image's parent figure's title, caption, license, and credit nodes.

A table!

Screenshot 2025-04-29 at 11 53 30 AM

## Notes

- Lists look bad inside of the tables right now. Especially because I'm hiding the borders that would indicate _why_ the spacing is so wonky. 
- prosemirror-table's `columnResizing` plugin currently breaks the editor, so it's disabled for now.
- There are still several outstanding bugs, including one RangeError seemingly to do with document positions. I'll try to address them before merging the PR.


packages/context-editor/package.json
```diff
@@ -105,8 +105,9 @@
 		"prosemirror-schema-list": "^1.5.1",
 		"prosemirror-state": "^1.4.3",
 		"prosemirror-suggest": "^3.0.0",
+		"prosemirror-tables": "^1.7.1",
 		"prosemirror-transform": "^1.10.0",
-		"prosemirror-view": "1.37.1",
+		"prosemirror-view": "1.39.2",
```
allisonking, line 110: I think we actually need this at `1.37.1` per https://github.com/handlewithcarecollective/react-prosemirror?tab=readme-ov-file#installation 

> Note: React ProseMirror releases are coupled to specific prosemirror-view releases, and are not guaranteed to work with other versions of prosemirror-view. Ensure that your version of prosemirror-view matches the version in React ProseMirror's peer dependencies!

3mcd: Hmmm okay. Everything seems to work with the bump, but I'll revert this change and silence the type errors I was getting.

tefkah:
was the goal to incorporate the designs that @ktyjscott shared here? https://knowledgefutures.slack.com/archives/CMAQM0BNV/p1741115441929979

i did not see it mentioned in the ticket, so maybe that was missed, but if it was skipped intentionally ofc ignore me!

3mcd:
@tefkah Good question. I am omitting the precise table styling and interactivity, instead opting to just get a basic prosemirror-tables integration working. I'll also add the attributes menu for the title and caption. But all of the additional commands from the context menu, table resizing, etc should be considered OOS for this PR.

allisonking (commented):
I think it looks great for a first pass! sorry for the merge conflicts 🙈

packages/context-editor/src/commands/tables.ts
```diff
@@ -0,0 +1,123 @@
...
+				finalOptions.withHeaderRow && index === 0 ? headerCells : cells
+			)
+		);
+	}
+
+	const table = types.table.createChecked(null, rows);
+	const tableFigure = state.schema.nodes.figure.createChecked(null, table);
+
+	return tableFigure;
+};
```
3mcd, line 78: Kind of surprising that `prosemirror-tables` doesn't export helper methods to construct tables initially

packages/context-editor/src/components/AttributePanel.tsx
```diff
@@ -36,66 +36,72 @@ const initPanelProps: PanelProps = {
...
 	const [height, setHeight] = useState(0);
 	const state = useEditorState();
-	const {
-		activeNode,
-		position: activeNodePosition,
-		setActiveNode,
-		setPosition: setActiveNodePosition,
-	} = useEditorContext();
+	const { position, setPosition } = useEditorContext();
+	const node = position === null ? null : state.doc.nodeAt(position);
```
3mcd, line 50: This might be a divisive change, but I removed the `activeNode` state from the editor context, opting to instead only store the position of the opened/active node. I then resolve the active node throughout the editor components using `state.doc.nodeAt()`. This feels a bit more prosemirror-y, and also eliminates a potential source of bugs where the `position` and `activeNode` could get out of sync.

packages/context-editor/src/components/AttributePanel.tsx
```diff
@@ -36,66 +36,72 @@ const initPanelProps: PanelProps = {
...
+
+	useEditorEffect(() => {
+		const p = state.selection.$from;
+		if (p.marks().length > 0) {
+			setPosition(p.pos);
+		}
+	}, [state]);
```
3mcd, line 57: This effect handles the opening of the attribute panel for marks.

packages/context-editor/src/components/StructureDecoration.tsx
```diff
@@ -42,36 +42,28 @@ const getBlockName = (node: Node) => {
...
 		}
+
 		const handleClick = () => {
-			if (activeNode && node.eq(activeNode)) {
-				setActiveNode(null);
-			} else {
-				setPosition(pos);
-				setActiveNode(node);
-			}
+			setPosition(blockPos === position ? null : blockPos);
```
3mcd, line 55: yay one setState instead of two

tefkah (changes requested):
looks great overall! 
the only things to fix imo are:
- attributes menu never closing
- prosemirror version mismatches (see https://github.com/pubpub/platform/pull/1242)
- failing test

but the general structure seems sound!

packages/context-editor/src/components/AttributePanel.tsx
```diff
@@ -221,11 +179,11 @@ export function AttributePanel({
...
 					position: "absolute",
 					background: "#fff",
-					top: position.top,
-					left: position.panelLeft,
+					top: offset.top,
+					left: offset.panelLeft,
 					width: 300,
 					padding: "1em",
-					height: height,
+					height: "auto",
```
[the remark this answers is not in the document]
3mcd: Good call. I rolled it back just to keep it consistent with the current animation. We can address later!

packages/context-editor/src/plugins/structureDecorations.tsx
```diff
@@ -9,20 +9,44 @@ export default () => {
...
+							node.type.name === "table_row" ||
+							node.type.name === "table_cell" ||
+							node.type.name === "table_header"
+						) {
+							nodeIsDescendantOfTable = true;
+						}
+					});
+					if (nodeIsDescendantOfTable) {
+						return;
+					}
```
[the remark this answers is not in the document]
3mcd: It was intended yes, just to demonstrate that the menu still works inside of tables. Figuring out the best way to render the block decorations specifically (like for lists) inside of the table felt a little out of scope for this PR.

packages/context-editor/src/style.css
```diff
@@ -76,13 +76,24 @@
 .ProseMirror ul,
 .ProseMirror ol,
 .ProseMirror li,
-.ProseMirror section {
+.ProseMirror section,
+.ProseMirror table,
```
tefkah, line 80: I think this should be like this. Otherwise the atrribute panel  is not in the correct place.

```suggestion
.ProseMirror .tableWrapper,
```

together with the removal of the left-border on the table below, it goes from this

image


to this

image

packages/context-editor/src/style.css
```diff
@@ -170,3 +181,96 @@
...
+	border-bottom: 1px solid #ccc;
+}
+
+.ProseMirror table td {
+	vertical-align: top;
+}
+
+.ProseMirror table {
+	border-left: 1px solid #777;
+}
```
tefkah, line 218: Unnecesary with change above. actually more than unnecessary, if not removed the left part of the table will look weird
```suggestion
```

packages/context-editor/src/style.css
```diff
@@ -170,3 +181,96 @@
...
+
+.wrap-widget + table {
+	position: relative;
+	top: -1px;
+}
+
+.wrap-widget button {
+	z-index: 10;
+}
```
tefkah, line 227: why is this necessary? it is causing the widgets to overlap with the menu

image

id suggest removing it


```suggestion

```

tefkah (approved):
it works great now! other than small spacing/ordering changes (i would suggest we apply them) it works great!

thank you for your service 🫡

PR #1210, opened 2025-04-29
Add valkey cache handler to next app

### LINKED ISSUE #1131: Autocache using redis
## Motivation

So we don't rely on file system based storage that slows down or crashes the app for large queries.

## Requirements

- A [ValKey](https://github.com/valkey-io/valkey) service is added to our dev, self-host, and production environments.
- Next is configured to store data cached with `unstable_cache` in ValKey.
- A maximum entry size is set to prevent extremely large queries from being cached.

## Acceptance Criteria

- Include the output of `mask tf plan` in the PR, and run `mask tf apply` only after the PR is approved.
- Justify the maximum entry size with some napkin math.

## Extra work

- Instrument ValKey to report cache hits/misses _somewhere_. Neither [Honeycomb](https://docs.honeycomb.io/send-data/telemetry-pipeline/sources/), nor [Sentry](https://github.com/getsentry/sentry-python/issues/3543) support ValKey as a data source. So this might be tricky.

kalilsn:
## Issue(s) Resolved
#1131 
This PR changes core to use valkey as its cache, as opposed to the filesystem. Theoretically it should be automatically instrumented by sentry since we're using a supported redis client.

I haven't yet set a max size, because I'm really not sure what it should be. IMO we shouldn't really worry about that yet, since the large values are a problem for the node server which can't gracefully run out of memory (which we should fix), whereas valkey will simply evict keys. Once we have some data about how the cache is performing, we can tune that configuration more.

## Test Plan
I've tested it locally and confirmed that it's writing and reading from the cache by running `redis-cli monitor` (you need to use the compiled app, not the dev server, since the dev server doesn't use the cache apparently). I'm not sure how else to test this before prod, but I feel good about it!


core/cache-handler.mjs
```diff
@@ -1,11 +1,88 @@
...
+		/** @type {import("redis").RedisClientType | null} */
+		let redisClient = null;
+		if (PHASE_PRODUCTION_BUILD !== process.env.NEXT_PHASE) {
+			try {
+				redisClient = createClient({
+					url: process.env.VALKEY_URL,
+					pingInterval: 10000,
+				});
+				redisClient.on("error", (e) => {
+					if (typeof process.env.NEXT_PRIVATE_DEBUG_CACHE !== "undefined") {
```
tefkah, line 36: maybe set this to be `if (process.env.CACHE_LOG) {` ? cuts down on one new env var

tefkah: nvm, this seems to be used internally in `@neshca/cache-handler` as well, let's leave it like this

tefkah (approved):
after testing it a bunch i can confirm it works swimmingly! for some reason it _does_ work in dev for me, but this has always been a bit fickle before this change, sometimes it does use the cache, sometimes it doesn't. shouldnt be relied on probably.

the only thing id ask: could you add a small entry to `docs` describing that we use valkey for the cache, and how to debug it? (ie by setting `NEXT_PRIVATE_CACHE_DEBUG=1`)

otherwise great work! it's just as fast if not slightly faster than the file system cache handler we were using, which probably means it's a lot faster on prod.

kalilsn:
>after testing it a bunch i can confirm it works swimmingly! for some reason it does work in dev for me, but this has always been a bit fickle before this change, sometimes it does use the cache, sometimes it doesn't. shouldnt be relied on probably.

Ok that makes me feel better about setting it up in dev. I was sure I've experienced caching behavior with the dev server but it prints a pretty definitive message saying the cache won't be used.

> the only thing id ask: could you add a small entry to docs describing that we use valkey for the cache, and how to debug it? (ie by setting NEXT_PRIVATE_CACHE_DEBUG=1)

Absolutely!

> otherwise great work! it's just as fast if not slightly faster than the file system cache handler we were using, which probably means it's a lot faster on prod.

That's great! If sentry doesn't actually give us good performance data in prod, I will manually instrument the cache handler so we can be sure. And even more importantly, this means all of our instances will use the same cache, so I'll bump our instance count up a bit after this is merged.

PR #1211, opened 2025-04-29
dev: add healthcheck for core

tefkah:
## Issue(s) Resolved

`core` going down and us not knowing about it

- Add healthcheck to `Dockerfile` directly. 
- Could likely also do this only in the ECR config, but i don't know where to put that ¯\_(ツ)_/¯
- pings core every 30s to see if everything is gucci.


kalilsn:
This is probably a good idea still for self-hosting but I learned that ECS ignores this kind of healthcheck, so I ended up setting it in our load balancer

tefkah:
ah fantastic! just looked that up too https://stackoverflow.com/a/75334903

then this can safely be closed :)

tefkah:
(for self-hosting i already added a health-check to the `docker-compose.yml`)

Issue #1148, opened 2025-04-07
bug: email action invitation form links with no text don't render


gabestein:
### What went wrong, step-by-step?
1. Create an email with a form invitation with not text anchor, e.g.: `:link{form="form-name"}`
2. Send email by running action
3. In sent email, no link will render
4. Add anchor text
5. Link will render

### What did you expect to happen?
Link token with no text anchor should be replaced with URL (e.g. `https://app.pubpub.org/magic-link?token=...`

PR #1222, opened 2025-05-01
Add default text for form links

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/1148

Make sure we still include some texts for form links. This wasn't working for forms because we figure out their links (with auth tokens) later in the flow, and the logic that defaults to the link occurs earlier than when the forms get their links.

## Test Plan
Follow steps in the issue, but your email should have some text.

With `:link{form="review"}`
image

With `:link{form="review" text="test"}`
image


core/lib/server/render/pub/renderMarkdownWithPub.ts
```diff
@@ -264,6 +264,14 @@ const renderMarkdownWithPubPlugin: Plugin<[utils.RenderWithPubContext]> = (conte
...
 							pubId: context.pub.id as PubsId,
 						});
+						assert(isParent(node));
+						// Include default text
+						node.children = [
+							{
+								type: "text",
+								value: attrs.text ?? props.href,
+							},
+						];
```
allisonking, line 274: this is copied from 

https://github.com/pubpub/platform/blob/974b002ee13d1d61accf775c06e13cf226dacd2d/core/lib/server/render/pub/renderMarkdownWithPub.ts#L168-L174

in `visitLinkDirective`. However, the `href` at this point is just an empty string because this defers the creation of the link to further down in the file

https://github.com/pubpub/platform/blob/974b002ee13d1d61accf775c06e13cf226dacd2d/core/lib/server/render/pub/renderMarkdownWithPub.ts#L261-L266

I'm not sure if there's a way we can reuse what was already there, but I'm guessing not since this exception for form links was already carved out.

3mcd (approved):

PR #1226, opened 2025-05-06
Increase connection count and set timeouts on pool

kalilsn:
## Issue(s) Resolved
Crashing every time the arcadia sitebuilder process runs 🤞 
In addition to memory issues, we can easily run out of database connections in the app, because they weren't being released back to the pool. We have plenty remaining from the db perspective (set to 225 rn, but that's just an aws default that we can change), so this change shouldn't be an issue. I've cautiously increased the max per container to 20, since we can theoretically scale up to 10 core containers (10*20 = 200, 200 < 225), though we could probably go higher.

https://knowledgefutures.slack.com/archives/CMAQM0BNV/p1746218771362249

## Test Plan
Deploy, wait for arcadia to publish, hope we don't see any downtime! I've also updated our [sentry uptime checker](https://kfg.sentry.io/alerts/rules/uptime/v7-core/107607/details/?project=4505959187480576&statsPeriod=24h) to alert us properly if it does go down.


tefkah (approved):
Great! I think increasing it to 20 is a good call. Just to be sure, did we actually notice Arcadia crashing while publishing again, or is this just precaution? I agree with increasing the pool size either way (10 is a very low default), just so i know to be a bit more suspisious of the number of connections we're establishing in the future as a possible problem

kalilsn:
Just precaution and following through on Friday's research! I haven't seen any crashes! I would imagine it could still cause a crash at this point due to memory, but ECS should already handle that a little better since there's a container restart policy now in case the process actually dies. And running out of connections in the pool will also force a restart now since the health check requires a database connection to succeed!

kalilsn:
In fact, it looks like arcadia published a bunch of times on tuesday and we didn't see any downtime at all, so maybe the memory isn't even a problem!

PR #1234, opened 2025-05-08
Fix clear cache to clear valkey instead of filesystem cache

### LINKED ISSUE #1233: sandbox users stuck on Settings page
1. User logs into either Sandbox community
2. User is redirected to user settings page
3. User selects a Community from the sidebar
4. User is redirected to user settings page

User should be redirected to the Community's Workflows Dashboard

https://sandbox.pubpub.org/settings

### What browser/operating system were you using?
Chrome

Wake Forest

kalilsn:
## Issue(s) Resolved
Resolves #1233 
After #1131 we no longer cache data on the filesystem, and instead store it in redis. In both local development and in the sandbox deployments, we regularly run the reset command, which also attempts to clear the cache. Since that cache-clearing function wasn't working, we could end up with stale data in the sandbox cache.

## Test Plan
It's easiest to see this clear the cache if you install redis-cli (or some kind of gui redis app). `redis-cli` monitor will let you see commands as they come in, and `redis cli scan 100` will print 100 keys so you can see if there's data in the cache.
- With the valkey cache running (`pnpm -w dev:cache:start`) and after using the app a bit, run `redis cli scan 100` to show there's data in the cache
- Optionally, run `redis-cli monitor` in a separate terminal
- Run `pnpm clear-cache`
- Scan again and see that the cache is clear + observe the flushall command sent in the monitor window

To actually recreate the initial sandbox bug requires a bit more effort, since you need to check out a previous commit (`1c21713d4321f728ee4b6362f2b63675542868c1`) and run reset.

I've also confirmed that this reset succeeds on lightsail by sshing into the preview instance for this branch and confirming that the migrations container exits with status 0 and that you can see the cache clearing command run in the logs for that container.

## Notes

I don't _think_ this is an issue in prod. My guess is the reason we're seeing strange behavior is that the seed scripts have a mixture of hardcoded and generated data, so we end up with some cache hits to stale data after reset. But that doesn't entirely make sense, so lmk if you have other ideas.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -17,7 +17,7 @@
 		"db:prisma": "pnpm prisma",
 		"db:studio": "pnpm prisma studio",
 		"db:generate-history-table": "pnpm exec tsx prisma/scripts/history-tables/generate-history-table.mts",
-		"clear-cache": "dotenv -e .env.local -e .env.development -- tsx --eval \"import {getRedisClient} from 'lib/redis'; (async () => {await (await getRedisClient()).flushall(); process.exit(); })();\"",
+		"clear-cache": "dotenv -e .env.local -e .env.development echo FLUSHALL | nc ${VALKEY_HOST} 6379",
 		"dev": "next dev -p 3000 --turbo | pino-pretty",
 		"build": "SKIP_VALIDATION=true NODE_OPTIONS=\"--max-old-space-size=8192\" next build",
 		"invite-users": "dotenv -e .env.local -e .env.development  tsx scripts/invite.ts",
--- a/core/package.json
+++ b/core/package.json
@@ -17,7 +17,7 @@
 		"db:prisma": "pnpm prisma",
 		"db:studio": "pnpm prisma studio",
 		"db:generate-history-table": "pnpm exec tsx prisma/scripts/history-tables/generate-history-table.mts",
-		"clear-cache": "rm -rf .next/cache/fetch-cache",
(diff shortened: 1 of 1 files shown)


core/package.json
```diff
@@ -17,7 +17,7 @@
 		"db:prisma": "pnpm prisma",
 		"db:studio": "pnpm prisma studio",
 		"db:generate-history-table": "pnpm exec tsx prisma/scripts/history-tables/generate-history-table.mts",
-		"clear-cache": "rm -rf .next/cache/fetch-cache",
+		"clear-cache": "dotenv -e .env.local -e .env.development -- tsx --eval \"import {getRedisClient} from 'lib/redis'; (async () => {await (await getRedisClient()).flushall(); process.exit(); })();\"",
```
kalilsn, line 20: this is pretty gross. i'm willing to hear that it should go in a file but i'm not sure where we'd like to put it

tefkah: i think we should just make a `core/scripts` again and put it there, i agree that this is pretty evil

tefkah (approved):
Seems to work! No clue what's happening in the preview env tho

kalilsn:
>No clue what's happening in the preview env tho
Failing to load env vars :(. `dotenv` doesn't put env vars in the environment so I was using `source`. but source throws an error if the passed file doesn't exist (`.env.local`), so I would have needed some pretty ugly conditionals to make that version work everywhere.

tefkah:
my guess is that that will also not work, could you not do `echo 'FLUSHALL' | dotenv -e .... -- nc ...`?

tefkah:
if that also doesn't work i'd just write a separate script 😅

Issue #1209, opened 2025-04-29
bug: URL fields don't appear as links in a paragraph element


eastofesten:
### What went wrong, step-by-step?
1. A Community has a URL schema field attached to a Pub Type
2. When creating a paragraph element in a form, a user inserts a field token using that URL schema field. 
3. When testing the form, the value of said field appears as text (not a hyperlink). 

### What did you expect to happen?
The value rendered by the token would appear as a hyperlink (as it does in an email.) 

### Have any screenshots? (Attach them by dragging the image here)

<img width="786" alt="Image" src="https://github.com/user-attachments/assets/60205061-2b47-4b46-a07c-91d7cfa2274d" />

tefkah:
hmm yes that would probably be best.

this should probably be a part of a slightly larger, where we provide some structure on how to render each pubfield in text. eg render dates like X, render links like Y, file-uploads like Z etc..

like, what do we do with RichText fields for instance?

PR #1239, opened 2025-05-12
feat: store richtext data as HTML rather than a Prosemirror tree

tefkah:
- **feat(serialization): add first pass at serialization**
- **feat(ui): add new show more component**
- **feat(serialization): add proper serialization libs**
- **refactor(serialization): parse and convert prosemirror tree during validation**
- **feat(fts|serialization): properly index html**
- **fix(serialization): sort of fix loading**
- **fix: fix weird redis install issue**
- **dev: make legacy seed have useful richtext**
- **dev: make legacy seed have useful richtext**
- **feat(math): serialize math to mathml and back reliably**
- **feat(serialization): find a happy way to store html and serialize on client**
- **chore: remove old validation function, fix test**
- **fix: fix the stories**
- **fix: really fix math, cleanup**
- **feat: add migration to clear current richtext values**


core/prisma/migrations/20250501133901_update_fts/migration.sql
tefkah, line 1: basically this allows for much better indexing of html by first stripping all the tags essentially.
this could be improved by creating to sets to index: headings and body, and weighting the headings more heavily in the search algo. i didn't do that for now as that's kinda overkill.

packages/context-editor/src/schemas/math.ts
```diff
@@ -1,4 +1,41 @@
...
+	// this is not nice, i would like to avoid manually calling `document.createElement`
+	// as we now need to keep track of setting `global.document` when this is called on the server
+	// forturnately, we only really server render the HTML when importing Legacy text content
+	// could be easily solved if `toDOM` was passed the document object, but ProseMirror does not seem to intend providing this
+	// https://discuss.prosemirror.net/t/getting-a-hold-of-the-document-used-by-prosemirror-from-todom/8392
+	if (!global.document) {
+		throw new Error(
+			"document not found. Trying to serialize math in a non-browser environment. To do this, set `global.document` to eg a `JSDOM` document before serializing."
+		);
+	}
```
tefkah, line 23: this is very annoying, if you know of a better way to have prosemirror convert a raw html string to a node let me know!

packages/schemas/src/index.ts
```diff
@@ -19,38 +19,27 @@ import {
...
+	[CoreSchemaType.URL]: () => URL,
+	[CoreSchemaType.Vector3]: () => Vector3,
+} as const satisfies Record<CoreSchemaType, (config?: unknown) => (typeof Schemas)[CoreSchemaType]>;
+
+export function getJsonSchemaByCoreSchemaType<C extends CoreSchemaType>(
+	coreSchemaType: C,
+	config?: unknown
+) {
+	return schemaMap[coreSchemaType](config) as ReturnType<(typeof schemaMap)[C]>;
 }
```
tefkah, line 43: this just provides better type info

packages/ui/src/show-more.tsx
```diff
@@ -0,0 +1,84 @@
...
+						<>
+							{showMoreText}
+							<ChevronDown size={16} />
+						</>
+					)}
+				</Button>
+			</div>
+		</div>
+	);
+}
```
tefkah, line 84: todo: ideally this does not show "show more" if the thing is smaller than the max size

core/prisma/seeds/ponies.snippet.html
tefkah, line 1: this is this article, which is one of the more complex ones we have JOTE
ideally we would have a nice large body that tests every feature of the editor

https://journal.trialanderror.org/pub/jointregeneration/release/6

core/cache-handler.mjs
```diff
@@ -191,6 +191,7 @@ function createRedisHandler({
 
 async function getCacheHandlerPromise() {
 	let redisClient = null;
+	console.log("im walkin here");
```
tefkah, line 194: whoops, ill remove this

core/app/components/forms/elements/ContextEditorElement.tsx
```diff
@@ -17,81 +18,73 @@ import { ContextEditorClient } from "../../ContextEditor/ContextEditorClient";
...
+								// if someone changes the doc but then reverts it--that will still count as dirty since react-hook-form is tracking that
+								const hasChanged = docHasChanged(initialDoc ?? EMPTY_DOC, state);
+								if (hasChanged) {
+									f.onChange(serializeProseMirrorDoc(state.doc));
+								}
+							}}
+							initialHtml={initialHtml}
+							disabled={disabled}
+							className="max-h-96 overflow-scroll"
+						/>
```
tefkah, line 67: so this is a bit cringe: we are still intermittently setting the value of a `RichText` field to the prosemirror tree in the form. i thought it a bit too much to convert the prosemirror tree to `html` on each change.

what's annoying is that this sort of breaks the nice dichotomy i had where only the actual `ContextEditor` knew that the `RichText` field should be converted to a `ProseMirror` tree, everything else just thinks it's HTML. this idea is maybe flawed in general.

core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx
```diff
@@ -221,6 +219,17 @@ const PubValue = ({ value }: { value: FullProcessedPubWithForm["values"][number]
...
+	if (value.schemaName === CoreSchemaType.RichText) {
+		return (
+			<ShowMore animate={false}>
+				<div
+					className="prose prose-sm"
+					dangerouslySetInnerHTML={{ __html: value.value as string }}
+				/>
+			</ShowMore>
+		);
+	}
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:231]: ideally we wouldnt do manual check here, but have InputComponents maybe define some default `ReadOnly` version of themselves which can be displayed here.

Comment (tefkah) [reply core/app/c/[communitySlug]/pubs/[pubId]/components/PubValues.tsx:231]: i might look into this here #1209

core/app/components/pubs/PubEditor/PubEditorClient.tsx
```diff
@@ -56,28 +61,48 @@ const preparePayload = ({
...
+		// we manually retrieve the value from the context editor here instead of having RHF set it
+		// bc this is much faster for very large documents
+		if (schemaName === CoreSchemaType.RichText && val === EvilContextEditorSymbol) {
+			const getter = contextEditorGetters[slug];
+			if (!getter) {
+				throw new Error(`No getter found for slug ${slug}`);
 			}
+			const value = getter.current?.getCurrentState();
+
+			valuesPayload[slug] = value && serializeProseMirrorDoc(value.doc);
```
tefkah, line 91: we could also just convert it to HTML here, let me know your opinion on that! that way, the server action does not need to be able to handle prosemirror at all, and we can maybe simplify the validation code somewhat

core/lib/server/validateFields.ts
```diff
@@ -4,13 +4,28 @@ import { getJsonSchemaByCoreSchemaType } from "schemas";
...
-	} catch {
-		return false;
+
+		const html = prosemirrorToHTMLServer(node);
+
+		return { success: true, value: html };
+	} catch (e) {
+		return { success: false, error: e };
 	}
 };
```
tefkah, line 30: this is somewhat of a big change, as it turns the `validation` into `parsing + coercion`. i don't think this is incorrect per se, but it's quite a large change.

this is not necessary per se anymore: we could, on form submission, make the client convert the prosemirror tree to html instead. let me know your opinion here!

core/prisma/migrations/20250501133901_update_fts/migration.sql
```diff
@@ -0,0 +1,140 @@
...
+            '&[^;]+;', -- removes HTML entities
+            ' ',
+            'gi'
+        ),
+        '\s+', -- collapse multiple spaces
+        ' ',
+        'g'
+    );
+END;
+$$ LANGUAGE plpgsql IMMUTABLE;
```
tefkah, line 22: `immutable` means its a pure function. postgres allegedly does some stuff with that info

core/prisma/seeds/ponies.snippet.html
tefkah, line 1: i think i can legally include this, bc the license is CC BY. might need to include a header that does the attribution

packages/context-editor/src/ContextEditor.tsx
```diff
@@ -75,12 +95,28 @@ export default function ContextEditor(props: ContextEditorProps) {
...
+	// this makes (some) sense in this case bc the editor is almost fully uncontrolled:
+	// you cannot pass in an `EditorState` that holds the actual value
+	useImperativeHandle(
+		props.getterRef,
+		() => ({
+			getCurrentState: () => editorState,
+		}),
+		[editorState]
+	);
+
```
tefkah, line 110: this is nifty, but a bit hacky. as i mention in the body, we might want to make the context editor more of a controlled component in the future in order to sync transluded pubfields, eg if i change `abstract:title` in the form i might expect it to change automatically here. we should first figure out transclusion tho

packages/context-editor/src/utils/serialize.ts
```diff
@@ -0,0 +1,68 @@
...
+
+	let toBeProcessedContent =
+		node instanceof Node ? node.content : Node.fromJSON(baseSchema, node).content;
+
+	const fragment = DOMSerializer.fromSchema(baseSchema).serializeFragment(toBeProcessedContent, {
+		document: doc,
+	});
+
+	const container = doc.createElement("div");
+	container.appendChild(fragment);
```
tefkah, line 31: no sure if there's a more elegant way to do this

packages/ui/src/show-more.tsx
tefkah: also this should maybe be a proper `<details><summary>` element for accessibilty

packages/context-editor/src/utils/serialize.ts
3mcd: From about ten minutes of searching, no, it doesn't seem like there's a much better way.

core/app/components/forms/elements/ContextEditorElement.tsx
```diff
@@ -1,97 +1,127 @@
...
+				<FormMessage />
+			</FormItem>
+		);
+	},
+	(prevProps, nextProps) => {
+		// delete prevProps.field;
+		// delete nextProps.field;
+		return true;
+	}
+);
```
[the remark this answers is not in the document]
tefkah: yeah it wasn't really necessary anymore, i was experimenting with some perf improvements/memoization and forgot to pull it out!

core/lib/editor/serialize-server.ts
```diff
@@ -0,0 +1,66 @@
...
+			value: unknown;
+			schemaName: CoreSchemaType;
+		}>;
+	},
+>(
+	pub: T,
+	opts?: {
+		toJson?: boolean;
+	}
+): T => {
```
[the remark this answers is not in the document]
tefkah: bc it's you i changed it <3

but i agree in the case of the generic here: that was kind of hard to read. 
for one-of types like the `opts` here i personally find just inlining it easier to read, but i don't feel that strongly about it!

PR #1247, opened 2025-05-14
fix: fix formswitcher bug

tefkah:
## Issue(s) Resolved

> ![NOTE]
> Made this a separate PR bc it messed up the tests, may need more testing

I added this to get around the following issue (see vid):

1. Go to create a Pub.
2. A pubId is internally initialized using `crypto.randomUUID()` in bc the UploadElement needs to know about it
3. Switch form
4. Content is reloaded! therefore `crypto.randomUUID()` is called again!
5. Create Pub
6. Redirect fails, bc clientside pubid is different from server one

To solve this, I madet this page always have a UUID through the searchParam. This is not optimal, if you navigate back from the edit page to this page it will behave somewhat strangely. That could be solved by doing an additional check if a UUID _is_ provided, namely to check whether that Pub already exists. If it does, try once more.

I'm not a big fan of this solution, as it forces an extra redirect. 

Another solution is to properly propagate the changes in UUID on the server to the client. The problem with this is the following:

1. Go to create a Pub.
2. Upload a File.
3. Switch the view
4. Create Pub
5. Upload file "PubId" no longer matches the pub it belongs to.

Maybe that's not a big deal and we should just do that, ill leave it to you!

The bug 

https://github.com/user-attachments/assets/c5cd30dd-830a-4a15-adf2-e375d5cf8ef1

### DIFF
--- a/core/app/c/[communitySlug]/pubs/create/page.tsx
+++ b/core/app/c/[communitySlug]/pubs/create/page.tsx
@@ -1,5 +1,6 @@
 import type { Metadata } from "next";
 
+import { unstable_cache } from "next/cache";
 import { notFound, redirect } from "next/navigation";
 
 import { type PubsId, type PubTypesId } from "db/public";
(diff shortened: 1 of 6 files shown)


3mcd:
@tefkah Please review the `unstable_cache` approach I pushed up. Seems to work well, although I'm not sure of the potential edge cases this might introduce.

PR #1248, opened 2025-05-14
rfc: add a new color field

tefkah:
## Issue(s) Resolved

No nice way of storing colors (necessary eg for storing community color pallete and pub background colors).

Adds a new way to store colors in PubPub by adding a new `Color` `CoreSchemaType`.

Also adds an associated color picker `FormElement`. 

Currently colors can only be stored as hex values. I added some support for at least accepting and validating RGB and HSL values, but you can't pick those yet.  That is mostly because the color pickers that I found are all too big and ugly, this one (`react-colorful`) that I kind of liked and it's kind of minimal (both in looks and in bundle size). In the future, I think we can create our own little color picker and accept multiple values that we want. I just didn't want to put that much time into it for a feature that I don't even know if we're going to add, although it seems like we are going to go ahead with it now.  

The `ColorPicker` `FormElement` can be configured to use presets, which I think is a very common use case for colors. I think you more rarely want people to pick _any_ color, and more likely want them to pick from a palette of certain colors. 
I added a `presetsOnly` option to let people pick _only_ pick from the predefined presets, or (by default) to pick from the presets or pick any color. 

## Test Plan

1. Go to eval editor on `starter`
2. set some presets
3. See them appear on the form when you create an `Evaluation`

### Update 20250519

https://github.com/user-attachments/assets/c64f5245-37a0-4105-bd65-6c7e65596f67

### Future work

- [ ] Add some "preset presets", eg some nice default palette(s) people can pick from, eg like Notions colors. 
- [ ] Allow admins to define some preset colors outside of the form builder, eg kind of like the variables for actions were going to build #1251 
- [ ] Allow admins to specify what color types to accept/store, eg RGB, HSL etc.
- [ ] Add support for very different color types, eg CMYK? Different color spaces? maybe too much.


core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
```diff
@@ -53,6 +54,17 @@ const DatePicker = dynamic(() => import("ui/date-picker").then((mod) => mod.Date
...
+	() =>
+		import("~/app/components/ContextEditor/ContextEditorClient").then(
+			(mod) => mod.ContextEditorClient
+		),
+	{
+		ssr: false,
+		loading: () => <Skeleton className="h-9 w-full" />,
+	}
+);
+
```
tefkah, line 67: this was not being lazy loaded for some reason, which slows down the initial loading of the form page a lot

packages/schemas/src/formats.ts
```diff
@@ -36,8 +36,28 @@ export function IsUuid(value: string): boolean {
...
+const HSLA = /hsla[\(](((([\d]{1,3}|[\d\%]{2,4}|[\d\.]{1,3})[\,]{0,1})[\s]*){4})[\)]/;
+
+const Color = new RegExp(
+	`^(${Hex.source}|${RGB.source}|${RGBA.source}|${HSL.source}|${HSLA.source})$`,
+	"i"
+);
+
+export function IsColor(value: string): boolean {
+	return Color.test(value);
+}
```
tefkah, line 56: i do really want to support all these, but for now it only supports HEX

core/app/components/FormBuilder/ElementPanel/ComponentConfig/ColorPicker.tsx
```diff
@@ -0,0 +1,257 @@
...
+							className="my-0 mr-2 h-7 px-2 py-0 text-sm"
+							aria-label={`Change label for color preset ${label}`}
+							defaultValue={label || color}
+							onBlur={(e) => saveLabelChange(e.target.value)}
+							onKeyDown={(e) =>
+								e.key === "Enter" && saveLabelChange(e.currentTarget.value)
+							}
+							// the autofocus kinda nice tho
+							// eslint-disable-next-line jsx-a11y/no-autofocus
+							autoFocus
```
tefkah, line 70: apparently using `autoFocus` is "cringe" and "not very accessible" but it's really annoying to use otherwise

core/app/components/FormBuilder/ElementPanel/InputComponentConfigurationForm.tsx
```diff
@@ -190,6 +202,35 @@ const componentInfo: Record<InputComponent, SchemaComponentData> = {
...
+							color={color || "#000000"}
+							onChange={(value) => {
+								setColor(value);
+							}}
+						/>
+					</PopoverContent>
+				</Popover>
+			);
+		},
+	},
```
tefkah, line 233: i am writing this popover code in too many places!!

packages/schemas/src/CoreSchemaWithIcons.ts
```diff
@@ -75,4 +76,8 @@ export const SCHEMA_TYPES_WITH_ICONS: Record<
 		description: "Rich text of any length",
 		icon: BookType,
 	},
+	[CoreSchemaType.Color]: {
+		description: "A color",
+		icon: Circle,
+	},
```
tefkah, line 82: is there a better one than `Circle`

tefkah: there is!

PR #1253, opened 2025-05-20
dev: make dev:setup not reset (unless you want to)

tefkah:
## Issue(s) Resolved

Annoying that `dev:setup` always resets your db even if you don't want to.

@isTravis has mentioned losing work bc of this.

Small housekeeping PR.

- Make `dev:setup` migrate only
- Make `dev:setup:reset` reset. This now also skips the (reduntant) migration step, making `dev:setup:reset` faster than the old `dev:setup`.

## Test Plan

-


3mcd (approved):

isTravis:
This saved me a bunch of time tonight! Thanks!

PR #1255, opened 2025-05-20
fix: use doi from deposit payload

### LINKED ISSUE #1252: Datacite action not updating deposited pubs
1.
2. ...

3mcd:
## Issue(s) Resolved

Resolves #1252

Tiny PR that
- uses the DOI derived from the `makeDatacitePayload` function instead of the action config
- adds an extra log to the DataCite action for prod debugging

## Test Plan

N/A


gabestein (approved):
Pending build and tests etc, LGTM! Thanks for fixing.

PR #1260, opened 2025-05-21
fix: make inviting existing users through email field work properly

tefkah:
- **fix: check whether email belongs to existing user before creating invite**
- **fix: make it easier to delete email input**
- **fix: update existing test to wrok correctly**

### DIFF
--- a/core/actions/email/action.ts
+++ b/core/actions/email/action.ts
@@ -15,7 +15,14 @@ export const action = defineAction({
 	name: Action.email,
 	config: {
 		schema: z.object({
-			recipientEmail: z.string().email().describe("Recipient email address").optional(),
+			recipientEmail: z
+				.string()
+				.email()
+				.describe("Recipient email address")
+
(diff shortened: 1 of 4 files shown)


core/playwright/externalFormInvite.spec.ts
tefkah, line 1: apparently we had a test for this, that test just did not work correctly, whoops

core/actions/email/action.ts
```diff
@@ -15,7 +15,14 @@ export const action = defineAction({
...
 	config: {
 		schema: z.object({
-			recipientEmail: z.string().email().describe("Recipient email address").optional(),
+			recipientEmail: z
+				.string()
+				.email()
+				.describe("Recipient email address")
+
+				// makes sure that "" is interpreted as undefined
+				.transform((val) => val || undefined)
```
3mcd, line 24: I wonder if we should be doing this in more places!

Issue #1214, opened 2025-04-29
bug: Cannot delete all items from multiValue form element input


gabestein:
### What went wrong, step-by-step?
1. Create a form with a stringArray field and multiValue input type
2. Add one item and save the Pub
3. Re-open the pub and try to delete the item and save the pub
4. Upon refresh, see that the item is still there

### What did you expect to happen?
The item is deleted when removed from the multiValue select on the form

### Have any screenshots? (Attach them by dragging the image here)
<img width="734" alt="Image" src="https://github.com/user-attachments/assets/bcbc0bd3-5d94-47a3-9be7-227c48df7c5c" />

PR #1261, opened 2025-05-21
Fix multivalue delete bug

allisonking:
## Issue(s) Resolved
Closes https://github.com/pubpub/platform/issues/1214

### DIFF
--- a/core/lib/server/pub.ts
+++ b/core/lib/server/pub.ts
@@ -209,10 +209,6 @@ export const doesPubExist = async (
 	return exists ? { exists: true as const, pub: pubs[0] } : { exists: false as const };
 };
 
-const isRelatedPubInit = (value: unknown): value is { value: unknown; relatedPubId: PubsId }[] =>
-	Array.isArray(value) &&
-	value.every((v) => typeof v === "object" && v && "value" in v && "relatedPubId" in v);
-
 /**
  * Transform pub values which can either be
  * {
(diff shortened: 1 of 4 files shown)


core/lib/server/pub-filters.db.test.ts
```diff
@@ -157,9 +157,9 @@ const seed = createSeed({
 			id: relationId,
 			pubType: "Basic Pub",
 			values: {
-				Relation: null,
 				Number: 50,
 			},
+			relatedPubs: {},
```
allisonking, line 162: this seed would fail after I made the changes to `normalizePubValues`. I think related pubs are meant to be declared via a separate `relatedPubs` key (outside of `values`) so I think this was originally not declared quite right (but also not used in the tests) but let me know if that's a bad assumption!

core/lib/server/pub.ts
```diff
@@ -225,18 +221,36 @@ const isRelatedPubInit = (value: unknown): value is { value: unknown; relatedPub
...
 ) => {
-	return Object.entries(pubValues).flatMap(([slug, value]) =>
-		isRelatedPubInit(value)
-			? value.map((v) => ({ slug, value: v.value, relatedPubId: v.relatedPubId }))
-			: ([{ slug, value, relatedPubId: undefined }] as {
-					slug: string;
-					value: T;
-					relatedPubId: PubsId | undefined;
-				}[])
+	const fields = await getFieldInfoForSlugs(
```
allisonking, line 229: I didn't really want to query for the fields here, but I couldn't think of another way around this. the main problem is that both arrays and related pubs can show up as an empty array `[]`. In this step, we need to determine the values that are relations vs not relations, and afaik there's no other way to determine what `[]` should be—an empty string array, or an empty related field?

allisonking: we call `normalizePubValues` in three places: here, in `updatePub`, and `createPubNewRecursive`. we don't do it here, but in `updatePub` and `createPubRecursive` there is an additional validation step that _also_ queries the fields, which is why I hadn't wanted to add it here as well initially, but couldn't think of a better way

tefkah: hmmmm yeah that is tricky... this once again shows i think we need some consistent way of marking values as "empty" or "to be removed", rather than just having `""` or `[]` for fields that really should be empty.

it does feel annoying to add an additional to get the pub metadata again. another option would be to reformat the input for this action, eg include the `schemaName` for each thing, but then we would have to rely on the caller to supply that data accurately which is also not good to do.

i think we should just rework this as part of our big Form refactor work at some point. maybe @3mcd was right (as he always is haha) and the validation should not be a part of eg `pubOp`, but a step we do before it (along with fetching the correct metadata for fields).

maybe there is a way we can get around it though: could we get away with interpreting `[]` as only "arrayish" values, never as an empty relatedPub field? 
currently, for both `MultiValueInput` and `RelatedPubBlock`, we send an empty one up as `[]`. while for the `MultiValueInput` that does seem necessary, for the `RelatedPubBlock` don't interpret the `[]` anyway: we look at the `deleted` property. while maybe kind of hacky, i feel this might be slightly nicer? again i think we should handle this whole flow differently most likely, but this could be a more minimal change.

Issue #1194, opened 2025-04-28
All Pubs Refactor - Pub Card


gabestein:
## Motivation

So users can find Pubs more easily

## Requirements

- Refactors All Pubs (and Workflows, since it's the same component) Pub cards to match mocks, including:
   - Type button (eventually will link to filtered view of All Pubs page)
   - Stage dropdown with move and "view stage" buttons
   - Relations dropdown with list of relations, ordered by type
   - Relate date updated
   - On-hover right-hand menu group that includes buttons to archive, take an action, update, or "select" the pub

## Mockups (if available)
https://www.figma.com/design/SqOTMElJn7fuPljfQL1gIY/PubPub-Platform?node-id=2641-49184&t=FR304cQACuF0sjrW-0

gabestein:
Already wondering if we should show both date created and date updated

PR #1270, opened 2025-05-27
Pub card

allisonking:
## Issue(s) Resolved
Most of https://github.com/pubpub/platform/issues/1194

Implements the new design for pub cards. Attempts to add storybook, but the mocking wasn't working well for me so I gave up for now :( 

Does not implement the "Relations" menu yet as that will require a new query, and this PR was already getting rather large.

## Test Plan

Visit the pub listings page and try out the different buttons! These cards will also appear on the individual stage pages too.

image
image

## Notes
This also moves `storybook` into `core`. I thought it'd be easier to import server components this way, but it's still hard 🥲

### DIFF
--- a/core/actions/email/action.ts
+++ b/core/actions/email/action.ts
@@ -15,7 +15,14 @@ export const action = defineAction({
 	name: Action.email,
 	config: {
 		schema: z.object({
-			recipientEmail: z.string().email().describe("Recipient email address").optional(),
+			recipientEmail: z
+				.string()
+				.email()
+				.describe("Recipient email address")
+
+				// makes sure that "" is interpreted as undefined
+				.transform((val) => val || undefined)
+				.optional(),
 			recipientMember: z.string().uuid().describe("Recipient member").optional(),
 			subject: stringWithTokens().max(500).describe("Email subject"),
 			body: markdown().min(0).describe("Email body"),
@@ -41,7 +48,13 @@ export const action = defineAction({
 	params: {
 		schema: z
 			.object({
-				recipientEmail: z.string().email().describe("Recipient email address").optional(),
+				recipientEmail: z
+					.string()
+					.email()
+					.describe("Recipient email address")
+					// makes sure that "" is interpreted as undefined
+					.transform((val) => val || undefined)
+					.optional(),
 				recipientMember: z
 					.string()
 					.uuid()
--- a/core/actions/email/run.ts
+++ b/core/actions/email/run.ts
@@ -1,6 +1,9 @@
 "use server";
 
-import type { CommunityMembershipsId } from "db/public";
+import type { Kysely } from "kysely";
+
+import type { Database } from "db/Database";
+import type { CommunitiesId, CommunityMembershipsId } from "db/public";
 import { logger } from "logger";
 import { assert, expect } from "utils";
 
@@ -12,8 +15,66 @@ import * as Email from "~/lib/server/email";
 import { maybeWithTrx } from "~/lib/server/maybeWithTrx";
 import { coalesceMemberships, selectCommunityMemberships } from "~/lib/server/member";
 import { renderMarkdownWithPub } from "~/lib/server/render/pub/renderMarkdownWithPub";
+import { getUser } from "~/lib/server/user";
 import { isClientException } from "~/lib/serverActions";
 import { defineRun } from "../types";
+
+const resolveRecipient = async (
+	recipientEmail: string | undefined,
+	recipientMemberId: CommunityMembershipsId | undefined,
+	communityId: CommunitiesId,
+	trx: Kysely<Database>
+): Promise<NonNullable<RenderWithPubContext["recipient"]>> => {
+	if (recipientMemberId !== undefined) {
+		const memberships = await selectCommunityMemberships({
+			id: recipientMemberId,
+		}).execute();
+		if (!memberships.length) {
+			throw new Error(`Could not find member with ID ${recipientMemberId}`);
+		}
+
+		const membership = coalesceMemberships(memberships);
+
+		return {
+			id: membership.id,
+			user: membership.user,
+		};
+	}
+
+	if (!recipientEmail) {
+		throw new Error("No recipient was specified");
+	}
+
+	// check if user exists
+	const user = await getUser({ email: recipientEmail }, trx).executeTakeFirst();
+
+	// this email is not associated with a user
+	if (!user) {
+		return {
+			email: recipientEmail,
+		};
+	}
+
+	// check if that user is a member of the community then
+	const memberships = await selectCommunityMemberships({
+		userId: user.id,
+		communityId,
+	}).execute();
+
+	if (!memberships.length) {
+		// we send an invite
+		return {
+			email: recipientEmail,
+		};
+	}
+
+	// we send the email to the user
+	const membership = coalesceMemberships(memberships);
+	return {
+		id: membership.id,
+		user: membership.user,
+	};
+};
 
 export const run = defineRun<typeof action>(
 	async ({ pub, config, args, communityId, actionRunId, userId }) => {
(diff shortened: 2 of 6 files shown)


core/.storybook/__mocks__/Stages.actions.mock.ts
```diff
@@ -0,0 +1,3 @@
+import { fn } from "@storybook/test";
+
+export const move = fn();
```
allisonking, line 3: I couldn't get this to work, even though as far as I could tell I mimicked the same mocks as `PubEditor.actions.mock.ts` 🥲

core/app/components/PubCard.tsx
```diff
@@ -0,0 +1,146 @@
...
+						className="w-8 px-4 py-2 [&_svg]:size-6"
+						icon={<Trash strokeWidth="1px" />}
+					/>
+					{pub.stage ? (
+						<PubsRunActionDropDownMenu
+							actionInstances={actionInstances}
+							stage={pub.stage}
+							pubId={pub.id}
+							// @ts-ignore TODO: do we need this?
+							pageContext={{}}
```
allisonking, line 130: I couldn't quite figure out what the page context was used for. it would involve a lot of prop passing around to get it here, so I left it empty for now, but let me know if it should actually be filled out?

packages/ui/src/card.tsx
```diff
@@ -20,25 +20,22 @@ const CardHeader = React.forwardRef<HTMLDivElement, React.HTMLAttributes<HTMLDiv
 );
```
allisonking, line 20: I updated this based on the update from shadcn (see changelog at bottom)
https://ui.shadcn.com/docs/components/card

This no longer assumes the card title will be an `h3`, so I updated some other parts accordingly

core/app/components/PubCard.tsx
```diff
@@ -0,0 +1,146 @@
...
+					<PubDescription pub={pub} />
+				</CardDescription>
+				<CardFooter className="flex gap-2 p-0 text-xs text-gray-600">
+					<div className="flex gap-1" title="Created at">
+						<Calendar size="16px" strokeWidth="1px" />
+						<span>{formatDateAsPossiblyDistance(new Date(pub.createdAt))}</span>
+					</div>
+					<div className="flex gap-1" title="Updated at">
+						<History size="16px" strokeWidth="1px" />
+						<span>{formatDateAsPossiblyDistance(new Date(pub.updatedAt))}</span>
```
allisonking, line 110: there's something not quite right with the timestamps here. If I save something at 2:35pm ET, it gets saved to the db as 18:35. but by the time it is part of this `pub` object, it becomes 22:35, and I'm not sure why.

tefkah: my guess:

- when you save, it converts to UTC in DB (ET is currently at -4). that is correct behavior i think.
-  but since we don't have timestamps in the db, when you get it back i guess either the browser or node thinks it's again in your local time (18:35 at -4UTC), but then converts it back to UTC again (adding another +4).

ive run into something similar in the past too, it's quite tricky to get right and "just" displaying the date quickly leads to hydration errors. eg if we send `<span>{new Date()}</span>` from the server at 12:00 (in -4) to me (in +2, so at 18:00), hydration fails bc my browser thinks that should be 18:00 while the server sends over 12:00.

here's an article that has _some_ solution, but it's still kinda wack: https://francoisbest.com/posts/2023/displaying-local-times-in-nextjs

my suggestion: keep it as is, and we open a separate ticket to address timezone related issues dealing with time!

ktyjscott:
This is looking so good @allisonking - great first pass. Here is my feedback, please let me know if you have any questions!

**Card Base**
* Remove Shadow, but add shadow back for hover state
* Entire card should be clickable, and lead the user to the pub details page

**Hover Actions**
* Make sure all the icons on this row are using neutral-500 as their color, hard to tell from the inspector
* Let’s start utilizing the trash-2 icon from lucide instead of the regular trash one, I just think it looks a little better
* Add a margin-right of 1rem to this group of icons
* When the checkbox on a pub card is active it should be visible at all times, right now it disappears when I stop hovering the card

**Type Badge**
* Needs background color gray-100
* Needs border 1 px solid gray-300
* Remove shadow
* Set font-weight to 600 (semibold)
* Set left and right padding to 0.35 rem
* I noticed this is a button but does nothing when clicked - I think this may cause some confusion/misclicks, should we remove? Or is there a reason this is a button?

**Stage Badge**
* Remove shadow
* Set font-weight to 600 (semibold)
* Set left and right padding to 0.35 rem
* Use `gap-0.5` instead of gap-2 (this spacing looks better)
* When opened, dropdown should be anchored to the left side of the badge (currently centered)

**Relations Badge**
* Remove shadow
* Set font-weight to 600 (semibold)
* Set left and right padding to 0.35 rem
* When opened, dropdown should be anchored to the left side of the badge (currently centered)
* Type Badge within dropdown should have a line-height of 1rem
* Type Badge within dropdown should have a letter-spacing of -0.1px

**Dates Row**
* First date shown (beside calendar icon) should be the ‘created on’ date - not sure if that is something we have or something we need to include in another ticket
* Second date - let’s remove the ‘about’ language here

core/app/c/[communitySlug]/developers/docs/stoplight.styles.css
Comment (tefkah) [inline core/app/c/[communitySlug]/developers/docs/stoplight.styles.css:1]: @allisonking could you re-minify this again? suddenly adds a giant file to the codebase 😅

core/app/components/PubCard.tsx
allisonking: thanks @tefkah , converted to an issue here! https://github.com/pubpub/platform/issues/1285#issue-3118082290

core/app/c/[communitySlug]/developers/docs/stoplight.styles.css
Comment (allisonking) [reply core/app/c/[communitySlug]/developers/docs/stoplight.styles.css:1]: ugh! I think it happened when I prettier'd manually, thanks for catching that!

tefkah:
>     * Making the whole card a link to the pub details page
>       
>       * This is tricky because there are buttons on the card, and right now if you click on a button, you'll also be brought to the pub detail page. I think there are two ways around this 1) move all of the buttons outside of the card, then position them manually via `position: absolute`. I started this, but it made certain other behavior weird, like the title truncating wouldn't work because the buttons weren't taking up real width anymore. 2) make all buttons `evt.stopPropagation()`. I wasn't sure how to do this though since the buttons are all server components right now. curious if @tefkah or @3mcd have run into anything like this that they have a good solution for?

yess this is possible! i don't have a good article for it, but ive implemented something similar before, see eg the cards on https://blog.trialanderror.org : the cards are links themselves, but the buttons are too.

here is the relevant code

https://github.com/TrialAndErrorOrg/websites/blob/6e52c7430ff39eca47de92fcdb40d925809042d3/apps/next-blog/src/app/components/blog/PostCard.tsx#L1-L122

i see now ive used this "link overlay" class, which i define here:

https://github.com/TrialAndErrorOrg/websites/blob/4b09e2ab96f533f957c7ab2e06676917df8eca73/apps/next-blog/src/styles/globals.css#L115-L130

also kinda relies on this `card` class, which is kinda hard too read, but you need to have it be `relative` 

https://github.com/TrialAndErrorOrg/websites/blob/4b09e2ab96f533f957c7ab2e06676917df8eca73/apps/next-blog/src/styles/globals.css#L86-L101


okay looking over it again i remember, roughly:

- make something the actual link to the thing you want to link to on the card, eg the title
- make card `position: relative`
- make the `::after` of it fill the whole card, make it have `z-0` (or the same as the normal card)
- make other links relative as well, and set their `z-10` (or any higher number, prolly not too high

this gets around the main issue, which is that you can't (or shouldnt) have `<a>`s inside of other `<a>`s. 

> > Making the card have a blue border when selected
>
>   This is more difficult because the card is a server component at the moment. we'd have to refactor to make the checkbox a client-side component in order to keep track of whether or not it is checked. I think we may have to do something like this once we do react to an item being checked, but I'm not quite sure what that would look like. I'd vote to punt this until we are implementing selection behavior

i think you could maybe do this with css only using the [`:has` selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:has)

just set some attribute on the selector itself (`data-pub-card-selected`) or something, then use [tailwinds has syntax](https://tailwindcss.com/blog/tailwindcss-v3-4#new-has-variant) to change the border!

i think `:has` support (is good enough)[https://caniuse.com/css-has] to not think about compatibilty

allisonking:
thanks @tefkah, both of your CSS suggestions worked 💯 ! this should be ready for review again, with relations merged in!

tefkah (approved):
great work allison! i made storybook work and fixed the tests.
@3mcd dunno if you still want to look at it, i only really changed the storybook setup and a few tests which i could call out if needed!

PR #1283, opened 2025-06-03
fix: allow sse streaming to work with nginx

tefkah:
## Issue(s) Resolved

https://knowledgefutures.slack.com/archives/CMAQM0BNV/p1748968611367069

SSE wasn't working on prod because of nginx.

My approach to solving it is to do as much as we can `core` side, bc i don't realllly want to mess with the `nginx` configuration.

It wasn't working for two reasons:

1. bc of how the headers are(n't) set, the client doesn't even receive the notifications. 

This can be solved a number of ways, but the easiest way is just to follow [Next's recommendations and disable buffering](https://nextjs.org/docs/app/guides/self-hosting#streaming-and-suspense)

I have also seen success by setting the following headers instead (not sure which one does it)

```
 	proxy_set_header Host $host;
    proxy_set_header X-Forwarded-Host $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Real-IP $remote_addr;


    proxy_set_header Host $host;
    proxy_set_header X-Forwarded-Host $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-Proto $scheme;
    proxy_set_header X-Forwarded-Port $server_port;
```

2. The default `proxy_read_timeout` is `60s`, meaning after 60s of no events `nginx` closes the connection.

We could fix this by just setting it to some larger value (like `1h`), but i fixed this by just sending a heartbeat signal every 15s.

## Test Plan

Can't be tested in the preview environment, as `caddy` is used there instead. 

I would recommend just trusting me (bro), as the local setup is very annoying.

If _do_ want to test it, you can test it locally by doing the following

add the following to `docker-compose.base.yml`, under `services`

```yaml
    nginx:
        image: nginx:latest
        ports:
            - "80:80"
            - "443:443"
        volumes:
            - ./_dev/nginx:/etc/nginx/conf.d
            - ./_dev/nginx/ssl:/etc/nginx/ssl:ro
        network_mode: host
```

create a `_dev` folder, with a `nginx` folder in it.

in it, place the following script as `generate-ssl-certs.sh`

```sh
#!/bin/bash

# create ssl directory if it doesn't exist
mkdir -p ./nginx/ssl

# generate self-signed certificate for localhost
openssl req -x509 -nodes -days 365 -newkey rsa:2048 \
    -keyout ./nginx/ssl/key.pem \
    -out ./nginx/ssl/cert.pem \
    -subj "/C=US/ST=State/L=City/O=Organization/CN=localhost" \
    -addext "subjectAltName=DNS:localhost,DNS:*.localhost,IP:127.0.0.1"

echo "SSL certificates generated in ./nginx/ssl/"
echo "cert: ./nginx/ssl/cert.pem"
echo "key: ./nginx/ssl/key.pem" 
```

and add the following config as `default.conf`
```
upstream nextjs {
  server localhost:3000;
}

# redirect http to https
server {
  listen 80;
  server_name _;
  return 301 https://$host$request_uri;
}

# https server
server {
  listen 443 ssl http2;
  server_name _;

  # ssl configuration
  ssl_certificate /etc/nginx/ssl/cert.pem;
  ssl_certificate_key /etc/nginx/ssl/key.pem;
  
  # modern ssl configuration
  ssl_protocols TLSv1.2 TLSv1.3;
  ssl_ciphers ECDHE-RSA-AES256-GCM-SHA512:DHE-RSA-AES256-GCM-SHA512:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384;
  ssl_prefer_server_ciphers off;
  ssl_session_timeout 1d;
  ssl_session_cache shared:SSL:50m;
  ssl_session_tickets off;

  # security headers
#   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains" always;
#   add_header X-Frame-Options DENY always;
#   add_header X-Content-Type-Options nosniff always;
#   add_header X-XSS-Protection "1; mode=block" always;

  location / {
    proxy_pass http://nextjs;

    proxy_set_header Host $host;
    proxy_set_header X-Forwarded-Host $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Real-IP $remote_addr;

    # proxy_read_timeout 500ms;

    # proxy_set_header Host $host;
    # proxy_set_header X-Forwarded-Host $host;
    # proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    # proxy_set_header X-Real-IP $remote_addr;
    # proxy_set_header X-Forwarded-Proto $scheme;
    # proxy_set_header X-Forwarded-Port $server_port;
    
    # websocket support
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
    
    # # timeouts
    # proxy_connect_timeout 60s;
    # proxy_send_timeout 60s;
    # proxy_read_timeout 60s;
  }

  location /legacy_healthcheck {
    return 200;
    add_header Content-Type text/plain;
  }
} 
```
then

```sh
cd _dev/nginx
chmod +x generate-ssl-certs.sh
./generate-ssl-certs.sh

pnpm -w dev:setup # docker compose up that thang
pnpm -F core dev
```

now you _should_ be able to go to `https://localhost/c/legacy/stages/manage` (not `:3000`, just localhost. note the `https`). accept the scary warning.

open that in two tabs, go to the same stage on both, and on one of them look at the `actions` tab, the other on the `pubs` tab.

on the `pubs` tab one, run an action.

you should see the little green thing appear on the `actions` tab one.

## Notes

maybe the `nginx` thing should be part of our default setup.


core/lib/notify/useSSEUpdates.ts
```diff
@@ -55,8 +61,58 @@ export function useSSEUpdates<T extends NotifyTables>({
...
+	};
+
+	// we don't want to show the error toast when reloading
+	useEffect(() => {
+		window.addEventListener("beforeunload", beforeUnload);
+
+		return () => {
+			window.removeEventListener("beforeunload", beforeUnload);
+		};
+	}, []);
```
tefkah, line 82: i dont know of a better way to do this

PR #1290, opened 2025-06-17
feat: search on all pub page

### LINKED ISSUE #1197: All Pubs Refactor - Search
## Motivation

So users can search for text within their pubs, allowing them to find pubs faster

## Requirements

- Adds a search bar to the all pubs page that updates the list of pubs in real-time (with appropriate debouncing), similar to how the current search functionality works
- When results are returned, pub cards should be updated to show the field name where the match was found, as well as a highlighted snippet of text surrounding the match, similar to current search functionality
- If there are no results, the page should display a no results found message, as shown in mocks
- Using the search shortcut (cmd/ctrl-k) should focus the search bar, rather than pop up the global search as it does currently

tefkah:
## Issue(s) Resolved

Resolves #1197 

Adds search capabilities to `getPubsWithRelatedValues` and adds a searching filter to `/pubs`.

Not done:
- [ ] Made global Cmd+K search use PubCards or this component. This is somewhat tricky to do as `PubCard` is still a server component. I'd need to rework it more, which i may do in #1293 
- [ ] No results screen/message (maybe i should still add that)

Video overview (kinda rambly sorry)

https://github.com/user-attachments/assets/0b75eb3b-638d-4aa1-a32d-46d6fff29cf6

## Test Plan

1. go to legacy/pubs
2. search some stuff
3. page through the results if possible

When searching

image

Changed the pagination a bit when searching, as i didn't want to find out how many results there are in total (that's kinda costly), so I just get `N = perPage + 1` items, check if `N > perPage`, if so i allow the user to paginate to the next page.

There's an argument to make to just remove pagination for searching entirely, and only show the first ~25 results, and the user just needs to provide a better query if they aint happy.

image

### DIFF
--- a/core/app/c/[communitySlug]/pubs/PubSearchInput.tsx
+++ b/core/app/c/[communitySlug]/pubs/PubSearchInput.tsx
@@ -1,0 +1,57 @@
+"use client";
+
+import React, { useState } from "react";
+import { Search } from "lucide-react";
+import { parseAsInteger, parseAsString, useQueryStates } from "nuqs";
+import { useDebouncedCallback } from "use-debounce";
+
+import { usePlatformModifierKey } from "ui/hooks";
+import { Input } from "ui/input";
+import { cn } from "utils";
+
+type PubSearchProps = React.PropsWithChildren<{}>;
+
+export const PubSearch = (props: PubSearchProps) => {
+	const [query, setQuery] = useQueryStates(
+		{
+			query: parseAsString.withDefault(""),
+			page: parseAsInteger.withDefault(1),
+		},
+		{
+			shallow: false,
+		}
+	);
+	const [value, setValue] = useState(query.query);
+
+	const { symbol, platform } = usePlatformModifierKey();
+
+	const debouncedSetQuery = useDebouncedCallback((value: string) => {
+		if (value.length >= 2 || value.length === 0) {
+			setQuery({ query: value });
+		}
+	}, 500);
+
+	return (
+		<div className="relative flex flex-col gap-4">
+			<div className="sticky top-0 z-20 flex items-center gap-x-2">
+				<Search
+					className="absolute left-2 top-1/2 -translate-y-1/2 text-gray-500"
+					size={16}
+				/>
+				<Input
+					value={value}
+					onChange={(e) => {
+						setValue(e.target.value);
+						debouncedSetQuery(e.target.value);
+					}}
+					placeholder="Search updates as you type..."
+					className="bg-white pl-8 tracking-wide shadow-none"
+				/>
+				<span className="absolute right-2 top-1/2 hidden -translate-y-1/2 items-center gap-x-1 font-mono text-xs text-gray-500 opacity-50 md:flex">
+					<span className={cn(platform === "mac" && "text-lg")}>{symbol}</span> K
+				</span>
+			</div>
+			<div className={cn(value !== query.query && "opacity-50")}>{props.children}</div>
+		</div>
+	);
+};
--- a/core/app/c/[communitySlug]/pubs/PubSearchInput.tsx
+++ b/core/app/c/[communitySlug]/pubs/PubSearchInput.tsx
@@ -1,11 +1,11 @@
 "use client";
 
-import React, { useDeferredValue, useEffect, useState } from "react";
+import React, { useDeferredValue, useEffect, useRef, useState } from "react";
 import { Search } from "lucide-react";
 import { parseAsInteger, parseAsString, useQueryStates } from "nuqs";
 import { useDebouncedCallback } from "use-debounce";
 
-import { usePlatformModifierKey } from "ui/hooks";
+import { KeyboardShortcutPriority, useKeyboardShortcut, usePlatformModifierKey } from "ui/hooks";
 import { Input } from "ui/input";
 import { cn } from "utils";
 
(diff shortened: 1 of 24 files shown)


core/app/c/[communitySlug]/pubs/page.tsx
```diff
@@ -33,17 +41,30 @@ export default async function Page(props: Props) {
...
-					userId={user.id}
-				/>
-			</div>
-		</div>
+		<ContentLayout
+			title={
+				<>
+					<BookOpen size={24} strokeWidth={1} className="mr-2 text-gray-500" /> Pubs
+				</>
+			}
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/page.tsx:49]: added the contentlayout for the pubs page

core/app/c/[communitySlug]/pubs/PubSearchInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+	);
+
+	// local input state for immediate UI responsiveness + sync with URL
+	// otherwise, when navigating back/forward or refreshing, the input will be empty
+	const [inputValue, setInputValue] = useState(query.query);
+
+	// deferred query to keep track of server updates
+	// without this, we can't only check if inputValue !== query.query,
+	// which only tells us that the debounce has happened
+	const deferredQuery = useDeferredValue(query.query);
```
Comment (tefkah) [inline core/app/c/[communitySlug]/pubs/PubSearchInput.tsx:34]: this is kind of mysterious to me, but this allows me to keep the body of the search at half-opacity until the content has fully refreshed. this could also be done by using `useTransition`.

packages/ui/src/hooks/useKeyBoardShortcut.tsx
```diff
@@ -0,0 +1,192 @@
...
+type Modifiers = "Mod" | "Ctrl" | "Shift" | "Alt" | "Meta";
+
+// prettier-ignore
+type Keys = "a" | "b" | "c" | "d" | "e" | "f" | "g" | "h" | "i" | "j" | "k" | "l" | "m" | "n" | "o" | "p" | "q" | "r" | "s" | "t" | "u" | "v" | "w" | "x" | "y" | "z" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9" | "0" | "F1" | "F2" | "F3" | "F4" | "F5" | "F6" | "F7" | "F8" | "F9" | "F10" | "F11" | "F12" | "ArrowUp" | "ArrowDown" | "ArrowLeft" | "ArrowRight" | "Enter" | "Escape" | "Backspace" | "Tab" | "Space" | "Delete" | "Home" | "End" | "PageUp" | "PageDown" | "CapsLock" | "NumLock" | "ScrollLock" | "Pause" | "PrintScreen" | "Insert";
+
+export type KeyCombination =
+	| `${Modifiers}+${Keys}`
+	| `${Modifiers}+${Modifiers}+${Keys}`
+	| `${Modifiers}+${Modifiers}+${Modifiers}+${Keys}`
+	| `${Modifiers}+${Modifiers}+${Modifiers}+${Modifiers}+${Keys}`;
```
tefkah, line 140: niceish type that allows for fully typed `Mod+k` like strings. doesn't really cover every key (only english-y ones), but that's fine bc it's us defining the shortcuts, and we (mostly) use English keyboards

packages/ui/src/hooks/useKeyBoardShortcut.tsx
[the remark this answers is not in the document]
tefkah: sadly this doesn't yet incorporate the editor, as the keyboard shortcut logic there is handled by prosemirror itself. 

the way i was thinking of doing that was something like:
- expose some hooks to enable/disable certain keys from here
- use this hook in the body of context editor
- pass enable/disable functions to keymap plugin
- have keymap disable those also used by the editor (cmd+k)

ideally it would disable/enable the shortcuts on blur/focus as well, but that might be too fragile. if you have a better idea please let me know!

core/lib/authentication/lucia.ts
```diff
@@ -232,6 +232,15 @@ const defaultOptions = {
...
 
+/**
+ * Separately cache this call because validateRequest can be called with different options
+ * leading to cache misses
+ */
+const validateLuciaSession = cache(async (sessionId: string) => {
+	const result = await lucia.validateSession(sessionId);
+	return result;
+});
+
```
tefkah, line 243: this is just a small thing i found when looking at logs:

we sometimes call `getPageLoginData` and sometimes `getLoginData` on the same request. i would expect the call to the db to be cached between these two calls, bc `getLoginData` is called by `getPageLoginData` and `getLoginData` is `React.cache`d.

https://github.com/pubpub/platform/blob/f633a0978a9d7e0415f5d94cd2d77f1dbbd10789/core/lib/authentication/loginData.ts#L17-L51

but that doesn't happen, bc we pass in some specific arguments to `getLoginData` (`allowedSessions`) when called by `getPageLoginData`. 

this leads to `getLoginData` to be called with diff arguments most of the time (we usually don't pass anything to `getLoginData`), leading to cache misses.

So it made more sense to me cache the actual call to the db as well, as that's the main bottleneck and always called with the same arguments (for the same request): the sessionid.
this takes the number of calls to the db for this purpose from 2 to 1 in many cases.

packages/contracts/src/resources/site.ts
```diff
@@ -194,6 +194,17 @@ export type MaybePubOptions = {
...
+	/**
+	 * The search query to use for matching values
+	 */
+	search?: string;
+
+	/**
+	 * Whether to include matched and highlighted values
+	 * @default true if `search` is defined
+	 */
+	withSearchValues?: boolean;
```
tefkah, line 207: might be redundant to offer this a setting in addition to `search`, but eh

3mcd (approved):
I love how snappy the search feels, and combined with the newer/condensed pub cards and pagination, the all pubs page is starting to look reaaaally good!

PR #1314, opened 2025-07-17
feat: nicer login experience

tefkah:
- **feat: add reveal password toggle**
- **feat: make login page look a lil nicer**
- **feat: oo so nice**
- **fix: better shadow**
- **feat: use nicer login button**

## Issue(s) Resolved

-

Makes the login form look slightly nicer by

- Adding a logo
- Adding a background
- centering the form more
- making you able to see your pw
- fixing the loading states of the form

https://github.com/user-attachments/assets/be64e477-48aa-47fb-b5bf-8467ffd8a0d2

image

### DIFF
--- a/core/actions/_lib/runActionInstance.db.test.ts
+++ b/core/actions/_lib/runActionInstance.db.test.ts
@@ -107,7 +107,7 @@ describe("runActionInstance", () => {
 			report: "Logged out some data, check your console.",
 			data: {},
 		});
-	});
+	}, 10_000);
 
 	it.skip("should properly blame the action run if an action modifies a pub", async () => {
 		const trx = getTrx();
--- a/core/actions/api/serverAction.ts
+++ b/core/actions/api/serverAction.ts
@@ -1,9 +1,11 @@
 "use server";
 
 import type { UsersId } from "db/public";
+import { Capabilities, MembershipType } from "db/public";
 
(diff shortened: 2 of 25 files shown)


core/app/components/SubmitButton.tsx
```diff
@@ -82,9 +82,9 @@ export const SubmitButton = ({
 			setButtonState("success");
 		} else if (isSubmitError) {
 			setErrorState();
+		} else {
+			setButtonState("idle");
 		}
-
-		setButtonState("idle");
```
tefkah, line 87: not sure how it was in this state for this long, but basically the loading wasn't working at all, it was just stuck in the idle always

core/app/login/LoginForm.tsx
```diff
@@ -34,17 +36,21 @@ export default function LoginForm() {
...
-		await runLoginWithPassword({
+		const result = await runLoginWithPassword({
 			email: formData.email,
 			password: formData.password,
 			redirectTo: searchParams.get("redirectTo") ?? null,
 		});
+
+		if (result && result.error) {
+			form.setError("root", { message: result.error });
+		}
```
tefkah, line 47: im not super happy about this, but this is basically a way of signaling to `FormSubmitButton` that there's a diff between "the submit action ran without error, but the sign in was unsuccessful" and "the user signed in successfully"

gabestein (approved):
Tested on the branch and works/looks great. Thanks for the improvement!

PR #1319, opened 2025-07-22
fix: allow contributors to delete pubs + manage capabalities from code

### LINKED ISSUE #1311: Allow contributors to delete pubs they have access to
## Motivation

So contributor users can delete pubs, per https://github.com/pubpub/platform/issues/656

## Requirements

Right now, it gives them an unauthorized error when they attempt to.

tefkah:
## Issue(s) Resolved

Resolves #1311 

Kind of an extremely roundabout way of just adding a new capability to the database, namely to allow pub contributors and pub editors, to delete pubs they are added to. 

In addition to that, I implemented this system for syncing the capabilities from code to the database rather than having them just kind of hard-coded in the database. 

This is something that I've been wanting to do for a while. And while it is a bit icky and a bit reminiscent of CorePubFields and facets, I think it is fine because we are already defining capabilities in the database. This is just ergonomics. 

Whether these things should live in the database at all maybe is a separate issue, but for now we can at least way more easily edit the pubs and have a better overview of which users have which capabilities.  The previous situation was, imo,  extremely annoying. Just looking through the migrations to try and  figure out whether why something wasn't working isnt great DX. Now at least we have this object which is at least marginally easier to go through.

## Test Plan

- Convince yourself added test is accurate
- Remove a capability from the `CAPABILITY_MAPPING`, maybe edit another one, and add a new capability, then run `pnpm capabilities:generate-migration`. Have a look at said migration and convince self it makes sense.
- Optional: add a new contributor to a pub, log in as contributor, observe you are able to remove pub.


core/playwright/pub.capabilities.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+			}
+		});
+
+		// going back to pub details page should show 404
+		await pubDetailsPage.goTo();
+
+		expect(requests).toHaveLength(1);
+		expect(requests[0].status()).toBe(404);
+	});
+});
```
tefkah, line 119: just a very simple test to see if a contributor can remove a pub, and if so, that the pub is actually removed

3mcd (approved):

PR #1320, opened 2025-07-23
fix: add user to form when creating pub to allow editing

3mcd:


tefkah:
looks good, except that there is an additional problem if you go from eg viewing the private review page to clicking on edit. ill go into more detail tomorrow, but this is kinda hairy, only happens when there's a context editor with a value. it should also happen when you switch the form view on the edit page, but it doesn't happen now because form switching is broken (i just think it never worked lol)

this is good to go tho!

tefkah (approved):

PR #1322, opened 2025-07-23
feat: add fallback attribute to value directive

### LINKED ISSUE #1299: Add fallback attribute to email directive
## Motivation

So users can send emails automatically even if a field in the email template is blank. E.g. right now, the ICOR/ORB community contains a "submission feedback" field that is optional on the form. They don't always want to fill it in, but do want to be able to include the feedback if they have it.

## Requirements

- Adds a fallback attribute that, if filled out, allows the email to be sent even if the field value doesn't exist
- Fallback can be an empty string
- If no fallback and field is empty, email should fail

3mcd:
## Issue(s) Resolved

Resolves #1299

Adds a new `fallback` attribute to the `:value` directive so emails can still be sent for pubs with optional/missing field values.

## Test Plan

1. Run the email action with a template containing a value directive for a value that doesn't exist. Specify a fallback attribute, e.g. `:value{field="pubpub:nonsense" fallback="(No value was present)"}`.
2. The sent email should render "(No value was present)" instead of failing to send.

### DIFF
--- a/core/lib/server/render/pub/renderMarkdownWithPub.ts
+++ b/core/lib/server/render/pub/renderMarkdownWithPub.ts
@@ -60,7 +60,7 @@ const visitValueDirective = (node: Directive, context: utils.RenderWithPubContex
 				val.schemaName === CoreSchemaType.DateTime
 					? // get the date in YYYY-MM-DD format
 						// we should allow the user to specify this
-						new Date(value as string).toISOString().split("T")[0]
+						new Date(val.value as string).toISOString().split("T")[0]
 					: val.value;
 		}
 	}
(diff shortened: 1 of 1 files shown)


gabestein:
Thanks for this! Just want to double-check: can empty string be a fallback?

tefkah (approved):
Looks and works good!

The only caveat I have is that `:value{field="title" fallback="something"}` will never fallback to `something`, even if title is technically empty (you always get `Untitled X ....`. I think that is fine, I don't think we should be thinking of titles like that, ideally we would set `title: "Untitled X..."` in the db rather than doing that at the application level, but that's for later.

I say merge it!

PR #1337, opened 2025-08-19
feat: use consistent headers almost everywhere

tefkah:
- **feat: add contentlayout for members page**
- **feat: add contentLayout to action run page**
- **feat: prettify token page**


core/playwright/fixtures/api-token-page.ts
tefkah, line 1: needed to make some minor adjustments so the tests still worked

PR #1339, opened 2025-08-20
chore: upgrade next (and other dependencies)

tefkah:
## Issue(s) Resolved

Mostly wanted to do this bc i messed up #1284  and i wanted to get the other updates out of the way before retrying it.

Not that many changes, next 15.5/4 does technically allow us to build with `turbopack`, but sadly it's not really faster.

Secondly, `next` 15.5 enables node middleware. This could allow us to remove all of the community/user checks inside of `c/[communitySlug]` and move that to the middleware instead and store the user/community in cookies.

That may not be a good idea tho, i say we leave it for now

### should we enable `turbo` for builds

for now, i'll leave it off, as it increases(!) CI build times

#### locally

Turbo:
```sh
Executed in   41.69 secs    fish           external
   usr time  122.94 secs  185.00 micros  122.94 secs
   sys time   27.17 secs  785.00 micros   27.17 secs
```

standard
```sh
Executed in   63.36 secs    fish           external
   usr time   90.41 secs    0.23 millis   90.41 secs
   sys time   14.20 secs    2.10 millis   14.20 secs
```

#### ci

standard: 4m 20s 

turbo: 4m 40s!! also the bundle size seems to be bigger.

### DIFF
--- a/core/playwright/formAccess.spec.ts
+++ b/core/playwright/formAccess.spec.ts
@@ -294,10 +294,7 @@ test.describe("public forms", () => {
 			}, url!);
 
 			await page.goto(decodedUrl);
-			await page
-				.getByRole("status")
-				.getByText("Your email is now verified", { exact: true })
-				.waitFor();
+			await page.getByText("Your email is now verified", { exact: true }).waitFor();
 			await page.waitForURL(fillUrl, { timeout: 5_000 });
 		});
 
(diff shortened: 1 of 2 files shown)


core/playwright/formAccess.spec.ts
```diff
@@ -294,10 +294,7 @@ test.describe("public forms", () => {
 			}, url!);
 
 			await page.goto(decodedUrl);
-			await page
-				.getByRole("status")
-				.getByText("Your email is now verified", { exact: true })
-				.waitFor();
+			await page.getByText("Your email is now verified", { exact: true }).waitFor();
 			await page.waitForURL(fillUrl, { timeout: 5_000 });
```
tefkah, line 298: whoops, this was meant for #1338

PR #1359, opened 2025-09-16
refactor: remove server component config option for actions

tefkah:
## Issue(s) Resolved

-

Looking forward to #1356 , I did not want to add _yet another_ weird `Action*FormWrapper`. I think this pattern was a mistake, and the current way we do custom Action Fields using custom `zod` types + client components is just better I think.

Therefore I removed all the option to add custom server components for Action Fields. You can only add client components now.

There is one complication:
- In order to use the `zod` approach the component needs to be defined in `ui`. 
- This is not always possible.
- In that case, you can manually refer to the component in the `fieldConfig`, see

https://github.com/pubpub/platform/blob/d8847589750966d0b6e12eba79f7dcb0aa87cf2f/core/actions/email/action.tsx#L48-L57

### Next

There are two more refactors I want to do concerning the Actions, which are related

1. Decouple the `ActionRunForm` from its button.

Currently we are rendering _so_ many of these forms on eg the All Pubs page. This is actually the bottleneck when searching with the setting at 100 pubs, not the db. We should instead only render one `ActionRunForm` per page, and use search params to load the relevant data.

Related to that, there are now a ton of `<PubFieldProvider>/<StagesProvider>` over the place, with the only use being to provide the fields and stages respectively for the `ActionRunForm`. Once we have decoupled the `ActionRunForm` from its button, we can just load the stages and fields over there.

An alternative I have considered to the above is just adding the `stages` and `fields` to the `CommunityProvider`. I didn't end up doing that because I don't want to get into the habit of just loading a ton of data at the start everywhere by default when it's likely not needed. I get `v6-initialData` related flashbacks haha

2. Treat all Action related forms and data kind of the same.

I don't like that we have a `Config`, `Run`, `Default`, and now with #1356 a `Rule` form for configuring actions.

I would rather they all use the same form (as in `react-hook-form`) and just do different things depending on submission depending on the usecase.

Relatedly, I also don't like that we have 4 sources of data for the forms. We have discussed this in the past and am not completely sure where we landed on that, but my understanding was that we were in favor of removing the separation between `config` and `params`, in that running an action would always use the same fields as the config.

Then the final configuration would be something like (properly merged):

```ts
const finalConfig = {
     ...defaultConfig,
     ...config,
     ...(params || ruleConfig)
}
```

3. (secret bonus)

This is partially what #1356 does, except for making it permanent so to say.

As we also discussed, I would really like it if we move away from these ways of thinking about actions:

- `ActionInstances` exist on a `Stage` with a certain Config
- `ActionInstances` are run in response to `Rules`, using their existing config, or by being triggered manually, which can be a combination of the existing config and the provided params, which may be different from the values that can be configured
- `ActionInstance` are always run on a `Pub`

and instead towards

- ActionInstances _are_ a `Rule` on a stage triggering an action with a certain Config
- Manual action triggers are just another rule. You can still provide params, but those are just overrides of the `config`
- `ActionInstances` can be run on `Pub`s or generic json objects.

### DIFF
--- a/core/playwright/pub.spec.ts
+++ b/core/playwright/pub.spec.ts
@@ -252,7 +252,8 @@ test.describe("Creating a pub", () => {
 		// Now update
 		await page.getByRole("link", { name: "Update" }).first().click();
 		await page.locator(".ProseMirror").click();
-
+		// move the cursor to the beginning of the editor
+		await page.keyboard.press("Home");
 		await page.keyboard.type("prefix ");
 
 		await page.getByRole("button", { name: "Save" }).click();


3mcd (approved):
Such a major improvement. Happy to be rid of the old pattern. This is a lot easier to understand!

PR #1365, opened 2025-09-19
fix: make file deletions/uploads work better and safer

tefkah:
- **fix: maybe make test less flaky**
- **fix: dont try to make permanent files permanent, otherwise you cannot add files to a pub**
- **fix: dont try to delete non-pubpub files**
- **fix: dont remove files if theyre used in other places**
- **fix: give s3 handler proper perms**

### DIFF
--- a/core/app/components/forms/actions.ts
+++ b/core/app/components/forms/actions.ts
@@ -4,6 +4,7 @@ import { sql } from "kysely";
 
 import type { CommunitiesId, FormsId, PubsId } from "db/public";
 import { logger } from "logger";
+import { tryCatch } from "utils/try-catch";
 
 import type { XOR } from "~/lib/types";
 import { db } from "~/lib/__tests__/db";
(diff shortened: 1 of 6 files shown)


core/app/components/forms/actions.ts
```diff
@@ -86,17 +92,17 @@ export const deleteFile = defineServerAction(async function deleteFile({
...
-									.selectFrom("pub_fields")
-									.select("slug")
-									.where("slug", "=", fieldSlug)
-							)
+						.where("pub_values.fieldId", "=", (eb) =>
+							eb
+								.selectFrom("pub_fields")
+								.select("pub_fields.id")
+								.where("pub_fields.slug", "=", fieldSlug)
+								.limit(1)
```
tefkah, line 105: previous iteration would delete _all_ files on all fields of the pub if they have the same url, which isn't what we want

core/lib/server/assets.ts
```diff
@@ -134,12 +134,27 @@ export const generateSignedAssetUploadUrl = async (
...
 	const client = getPublicS3Client();
 	const bucket = env.ASSETS_BUCKET_NAME;
 
 	const fileKey = fileUrl.split(new RegExp(`^.+${env.ASSETS_BUCKET_NAME}/`))[1];
 
+	if (!fileKey) {
+		logger.error({ msg: "Unable to parse URL of uploaded file", fileUrl });
+		throw new InvalidFileUrlError("Unable to parse URL of uploaded file");
+	}
+
```
tefkah, line 157: this check should be better, but it's ok for now

3mcd (approved):

PR #1411, opened 2025-12-17
feat: automation resolvers

3mcd:
## Issue(s) Resolved

N/A

This PR implements **automation resolvers**, which provide the means to dynamically "load" a Pub (or some custom JSON) into the automation context.

If a Pub or JSON already existed on the automation context, it will be overwritten with the new, resolved value. Otherwise the resolved value is merged (meaning an automation context can have `{pub}`, `{json}`, or `{pub, json}`.


tefkah (changes requested):
i think we should these two things, but otherwise good to go!

tefkah (approved):
LGTM!

PR #1428, opened 2026-02-23
feat: de-astro-ify the site builder; add extensions to pages

3mcd:


site-builder-2/server/server.ts
```diff
@@ -384,6 +319,97 @@ const verifySiteBuilderToken = async (authHeader: string, communitySlug: string)
...
+
+			const normalized = slug.replace(/\/+$/, "")
+			const fileName = normalized === "" ? `index.${extension}` : `${normalized}.${extension}`
+			const filePath = path.join(distDir, fileName)
+			await fs.mkdir(path.dirname(filePath), { recursive: true })
+			await fs.writeFile(filePath, fileContent, "utf-8")
+		}
+	}
+}
+
```
tefkah, line 429: worlds teeniest sitebuilder! 

only thing i would change: not a huge fan of "for ... await", we could probably just promise.all that

also i think there is a default max length for the response of `pub.getMany`, even with `pubIds` provided, we should probably fetch that in steps until there's nothing left

tefkah (approved):
lgtm!

i only had one comment, which i fixed later (see other comments)

I changed 4 things in the commits i added:

- optimized the fetching a bit 
- put files in `{slug}/index.html`, rather than in `{slug}.html`. this way `/{slug}` takes you to the file, rather than having to go to `/{slug}.html`. 
- added some logging with `pino`
- added a link to the `json` file in the template

otherwise everything works swimmingly, great job! soooo much simpler now

